PR #398, opened 2015-10-29
(PUP-5064) Pin specific Windows packaging version

Iristyle:
- Previously the Windows packaging was tracking the master branch.
   However, use a specific version here.


Iristyle:
Sticking with tracking `master` on PFW, rather than pinning.  Closing without merge.

PR #445, opened 2015-12-03
(RE-6012, RE-6116, RE-6111) Build OpenSSL, Curl, and our ca certs on windows

melissa:


configs/projects/puppet-agent.rb
```diff
@@ -89,20 +145,29 @@
...
-  proj.component "ruby-shadow" unless platform.is_aix?
-  proj.component "ruby-augeas"
+  if platform.is_windows?
+    proj.component "rubygem-ffi"
+    proj.component "rubygem-minitar"
+    proj.component "rubygem-win32-dir"
+    proj.component "rubygem-win32-eventlog"
+    proj.component "rubygem-win32-process"
+    proj.component "rubygem-win32-security"
+    proj.component "rubygem-win32-service"
```
joshcooper: Super cool to see this! Are there individual `config/components/rubygem-win32-dir.rb` components? I'm not seeing them in this PR, maybe need to be `git add`ed?

configs/projects/puppet-agent.rb
```diff
@@ -1,34 +1,63 @@
...
 
-  if platform.is_eos?
-    proj.setting(:sysconfdir, "/persist/sys/etc/puppetlabs")
-    proj.setting(:link_sysconfdir, "/etc/puppetlabs")
-  elsif platform.is_osx?
-    proj.setting(:sysconfdir, "/private/etc/puppetlabs")
+  # Project level settings our components will care about
+  # Windows has its own seperate layout
+  if platform.is_windows?
+    proj.setting(:install_root, "#{platform.drive_root}/ProgramFiles/PuppetLabs")
```
joshcooper: You might be able to use 8.3 pathing, but that may also introduce other issues:

```
"#{platform.drive_root}/Progra~1/Puppet~1"
```

configs/components/runtime.rb
```diff
@@ -23,6 +25,16 @@
   if platform.is_aix?
     pkg.install_file File.join(libdir, "libstdc++.a"), "/opt/puppetlabs/puppet/lib/libstdc++.a"
     pkg.install_file File.join(libdir, "libgcc_s.a"), "/opt/puppetlabs/puppet/lib/libgcc_s.a"
+  elsif platform.is_windows?
+    arch = platform.architecture == "x64" ? "64" : "32"
+    lib_type = platform.architecture == "x64" ? "seh" : "sjlj"
+    pkg.install_file "/cygdrive/c/tools/mingw#{arch}/bin/libgcc_s_#{lib_type}-1.dll", settings[:bindir]
+    pkg.install_file "/cygdrive/c/tools/mingw#{arch}/bin/libstdc++-6.dll", settings[:bindir]
+    pkg.install_file "/cygdrive/c/tools/mingw#{arch}/bin/libwinpthread-1.dll", settings[:bindir]
```
mckern: We have a function for the first part of this path, right? And we can probably normalize the entire binpath here.

configs/projects/puppet-agent.rb
melissa: These have yet to be implemented. I need to build ruby first. I probably shouldn't have included them in this PR...

melissa:
I made some changes, and with the opening of the puppet-agent repo, I had to reopen this PR https://github.com/puppetlabs/puppet-agent/pull/455

configs/projects/puppet-agent.rb
Iristyle: I assume we also support patches against these dependencies - but not sure where that happens? Are they stored in `resources/patches/ruby` or somewhere else?

configs/projects/puppet-agent.rb
Iristyle: Since this is a VMPooler host, and should be predictable, the `~1` hardcoding should be OK - but note that it's not always `~1`, depending on what other dirs are present on disk and when they were created.

configs/projects/puppet-agent.rb
melissa: @Iristyle this PR is closed in favor of the one linked. I have yet to implement these gem components.

PR #454, opened 2015-12-16
(PA-131) Add leatherman as its own component

branan:
This enables us to un-vendor leatherman from all of our individual
projects. Until those projects are ready to consume an external
leatherman install, this will go unused. We can't remove the vendoring
until we have this in place in P-A, though.

### DIFF
--- a/configs/components/leatherman.rb
+++ b/configs/components/leatherman.rb
@@ -1,0 +1,93 @@
+component "leatherman" do |pkg, settings, platform|
+  pkg.load_from_json('configs/components/leatherman.json')
+
+  if platform.is_osx?
+    pkg.build_requires "cmake"
+    pkg.build_requres "boost"
+  elsif platform.name =~ /huaweios/
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/HuaweiOS/#{platform.os_version}/ppce500mc/pl-gcc-4.8.2-1.huaweios6.ppce500mc.rpm"
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/HuaweiOS/#{platform.os_version}/ppce500mc/pl-cmake-3.2.3-1.huaweios6.ppce500mc.rpm"
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/HuaweiOS/#{platform.os_version}/ppce500mc/pl-boost-1.58.0-1.huaweios6.ppce500mc.rpm"
+  elsif platform.name =~ /solaris-10/
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/solaris/10/pl-gcc-4.8.2.#{platform.architecture}.pkg.gz"
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/solaris/10/pl-binutils-2.25.#{platform.architecture}.pkg.gz"
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/solaris/10/pl-boost-1.58.0-1.#{platform.architecture}.pkg.gz"
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/solaris/10/pl-cmake-3.2.3-2.i386.pkg.gz"
+  elsif platform.name =~ /solaris-11/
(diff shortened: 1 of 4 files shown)


branan:
@melissa It'd be rad to get your feedback on this today (meaning Thursday), and I'll take care of it when my Thursday rolls around

configs/components/leatherman.rb
```diff
@@ -0,0 +1,93 @@
...
+    test = ":"
+  else
+    test = "#{platform[:make]} test ARGS=-V"
+  end
+
+  pkg.build do
+    # Until a `check` target exists, run tests are part of the build.
+    [
+      "#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)",
+      "env LEATHERMAN_RUBY=#{settings[:libdir]}/$(shell #{ruby} -e 'print RbConfig::CONFIG[\"LIBRUBY_SO\"]') #{test}"
```
branan: we need to set LEATHERMAN_RUBY here becuase *facter* is the project that knows where the ruby library is located, not leatherman.

MikaelSmith: Maybe we'll want to change that at some point. For now this makes sense.

configs/components/leatherman.rb
```diff
@@ -0,0 +1,93 @@
+component "leatherman" do |pkg, settings, platform|
+  pkg.load_from_json('configs/components/leatherman.json')
+
+  if platform.is_osx?
+    pkg.build_requires "cmake"
+    pkg.build_requres "boost"
```
MikaelSmith: Typo

MikaelSmith:
Leatherman and cpp-pcp-client are similar, in that they only create static libraries. So installing this to prefix is consistent with cpp-pcp-client.

configs/projects/puppet-agent.rb
```diff
@@ -89,6 +89,7 @@
   end
 
   # First our stuff
+  proj.component "leatherman"
```
melissa: I feel like this should be at the end of the list....

melissa:
@branan the looks like it makes sense. Does it build for all platforms?

configs/components/leatherman.rb
```diff
@@ -0,0 +1,93 @@
...
+    cmake = "/opt/pl-build-tools/bin/cmake"
+  end
+
+  pkg.configure do
+    ["#{cmake} \
+        #{toolchain} \
+        -DCMAKE_VERBOSE_MAKEFILE=ON \
+        -DCMAKE_PREFIX_PATH=#{settings[:prefix]} \
+        -DCMAKE_INSTALL_PREFIX=#{settings[:prefix]} \
+        -DLEATHERMAN_SHARED=TRUE \
```
MikaelSmith: Oh, doing a shared library install?

MikaelSmith:
The Windows build scripts also need to be updated, unless we want to hold this open until Windows Vanagon builds are done.

branan:
@MikaelSmith Oh yeah, windows :(. I think we do want to update the script, unless the Windows Vanagon stuff is very very nearly finished. We can't merge the Facter PR until windows can build with the submodule.

configs/components/leatherman.rb
branan: I think my cold-medicine-addled brain thought that, since we were installing to the prefix anyway, we might as well switch to SOs. I'm happy to go back to static for now if we're not quite ready for a shared build.

configs/projects/puppet-agent.rb
branan: Yup. I moved it up because I forgot to save my changes to the facter file and was very confused about why my new build dependency wasn't making leatherman build first -_-

configs/components/leatherman.rb
branan: how does this even run?

MikaelSmith: Didn't build on OS X? Probably not invoking that part of the script.

branan: oh right, this is ruby -_-

branan:
@melissa @MikaelSmith This is updated to support Windows now. Still needs to be tested on a few more platforms

branan:
This apparently got disconnected from my fork when we switched everything public. I'm going to close and re-open this PR

PR #455, opened 2015-12-17
(RE-6013, RE-6012, RE-6116, RE-611) Build OpenSSL, Curl, and our ca certs for windows

melissa:
This initial commit adds in the base work to build the puppet-agent project on windows, and modifies the curl, openssl, puppet-ca-bundle and runtime components to run on windows.

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -5,17 +5,42 @@ component 'curl' do |pkg, settings, platform|
 
   pkg.build_requires "openssl"
 
+  prefix = settings[:prefix]
+  pem_file = "#{settings[:prefix]}/ssl/cert.pem"
+
+  make = platform[:make]
+
+  if platform.is_windows?
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
+    pkg.build_requires "runtime"
+
+    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "CC" => settings[:cc]
+    pkg.environment "CXX" => settings[:cxx]
+
+    prefix = platform.convert_to_windows_path(settings[:prefix])
+    pem_file = platform.convert_to_windows_path("#{settings[:puppet_configdir]}/ssl/cert.pem")
+
+    make = "/usr/bin/make"
+  end
+
   pkg.configure do
     ["LDFLAGS='#{settings[:ldflags]}' \
-     ./configure --prefix=#{settings[:prefix]} \
-     --with-ssl=#{settings[:prefix]} --enable-threaded-resolver --disable-ldap --disable-ldaps --with-ca-bundle=#{settings[:prefix]}/ssl/cert.pem"]
+     ./configure --prefix=#{prefix} \
+        --with-ssl=#{prefix} \
+        --enable-threaded-resolver \
+        --disable-ldap \
+        --disable-ldaps \
+        --with-ca-bundle=#{pem_file} \
+        #{settings[:host]}"]
   end
 
   pkg.build do
-    ["#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
+    ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
   end
 
   pkg.install do
-    ["#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
+    ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
   end
 end
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -4,6 +4,7 @@ component 'curl' do |pkg, settings, platform|
   pkg.url "http://buildsources.delivery.puppetlabs.net/curl-#{pkg.get_version}.tar.gz"
 
   pkg.build_requires "openssl"
+  pkg.build_requires "puppet-ca-bundle"
 
   prefix = settings[:prefix]
   pem_file = "#{settings[:prefix]}/ssl/cert.pem"
--- a/configs/components/openssl.rb
+++ b/configs/components/openssl.rb
@@ -30,7 +30,19 @@ component "openssl" do |pkg, settings, platform|
     pkg.build_requires 'runtime'
   elsif platform.is_osx?
     pkg.build_requires 'makedepend'
+  elsif platform.is_windows?
+    pkg.apply_patch 'resources/patches/openssl/openssl-1.0.0l-use-gcc-instead-of-makedepend.patch'
+    # This patch removes the option `-DOPENSSL_USE_APPLINK` from the mingw openssl congifure target
+    # This brings mingw more in line with what is happening with mingw64. All applink does it makes
+    # it possible to use the .dll compiled with one compiler with an application compiled with a
+    # different compiler. Given our openssl should only be interacting with things that we build,
+    # we can ensure everything is build with the same compiler.
+    pkg.apply_patch 'resources/patches/openssl/openssl-mingw-do-not-build-applink.patch'
+    pkg.build_requires "runtime"
   end
+
+  prefix = settings[:prefix]
+  make = platform[:make]
 
   if platform.is_osx?
     pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH:/usr/local/bin"
@@ -54,6 +66,17 @@ component "openssl" do |pkg, settings, platform|
     ldflags = ''
     cflags = "$${CFLAGS} -static-libgcc"
     pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
+  elsif platform.is_windows?
+    target = platform.architecture == "x64" ? "mingw64" : "mingw"
+    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "CC" => settings[:cc]
+    pkg.environment "CXX" => settings[:cxx]
+    make = "/usr/bin/make"
+    pkg.environment "MAKE" => make
+    prefix = platform.convert_to_windows_path(settings[:prefix])
+    cflags = settings[:cflags]
+    ldflags = settings[:ldflags]
   else
     pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH:/usr/local/bin"
     if platform.architecture =~ /86$/
@@ -78,9 +101,9 @@ component "openssl" do |pkg, settings, platform|
     # since configure uses the existence of a lib64 directory to determine
     # if it should install its own libs into a multilib dir. Yay OpenSSL!
     "./Configure \
-      --prefix=#{settings[:prefix]} \
+      --prefix=#{prefix} \
       --libdir=lib \
-      --openssldir=#{settings[:prefix]}/ssl \
+      --openssldir=#{prefix}/ssl \
       shared \
       no-asm \
       #{target} \
@@ -103,8 +126,8 @@ component "openssl" do |pkg, settings, platform|
   end
 
   pkg.build do
-    ["#{platform[:make]} depend",
-    "#{platform[:make]}"]
+    ["#{make} depend",
+    "#{make}"]
   end
 
   if platform.is_aix?
@@ -113,8 +136,10 @@ component "openssl" do |pkg, settings, platform|
     end
   end
 
+  install_prefix = "INSTALL_PREFIX=/" unless platform.is_windows?
+
   pkg.install do
-    ["#{platform[:make]} INSTALL_PREFIX=/ install"]
+    ["#{make} #{install_prefix} install"]
   end
 
   pkg.install_file "LICENSE", "#{settings[:prefix]}/share/doc/openssl-#{pkg.get_version}/LICENSE"
--- a/configs/components/runtime.rb
+++ b/configs/components/runtime.rb
@@ -8,6 +8,8 @@ component "runtime" do |pkg, settings, platform|
   elsif platform.is_aix?
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-gcc-5.2.0-1.aix#{platform.os_version}.ppc.rpm"
     libdir = "/opt/pl-build-tools/lib/gcc/powerpc-ibm-aix#{platform.os_version}.0.0/5.2.0/"
+  elsif platform.is_windows?
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
   else
     pkg.build_requires "pl-gcc"
   end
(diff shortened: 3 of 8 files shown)


configs/projects/puppet-agent.rb
```diff
@@ -1,34 +1,63 @@
...
 
-  if platform.is_eos?
-    proj.setting(:sysconfdir, "/persist/sys/etc/puppetlabs")
-    proj.setting(:link_sysconfdir, "/etc/puppetlabs")
-  elsif platform.is_osx?
-    proj.setting(:sysconfdir, "/private/etc/puppetlabs")
+  # Project level settings our components will care about
+  # Windows has its own seperate layout
+  if platform.is_windows?
+    proj.setting(:install_root, "#{platform.drive_root}/ProgramFiles/PuppetLabs")
```
melissa: from @joshcooper re: getting around spaces in paths: `"#{platform.drive_root}/Progra~1/Puppet~1"`

joshcooper: Also I should mention that the above isn't 100% reliable, as it could be Progra~2 or Puppet~2. It all depends on what other directories exist. That said if we're building on clean VMs I'm less concerned.

melissa:
32 bit OpenSSL is failing to build :(

configs/components/curl.rb
```diff
@@ -4,18 +4,45 @@
...
+
+  if platform.is_windows?
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
+    pkg.build_requires "runtime"
+
+    arch = platform.architecture == "x64" ? "64" : "32"
+    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "CC" => settings[:cc]
+    pkg.environment "CXX" => settings[:cxx]
```
jcoconnor: Should we adopt these "settings" for pl-build-tools as well?

jcoconnor:
Background question - why aren't be building this in pl-build-tools ?

configs/components/curl.rb
```diff
@@ -4,18 +4,45 @@
...
   end
 
   pkg.build do
-    ["#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
+    ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
   end
 
   pkg.install do
-    ["#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
+    ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
```
jcoconnor: Why the change and should we do this in pl-tools-vanagon ?

configs/projects/puppet-agent.rb
glennsarti: @joshcooper +1 to you're pretty safe with shortnames for Program Files.  Another gotcha will be ProgFiles vs ProgFiles (x86).  Would this be something to be concerned about?

configs/projects/puppet-agent.rb
```diff
@@ -47,6 +76,14 @@
         proj.setting(:host_gem, "/opt/pl-build-tools/bin/gem")
       end
     end
+  elsif platform.is_windows?
+    # For windows, we need to ensure we are building for mingw not cygwin
+    if platform.architecture == "x64"
+      platform_triple = "x86_64-unknown-mingw32"
+    else
+      platform_triple = "i686-unknown-mingw32"
+    end
```
jcoconnor: Use Ternary here ?

configs/components/openssl.rb
```diff
@@ -54,6 +66,23 @@
...
     cflags = "$${CFLAGS} -static-libgcc"
     pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
+  elsif platform.is_windows?
+    if platform.architecture == "x64"
+      target = "mingw64"
+      arch = "64"
+    else
+      target = "mingw"
+      arch = "32"
+    end
```
jcoconnor: Why not use ternary here similar to runtime.rb below?

configs/components/curl.rb
melissa: If you want to... I don't want to go back and rebuild all those packages just for this though

configs/components/curl.rb
melissa: Because we're calling a different make... Is pl-build-tools-vanagon not working?

configs/projects/puppet-agent.rb
melissa: @glennsarti I think we're just installing in ProgFiles, so we shouldn't have to worry about ProgFiles (x86). Josh can verify that though

configs/components/curl.rb
jcoconnor: ok - was just asking ;)

configs/projects/puppet-agent.rb
joshcooper: @glennsarti good point. I was assuming we're always building 32-bit MSIs on 32-bit Windows and building 64-bit MSIs on 64-bit Windows, but I think that's a bad assumption, since I think we always build on 2012r2. So yeah ProgFiles (x86) vs ProgFiles could be an issue. It appears that ~1 maps to ProgFiles and ~2 maps to ProgFiles (x86):

```
Administrator@mm1fxccjkyk8b1h /cygdrive/c
$ cygpath -ad "Program Files"
C:\PROGRA~1

Administrator@mm1fxccjkyk8b1h /cygdrive/c
$ cygpath -ad "Program Files (x86)"
C:\PROGRA~2
```

I would expect that to always be the case, but maybe not?

glennsarti: The program files dirs are laid down during OS install so while it is _possible_ to make them not ~1 and ~2, it's such an extreme edge case.  Also it is possible to disable 8.3 name generation in NTFS but again, it's an edge case and _generally_ you wouldn't do it prior to the OS installation there fore the 8.3 names for prog file will still get generated.

configs/components/curl.rb
melissa: if you want to make this change in pl-build-tools-vanagon, you're welcome to, but I'm really not worried about it. I consider that repo our practice ground. We found out some things that worked and some things that didn't. In puppet-agent, we get to make things more elegant and official.

configs/components/openssl.rb
melissa: No reason. It just seemed more legible at the time. Do you think it would be more logical to have these be ternaries?

jcoconnor: More consistency really - functionally doesn't make a difference. You definitely don't need a ternary on line 116+ as it would make it unreadable.

jcoconnor:
Have successfully built this for Windows 64 bit, with the following mods to the build env:
* Disable facter, ruby depends (not done yet).
* Ensure latest vanagon/vanagon_for_the_win branch is pulled.
* Cherry picked in change 74b44ad34ec9bce671ca530b2113b9aa2e912022 to vanagon.

So this looks good (on 64).
:+1:

melissa:
@jcoconnor what are you asking about above when you ask why we're not building it in pl-build-tools? OpenSSL? Curl? Puppet ca bundle? Runtime? We're building a lot of things in this PR... You have to be a little more specific with your question.

The main reason is that we need to deliver the files as a part of the puppet agent package. The easiest way to do that is build them here so vanagon knows to include the files in the final package (remember the file-list functionality I was telling you about?).

configs/projects/puppet-agent.rb
melissa: @joshcooper how do I go about setting it up to pick up `Puppet~1` as Puppet Labs? Is that just going to look for the first directory in that path with a space in it?

joshcooper: I think it's safe to hard code as Puppet~1 since the VM shouldn't have Puppet already installed.

configs/components/curl.rb
```diff
@@ -4,18 +4,45 @@
...
+  prefix = settings[:prefix]
+  pem_file = "#{settings[:prefix]}/ssl/cert.pem"
+
+  make = platform[:make]
+
+  if platform.is_windows?
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
+    pkg.build_requires "runtime"
+
+    arch = platform.architecture == "x64" ? "64" : "32"
```
MikaelSmith: This looks like it's no longer needed.

configs/components/openssl.rb
```diff
@@ -54,6 +66,23 @@
     ldflags = ''
     cflags = "$${CFLAGS} -static-libgcc"
     pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
+  elsif platform.is_windows?
+    if platform.architecture == "x64"
```
MikaelSmith: arch is no longer needed, this could be replaced with `target = platform.architecture == "x64" ? "mingw64" : "mingw"`

configs/components/openssl.rb
```diff
@@ -54,6 +66,23 @@
...
+      target = "mingw64"
+      arch = "64"
+    else
+      target = "mingw"
+      arch = "32"
+    end
+    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:#{settings[:tools_root]}/bin:$$PATH"
```
MikaelSmith: Why tools_root here?

configs/components/runtime.rb
```diff
@@ -23,6 +25,16 @@
   if platform.is_aix?
     pkg.install_file File.join(libdir, "libstdc++.a"), "/opt/puppetlabs/puppet/lib/libstdc++.a"
     pkg.install_file File.join(libdir, "libgcc_s.a"), "/opt/puppetlabs/puppet/lib/libgcc_s.a"
+  elsif platform.is_windows?
+    arch = platform.architecture == "x64" ? "64" : "32"
+    lib_type = platform.architecture == "x64" ? "seh" : "sjlj"
+    pkg.install_file "#{platform.drive_root}/tools/mingw#{arch}/bin/libgcc_s_#{lib_type}-1.dll", "#{settings[:bindir]}/libgcc_s_#{lib_type}-1.dll"
```
MikaelSmith: This could use gcc_bindir.

configs/projects/puppet-agent.rb
```diff
@@ -1,34 +1,68 @@
...
-  elsif platform.is_osx?
-    proj.setting(:sysconfdir, "/private/etc/puppetlabs")
+  # Project level settings our components will care about
+  # Windows has its own seperate layout
+  if platform.is_windows?
+    # Our install prefix can't have spaces in it, so we take advantage of short cuts.
+    # However, in order to take advantage of these shortcuts, we need to explicitely
+    # create the directories.
+    proj.setting(:install_root, "#{platform.drive_root}/Progra~1/Puppet~1")
+    proj.directory "#{platform.drive_root}/Program Files/Puppet Labs"
```
melissa: @joshcooper I'm pretty sure this is working now! The only issue is that the install prefix is changing from `Puppet` to `puppet`, and I'm not sure why. We're setting it to upper case puppet on line 13.

MikaelSmith: Down below directory creation is using install_root. I'm not sure what that will do...

melissa: Good point, I'm only setting that if we're not building on windows

configs/projects/puppet-agent.rb
melissa: I think that would be too long of a line. This is more readable.

configs/components/curl.rb
```diff
@@ -4,18 +4,44 @@
...
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
+    pkg.build_requires "runtime"
+
+    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "CC" => settings[:cc]
+    pkg.environment "CXX" => settings[:cxx]
+
+    prefix = platform.convert_to_windows_path(settings[:prefix])
+    pem_file = platform.convert_to_windows_path("#{settings[:puppet_configdir]}/ssl/cert.pem")
```
melissa: @joshcooper is this where we decided to install the pem file? This currently expands to `C:/ProgramData/PuppetLabs/puppet/etc/ssl/cert.pem`

joshcooper: The ssldir looks correct based on https://github.com/puppetlabs/puppet-specifications/blame/master/file_paths.md#L222. But is it safe to drop `cert.pem` in there? Or can it conflict with things puppet may write to its `ssldir`?

MikaelSmith: I would probably give it a more descriptive name.

However, I don't see anything it would conflict with. Nothing at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L645-L770 does.

configs/projects/puppet-agent.rb
```diff
@@ -82,8 +121,26 @@
   end
 
   # Platform specific
-  proj.setting(:cflags, "-I#{proj.includedir} -I/opt/pl-build-tools/include")
-  proj.setting(:ldflags, "-L#{proj.libdir} -L/opt/pl-build-tools/lib -Wl,-rpath=#{proj.libdir}")
+  if platform.is_windows?
+    if platform.architecture == "x64"
+      arch = "64"
+      proj.setting(:cflags, "-IC:/tools/pl-build-tools/include -IC:/tools/mingw#{arch}/include -IC:/tools/mingw64/x86_64-w64-mingw32/include -I#{platform.convert_to_windows_path(proj.includedir)}")
```
MikaelSmith: Why do we need mingw64/x86_64-w64/mingw32/include? GCC should already know about that location.

MikaelSmith:
Other than what seems like extra settings in cflags and ldflags, :+1:

configs/components/curl.rb
```diff
@@ -4,18 +4,44 @@
...
+
+    make = "/usr/bin/make"
```
stahnma: Is there a reason we aren't using platform make throughout?

configs/components/curl.rb
```diff
@@ -4,18 +4,44 @@
...
+        --enable-threaded-resolver \
+        --disable-ldap \
+        --disable-ldaps \
+        --with-ca-bundle=#{pem_file} \
+        #{settings[:host]}"]
   end
 
   pkg.build do
-    ["#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
+    ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
```
stahnma: Why change from platform make to make?

configs/components/curl.rb
stahnma: Hmm. I think if we're going to call a different make here, we should make the default still be platform[:make] rather than make = "/usr/bin/make"...

configs/components/curl.rb
stahnma: Nevermind, I get it.

configs/components/curl.rb
stahnma: Missed the reassignment above. :fearful:

configs/components/curl.rb
stahnma: You did that, I misunderstood what was happening.

stahnma:
:+1:

configs/components/curl.rb
haus: If a different make is being used for all components in the agent, it would be worth setting `platform.make "/usr/bin/make"` in the windows platform config. Then you can avoid the intermediate variable here.

configs/projects/puppet-agent.rb
haus: This is settable in the platform config.

configs/platforms/win-2012-x64.rb
```diff
@@ -0,0 +1,19 @@
+platform "win-2012-x64" do |plat|
+  plat.vmpooler_template "win-2012r2-x86_64.make"
+
```
haus: you can define the platform triple here and in the x86 one below instead of in the project config.

configs/components/curl.rb
melissa: @haus it's not being used by all components :(

haus: sadness

jcoconnor: sadness is right, but we have found a small subset of components that won't work with ming-make, hence the need to change it. libffi in pl-build-tools-vanagon is another example that needs MSYS make or equiv.

configs/components/openssl.rb
```diff
@@ -54,6 +66,17 @@
     ldflags = ''
     cflags = "$${CFLAGS} -static-libgcc"
     pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
+  elsif platform.is_windows?
+    target = platform.architecture == "x64" ? "mingw64" : "mingw"
+    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "CC" => settings[:cc]
+    pkg.environment "CXX" => settings[:cxx]
+    make = "/usr/bin/make"
```
jcoconnor: Minor issue - my preference would be to group all the pkg.environment statements together and put the make reassignment after them. This is an odd-case, but wondering should you put the make assignment back to line 70 and follow it with pkg.environment for the make (if you think this is OCD speaking, feel free to ignore).

configs/platforms/win-2012-x64.rb
melissa: done

PR #456, opened 2015-12-17
(PA-68) Windows MSI for puppet-agent should not ship C++ unit test executables

glennsarti:
Added environment variable to determine whether test files should be removed from the puppet agent MSI package
If test files are not to be preserved, which by default they are not, then the files are removed prior to archiving.  This is achieved by deleting using simple filename filters for both pxp-agent and facter

### DIFF
--- a/bin/build-windows.rb
+++ b/bin/build-windows.rb
@@ -25,6 +25,9 @@ BUILD_SOURCE         = ENV['BUILD_SOURCE'] || '0'
 
 PRESERVE             = ENV['PRESERVE'] || false
 
+# Whether to keep unit test files in the packaged MSI
+PRESERVE_TESTFILES   = ENV['PRESERVE_TESTFILES'] || false
+
 # Parsed information that we need to specify in order to know where to find different built facter bits
 # and correctly pass information to the facter build script
 script_arch          = "#{ARCH == 'x64' ? '64' : '32'}"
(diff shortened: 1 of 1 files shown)


Iristyle:
Looking pretty good @glennsarti ... just a few more minor notes about commit msgs - we usually follow the [50 / 72 rule](http://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting) for msgs (for better log viewing at the command line).  So 50 char width subject and 72 char width body, separated by a newline... and we typically put the ticket number in parens as you have it in the PR subject line.

```
(PA-68) Remove test binaries from Windows MSI

 - Added environment variable to determine whether test files should be removed
   from the puppet agent MSI package.  If test files are not to be preserved,
   which by default they are not, then the files are removed prior to archiving.
   This is achieved by deleting using simple filename filters for both pxp-agent
   and Facter.
```

Aside from that minor issue, great first PR!

bin/build-windows.rb
```diff
@@ -137,6 +140,16 @@ def clone_and_rynsc_private_repo(fork, ref, hostname, ssh_key)
...
+if !(PRESERVE_TESTFILES)
+  # Remove test files prior to zipping
+  # facter
+  Kernel.system("set -vx;#{ssh_command} \"source .bash_profile ; rm #{archive_dest}/#{facter_zipname}/bin/*_test.exe ; rm #{archive_dest}/#{facter_zipname}/bin/lth_cat.exe ;\"")
+  fail "Cleaning source files prior to packaging failed" unless $?.success?
+  # pxp-agent
+  Kernel.system("set -vx;#{ssh_command} \"source .bash_profile ; rm #{archive_dest}/#{pxp_zipname}/bin/*-unittests.* ; rm #{archive_dest}/#{pxp_zipname}/bin/lth_cat.exe ;\"")
+  fail "Cleaning source files prior to packaging failed" unless $?.success?
+end
+
```
joshcooper: I'm not sure if this will cause problems, but we are planning on moving leatherman to a distinct puppet-agent component. Will test executables be reintroduced once that happens? /cc @MikaelSmith

Iristyle: @joshcooper if that happens, it probably changes other things in this script as well (including how files are copied into the directories).  So I think we can address the removal when that new component is introduced.

Iristyle:
@glennsarti looks like you need to squash your 3 commits down into one.

jpogran:
I'm coming late to the discussion, so this may have been discussed before. If so, please ignore.

We could add to the xslt filter in WIX to ignore these test files without having to change compile/build processes. This puts the decision of what to include in the MSI build process, and could possibly address @joshcooper comment

Iristyle:
@jpogran that's an interesting idea.  Can you spike a PR on that one for the sake of comparison?

jpogran:
In progress

Iristyle:
Thanks for adding your idea @jpogran. So as a quick follow up here:

* I had originally advised Glenn that we should do this up-front as the zip is built, so that the MSI building process is as dumb as possible and has not knowledge of the zip contents.  I was also thinking the only solution was to modify the Rake tasks to ignore zip contents - and I didn't particularly like that option.
* However, I forgot that we had talked about producing assets (i.e. the zip file) with the tests inside, so that the zip can be extracted so that the tests can be run in environments outside of where the bits are originally compiled.  While Linux platforms must build on the target platform, and run tests there, Windows is different. Windows has the luxury of really having only 2 build targets - x86 and x64.  So we can theoretically re-run tests across all Windows SKUs if the zip file contains them (we don't do this right now, and have some work to do to make it happen - /cc @melissa @jcoconnor)
* So the XSLT approach might be a reasonable one, assuming that we can standardize the names of test binaries or their relative paths during compilation.  I had forgotten this was possible - thanks for reminding us @jpogran.

We ultimately have to keep the logic *somewhere* for what to exclude when creating the MSI, and maybe XSLT makes as much sense as anywhere else.  The tradeoff is that the zips are fatter, but I don't think that's a big deal.

jpogran:
Added PR that uses filters https://github.com/puppetlabs/puppet_for_the_win/pull/132

While a little verbose, we don't have to worry about the locations of the files using this filter approach because of the way we are calling heat. We point heat to each directory, and the filter is a recursive search based on a simplistic match. We could make this one line less for each filter by only looking for Components and not ComponentRefs also.

We could alternately duplicate line 15 in filters.xslt for each file, but we would then have to get the path and name exactly correct for each file.

Iristyle:
Closing in favor of https://github.com/puppetlabs/puppet_for_the_win/pull/132

PR #459, opened 2015-12-18
(maint) Track the stable branch of cpp-pcp-client

kylog:
Previously puppet-agent#stable was tracking cpp-pcp-client#master,
which led to a little confusion. There's room for discussion about
how we want to handle internal components that do *not* have promoting
pipelines (e.g. branches or tags), and right now cpp-pcp-client is
the only example of such. But for now, this is more consistent with
what puppet-agent does for most other internal components.

### DIFF
--- a/configs/components/cpp-pcp-client.json
+++ b/configs/components/cpp-pcp-client.json
@@ -1,2 +1,2 @@
-{"url": "git@github.com:puppetlabs/cpp-pcp-client.git", "ref": "refs/tags/1.0.1"}
+{"url": "git@github.com:puppetlabs/cpp-pcp-client.git", "ref": "stable"}
 


kylog:
/cc @parisiale and @james-stocks since we bumped into this yesterday.

Also /cc @MikaelSmith and @branan because when leatherman is in puppet-agent, we may want to think about the question I raised in the comment (track branch or tag).

kylog:
Oh also, when we merge stable up to master, we wouldn't want to carry this change up.

MikaelSmith:
Makes sense.

stahnma:
22 days old, still the right thing to do here? If so, plz merge. If not, plz close.

MikaelSmith:
@kylog this still make sense?

kylog:
It made sense a month ago and it will make sense after we release, but *now* we need to PR tags for all the components. So hold off on this for now and we can merge it after we release.

MikaelSmith:
Needs a rebase. And sounds like we're almost ready to do it.

kylog:
Rebased.

MikaelSmith:
This should be ready to merge now.

MikaelSmith:
Actually, waiting until [PA-143](https://tickets.puppetlabs.com/browse/PA-143) is Closed.

PR #460, opened 2015-12-21
(RE-6121) Update to latest openssl

underscorgan:
Update to openssl 1.0.2e.

### DIFF
--- a/configs/platforms/solaris-10-i386.rb
+++ b/configs/platforms/solaris-10-i386.rb
@@ -38,6 +38,8 @@ basedir=default" > /var/tmp/vanagon-noask;
   pkgadd -n -a /var/tmp/vanagon-noask -G -d http://get.opencsw.org/now all;
   /opt/csw/bin/pkgutil -y -i curl rsync gmake pkgconfig ggrep;
   ln -sf /opt/csw/bin/rsync /usr/bin/rsync;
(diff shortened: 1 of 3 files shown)


underscorgan:
Test build (before rebase) http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/239/

configs/platforms/solaris-10-i386.rb
```diff
@@ -38,6 +38,8 @@
   pkgadd -n -a /var/tmp/vanagon-noask -G -d http://get.opencsw.org/now all;
   /opt/csw/bin/pkgutil -y -i curl rsync gmake pkgconfig ggrep;
   ln -sf /opt/csw/bin/rsync /usr/bin/rsync;
+  # RE-6121 openssl 1.0.2e requires functionality not in sytem grep
+  ln -sf /opt/csw/bin/ggrep /usr/bin/grep;
```
mckern: Of course.

mckern:
:+1:

configs/platforms/solaris-10-i386.rb
haus: What about aix? How is the grep there? Or does that work because /opt/freeware/bin is in PATH?

PR #461, opened 2015-12-21
(RE-5733, RE-5734) Build ruby and rubygems on windows

melissa:
This PR mainly updates the windows feature branch to build ruby and individually manage all required rubygems.

This PR also fixes a few assumptions that were incorrect prior to this.

1) Ruby expects the ca cert pem file to be in the openssl prefix directory. There may be a way to specify a custom location during the configure, but I have yet to figure out what that is.
2) We should be defaulting to cygwin make rather than mingw make. We call out to shell in a few places in the makefile, and if we're running the makefile with mingw make, these shell calls don't really do anything, or they just spin. It's also simpler to default to cygwin make and override with mingw make if the need arrises.

### DIFF
--- a/configs/components/ruby.rb
+++ b/configs/components/ruby.rb
@@ -72,6 +72,13 @@ component "ruby" do |pkg, settings, platform|
     special_flags = "--build=#{settings[:platform_triple]}"
   end
 
+  if platform.is_windows?
+    pkg.apply_patch "#{base}/windows_ruby_2.1_update_to_rubygems_2.4.5.patch"
+    pkg.apply_patch "#{base}/windows_fixup_generated_batch_files.patch"
+    pkg.apply_patch "#{base}/windows_remove_DL_deprecated_warning.patch"
+    pkg.apply_patch "#{base}/windows_ruby_2.1_update_to_rubygems_2.4.5.1.patch"
+  end
+
   # Cross-compiles require a hand-built rbconfig from the target system
   if platform.is_solaris? || platform.is_aix?
     pkg.add_source "file://resources/files/rbconfig-#{settings[:platform_triple]}.rb", sum: rbconfig_info[settings[:platform_triple]][:sum]
@@ -88,6 +95,8 @@ component "ruby" do |pkg, settings, platform|
     pkg.build_requires "http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/zlib-devel-1.2.3-4.aix5.2.ppc.rpm"
   elsif platform.is_rpm?
     pkg.build_requires "zlib-devel"
+  elsif platform.is_windows?
+    pkg.build_requires "pl-zlib-#{platform.architecture}"
   end
 
   if platform.is_solaris?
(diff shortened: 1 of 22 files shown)


melissa:
This is dependent on https://github.com/puppetlabs/vanagon/pull/239 for successfully adding the patch

configs/components/ruby.rb
```diff
@@ -108,13 +117,34 @@
...
+    pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "CC" => settings[:cc]
+    pkg.environment "CXX" => settings[:cxx]
+    pkg.environment "LDFLAGS" => settings[:ldflags]
+    pkg.environment "CFLAGS" => settings[:cflags]
+    prefix = platform.convert_to_windows_path(settings[:prefix])
+
+    pkg.environment "MAKE" => platform[:make]
+
+    special_flags = "CPPFLAGS='-DFD_SETSIZE=2048' debugflags=-g"
```
mckern: nitpick: comment pointing out why file descriptor size needs to be changed for windows? Not blocking.

configs/components/rubygem-win32-dir.rb
```diff
@@ -0,0 +1,27 @@
...
+  pkg.version "0.4.9"
+  pkg.md5sum "df14aa01bd6011f4b6332a05e15b7fb8"
+  pkg.url "http://buildsources.delivery.puppetlabs.net/win32-dir-#{pkg.get_version}.gem"
+
+  pkg.build_requires "ruby"
+
+  pkg.environment "PATH" => "#{settings[:bindir]}:$$PATH"
+
+  # Because we are cross-compiling on sparc, we can't use the rubygems we just built.
+  # Instead we use the host gem installation and override GEM_HOME. Yay?
```
mckern: We have this comment as boilerplate in a lot of components. I'm not entirely sure it's applicable anymore. I'd probably snip it out or rewrite it.

mckern:
Components look good; I did not review the patches because that way lies madness. Aside from a comment about errant boilerplate comments in all the new Windows-specific ruby gems, I'm :+1:.

configs/components/rubygem-win32-dir.rb
melissa: These flags are passed in the rubyinstalled build. I figured I'd bring them forward. No reason other then that.

configs/components/ruby.rb
melissa: These flags are passed in the rubyinstalled build. I figured I'd bring them forward. No reason other then that.

configs/components/rubygem-win32-dir.rb
melissa: It's still nice to have an explanation of why this variable is set in the first place, when logically it shouldn't be necessary. If you feel strongly about removing it, I will.

configs/components/rubygem-ffi.rb
```diff
@@ -0,0 +1,35 @@
+component "rubygem-ffi" do |pkg, settings, platform|
+  pkg.version "1.9.6"
+
+  if platform.is_windows?
+    if platform.architecture == "x64"
+      pkg.md5sum "daf0d310a5c498906f94f8c10006fd39"
+      pkg.url "http://buildsources.delivery.puppetlabs.net/ffi-#{pkg.get_version}-x64-mingw32.gem"
+    else
+      pkg.md5sum "5a0b3f3602f65a2b7fe6569a3333a068"
+      pkg.url "http://buildsources.delivery.puppetlabs.net/ffi-#{pkg.get_version}-x86-mingw32.gem"
```
ferventcoder: I thought we were setting up an internal gems repository. Is this really where it would live?

ferventcoder:
When we patch these gems, what process do we have in place where we can have patched versions of these gems? I would prefer that it would be as transparent as possible, like when we are patching other aspects of ruby, instead of just uploading a custom version to the internal gems mirror.

configs/components/rubygem-ffi.rb
melissa: It's where we're installing all our sources from, including other gems (https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/rubygem-deep-merge.rb#L4). We might switch this over later, but I'm not sure. I don't know the long term plan with these. I know Mike's been looking at setting up a rubygem server on nexus

configs/components/rubygem-win32-service.rb
```diff
@@ -0,0 +1,29 @@
...
+    pkg.environment "GEM_HOME" => settings[:gem_home]
+  end
+
+  # PA-25 in order to install gems in a cross-compiled environment we need to
+  # set RUBYLIB to include puppet and hiera, so that their gemspecs can resolve
+  # hiera/version and puppet/version requires. Without this the gem install
+  # will fail by blowing out the stack.
+  pkg.environment "RUBYLIB" => "#{settings[:ruby_vendordir]}:$$RUBYLIB"
+
+  pkg.apply_patch "resources/patches/rubygem-win32-service/PUP-4390-win32-service-0.8.6-restore_win_2003_compat.patch", destination: "#{settings[:gem_home]}/gems/win32-service-#{pkg.get_version}", after: "install"
```
melissa: @ferventcoder re: your question about patching. Does this help?

melissa: So, we'll use the upstream version (from an internal location) and add any additional patches to that gem.

jcoconnor:
@melissa This looks good - have put a few comments/queries in the code.
Have also built this on both x64 and x86 platforms.
:+1:

configs/components/ruby.rb
MikaelSmith: Explanation of why Ruby might do that at https://msdn.microsoft.com/en-us/library/windows/desktop/ms739169(v=vs.85).aspx.

MikaelSmith:
:+1:

configs/components/rubygem-ffi.rb
ferventcoder: :+1:

ferventcoder:
I'm good with the parts I understand. :)

configs/components/rubygem-deep-merge.rb
```diff
@@ -7,9 +7,15 @@
 
   pkg.build_requires "ruby"
 
+  pkg.environment "PATH" => "#{settings[:bindir]}:$$PATH"
+
   # Because we are cross-compiling on sparc, we can't use the rubygems we just built.
   # Instead we use the host gem installation and override GEM_HOME. Yay?
-  pkg.environment "GEM_HOME" => settings[:gem_home]
+  if platform.is_windows?
+    pkg.environment "GEM_HOME" => platform.convert_to_windows_path(settings[:gem_home])
```
haus: why not set GEM_HOME for windows in the project config and do the conditional once instead of doing it for all the gems?

melissa: That's a really good point... I'll get a new PR up with that change

PR #463, opened 2015-12-21
(maint) Track puppet_for_the_win#stable

Iristyle:
- In 418268e a commit was introduced for stable to track stable.
   In 0270b3a that was reverted for some reason, and stable began to
   track puppet_for_the_win#master.  Let's set this back to stable.

 - Note that releases 1.3.2, 1.3.1, 1.3.0, 1.2.7, etc were all
   properly pinned to the appropriate version tag in puppet_for_the_win

### DIFF
--- a/configs/components/windows_puppet.json
+++ b/configs/components/windows_puppet.json
@@ -1,4 +1,4 @@
 {
   "url": "git://github.delivery.puppetlabs.net/puppetlabs-puppet_for_the_win.git",
-  "ref": "origin/master"
+  "ref": "origin/stable"
 }


kylog:
:+1: 

I meant to ask about this the other day but it slipped my mind.

Related: when we merge puppet-agent stable->master, I assume we do *not* want this - correct?

joshcooper:
:+1: At some point in the lead up to puppet-agent 1.3.3, we will want to tag PFTW and pin puppet-agent#stable to that tag. The sooner we get that sorted out (in the next few weeks) the better so that there are fewer things to tweak when we do the actual release.

ferventcoder:
:+1:

ferventcoder:
> Related: when we merge puppet-agent stable->master, I assume we do not want this - correct?

What @joshcooper said... and yes we don't want this :)

Iristyle:
Ok - so @ferventcoder you want to merge this then?

PR #464, opened 2015-12-21
(HUAWEI-2) Add ruby gems for net-netconf and depdencies

ScottGarman:
This adds component definitions for the following gems:

* net-netconf
* net-scp (dependency of net-netconf)
* nokogiri (dependency of net-netconf)
* mini_portile (dependency of nokogiri)

These gems will be included in the Huawei agent package.

### DIFF
--- a/configs/components/rubygem-mini_portile.rb
+++ b/configs/components/rubygem-mini_portile.rb
@@ -1,0 +1,15 @@
+component "rubygem-mini_portile" do |pkg, settings, platform|
+  pkg.version "0.6.2"
+  pkg.md5sum "281cc0d974d3810d1195ad4a863ba5b6"
+  pkg.url "http://buildsources.delivery.puppetlabs.net/mini_portile-#{pkg.get_version}.gem"
+
+  pkg.build_requires "ruby"
+
+  # Because we are cross-compiling on ppc, we can't use the rubygems we just built.
+  # Instead we use the host gem installation and override GEM_HOME. Yay?
+  pkg.environment "GEM_HOME" => settings[:gem_home]
+
+  pkg.install do
+    ["#{settings[:gem_install]} mini_portile-#{pkg.get_version}.gem"]
+  end
+end
--- a/configs/components/rubygem-nokogiri.rb
+++ b/configs/components/rubygem-nokogiri.rb
@@ -1,0 +1,16 @@
+component "rubygem-nokogiri" do |pkg, settings, platform|
+  pkg.version "1.6.6.2"
+  pkg.md5sum "fc9f91534bf93d57b84f625b55732a7c"
+  pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-#{pkg.get_version}.gem"
+
+  pkg.build_requires "ruby"
+  pkg.build_requires "rubygem-mini_portile"
+
+  # Because we are cross-compiling on ppc, we can't use the rubygems we just built.
+  # Instead we use the host gem installation and override GEM_HOME. Yay?
+  pkg.environment "GEM_HOME" => settings[:gem_home]
+
+  pkg.install do
+    ["#{settings[:gem_install]} nokogiri-#{pkg.get_version}.gem -- --use-system-libraries --with-xml2-lib=/opt/puppetlabs/puppet/lib --with-xml2-include=/opt/puppetlabs/puppet/include/libxml2 --with-xslt-lib=/opt/puppetlabs/puppet/lib --with-xslt-include=/opt/puppetlabs/puppet/include/libxslt"]
+  end
+end
--- a/configs/components/rubygem-net-netconf.rb
+++ b/configs/components/rubygem-net-netconf.rb
@@ -1,0 +1,17 @@
+component "rubygem-net-netconf" do |pkg, settings, platform|
+  pkg.version "0.4.3"
+  pkg.md5sum "fa173b0965766a427d8692f6b31c85a4"
+  pkg.url "http://buildsources.delivery.puppetlabs.net/net-netconf-#{pkg.get_version}.gem"
+
+  pkg.build_requires "ruby"
+  pkg.build_requires "rubygem-nokogiri"
+  pkg.build_requires "rubygem-net-scp"
+
+  # Because we are cross-compiling on ppc, we can't use the rubygems we just built.
+  # Instead we use the host gem installation and override GEM_HOME. Yay?
+  pkg.environment "GEM_HOME" => settings[:gem_home]
+
+  pkg.install do
+    ["#{settings[:gem_install]} net-netconf-#{pkg.get_version}.gem"]
+  end
+end
--- a/configs/components/rubygem-net-scp.rb
+++ b/configs/components/rubygem-net-scp.rb
@@ -1,0 +1,16 @@
+component "rubygem-net-scp" do |pkg, settings, platform|
+  pkg.version "1.2.1"
+  pkg.md5sum "abeec1cab9696e02069e74bd3eac8a1b"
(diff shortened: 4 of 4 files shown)


configs/components/rubygem-nokogiri.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  pkg.build_requires "ruby"
+  pkg.build_requires "rubygem-mini_portile"
+
+  # Because we are cross-compiling on ppc, we can't use the rubygems we just built.
+  # Instead we use the host gem installation and override GEM_HOME. Yay?
+  pkg.environment "GEM_HOME" => settings[:gem_home]
+
+  pkg.install do
+    ["#{settings[:gem_install]} nokogiri-#{pkg.get_version}.gem"]
```
haus: Does this download its own libxml and libxslt or use the puppet-agent versions?

ScottGarman: I'm going to have to re-verify this. Good catch.

ScottGarman:
@haus This is kind of odd. The only dynamically linked binaries I'm finding in the nokogiri distro are:

root@huawei:/opt/puppetlabs/puppet/lib/ruby/gems/2.1.0/gems/nokogiri-1.6.6.2# ldd ./ext/nokogiri/nokogiri.so
	linux-vdso32.so.1 (0x00100000)
	libruby.so.2.1 => /opt/puppetlabs/puppet/lib/libruby.so.2.1 (0x6fa74000)
	libm.so.6 => /lib/powerpc-linux-gnu/libm.so.6 (0x6f99e000)
	libdl.so.2 => /lib/powerpc-linux-gnu/libdl.so.2 (0x6f97a000)
	libz.so.1 => /lib/powerpc-linux-gnu/libz.so.1 (0x6f93f000)
	libpthread.so.0 => /lib/powerpc-linux-gnu/libpthread.so.0 (0x6f904000)
	libcrypt.so.1 => /lib/powerpc-linux-gnu/libcrypt.so.1 (0x6f8b3000)
	libc.so.6 => /lib/powerpc-linux-gnu/libc.so.6 (0x6f714000)
	/lib/ld.so.1 (0x206aa000)
root@huawei:/opt/puppetlabs/puppet/lib/ruby/gems/2.1.0/gems/nokogiri-1.6.6.2# 
root@huawei:/opt/puppetlabs/puppet/lib/ruby/gems/2.1.0/gems/nokogiri-1.6.6.2# ldd ./lib/nokogiri/nokogiri.so
	linux-vdso32.so.1 (0x00100000)
	libruby.so.2.1 => /opt/puppetlabs/puppet/lib/libruby.so.2.1 (0x6fa74000)
	libm.so.6 => /lib/powerpc-linux-gnu/libm.so.6 (0x6f99e000)
	libdl.so.2 => /lib/powerpc-linux-gnu/libdl.so.2 (0x6f97a000)
	libz.so.1 => /lib/powerpc-linux-gnu/libz.so.1 (0x6f93f000)
	libpthread.so.0 => /lib/powerpc-linux-gnu/libpthread.so.0 (0x6f904000)
	libcrypt.so.1 => /lib/powerpc-linux-gnu/libcrypt.so.1 (0x6f8b3000)
	libc.so.6 => /lib/powerpc-linux-gnu/libc.so.6 (0x6f714000)
	/lib/ld.so.1 (0x2008f000)
 
As you can see, neither of them link to libxml, regardless of where it's hosted.

ScottGarman:
I've confirmed that according to the nokogiri docs, v1.6.0 and later downloads and statically links in its own libxml/xslt. I assume we really want it dynamically linked to our own libraries, so I'll update this PR with that when ready.

ScottGarman:
Ok, I've updated the nokogiri gem installation to link against our vendored libxml2 and libxslt. Verified by running ldd on the .so:

ldd /opt/puppetlabs/puppet/lib/ruby/gems/2.1.0/extensions/powerpc-linux/2.1.0/nokogiri-1.6.6.2/nokogiri/nokogiri.so 
	linux-vdso32.so.1 (0x00100000)
	libruby.so.2.1 => /opt/puppetlabs/puppet/lib/libruby.so.2.1 (0x6fcb9000)
	libexslt.so.0 => /opt/puppetlabs/puppet/lib/libexslt.so.0 (0x6fc7f000)
	libxslt.so.1 => /opt/puppetlabs/puppet/lib/libxslt.so.1 (0x6fc0a000)
	libxml2.so.2 => /opt/puppetlabs/puppet/lib/libxml2.so.2 (0x6fa02000)
	libpthread.so.0 => /lib/powerpc-linux-gnu/libpthread.so.0 (0x6f9c7000)
	libdl.so.2 => /lib/powerpc-linux-gnu/libdl.so.2 (0x6f9a3000)
	libcrypt.so.1 => /lib/powerpc-linux-gnu/libcrypt.so.1 (0x6f952000)
	libm.so.6 => /lib/powerpc-linux-gnu/libm.so.6 (0x6f87c000)
	libc.so.6 => /lib/powerpc-linux-gnu/libc.so.6 (0x6f6dd000)
	/lib/ld.so.1 (0x201e9000)
	libz.so.1 => /lib/powerpc-linux-gnu/libz.so.1 (0x6f6a2000)

melissa:
Where do these get included? Don't we need to update the project file also?

ScottGarman:
@melissa valid question. For Huawei I have not submitted a PR for a platform file because the eventual goal is to cross-compile it. So this work is a partial step to get included, and can be used not just for the Huawei platform, but potentially other platforms if they want to include nokogiri and/or net-netconf.

melissa:
:+1: cool

stahnma:
I feel like we should just put them in the project file. That way, everybody gets a nokogiri and makes the azure module, et al easier to work with. However, this probably needs a broader discussion. 

Ping @josh @kylo @garethr

ScottGarman:
Sure, but please let's get this PR merged first and I can submit a follow-up to add them to the puppet-agent project file if that's what we choose to do. We need the component defs either way.

configs/components/rubygem-mini_portile.rb
```diff
@@ -0,0 +1,15 @@
+component "rubygem-mini_portile" do |pkg, settings, platform|
+  pkg.version "0.6.2"
+  pkg.md5sum "281cc0d974d3810d1195ad4a863ba5b6"
+  pkg.url "http://buildsources.delivery.puppetlabs.net/mini_portile-#{pkg.get_version}.gem"
+
+  pkg.build_requires "ruby"
+
+  # Because we are cross-compiling on ppc, we can't use the rubygems we just built.
+  # Instead we use the host gem installation and override GEM_HOME. Yay?
+  pkg.environment "GEM_HOME" => settings[:gem_home]
```
stahnma: Don't we only want to do this on platforms where we are cross compiling?

McdonaldSeanp:
So the PR looks good to me pending Mike's last question :+1:

ScottGarman:
@haus does the GEM_HOME environment change break things for non-cross compiled agents? Huawei hasn't reported any issues, so I *think* it may be working fine.

configs/components/rubygem-mini_portile.rb
melissa: This is set for all gems everywhere, so it hasn't broken anything yet. As long as we set https://github.com/puppetlabs/puppet-agent/blob/master/configs/projects/puppet-agent.rb#L25 correctly, it doesn't look like it does anything bad

garethr:
@stahnma I'm generally a fan of gems being installed as needed, as the version required then isn't coupled to the puppet-agent package. However, in the case of nokogiri a) I don't see new APIs or backwards incompatibility often (anecdotal) and b) it makes both the Azure and vSphere module (possibly some of the netdev modules?) easier to install. So in my view useful, low risk but not essential if it causes problems elsewhere.

PR #465, opened 2015-12-22
(maint) Remove duplicate settings

melissa:
We currently have duplicate entires for eos settings. This is confusing
and unnecessary. This commit removes the duplicate entries.


melissa:
@joshcooper I think this was a result of the merge up at https://github.com/puppetlabs/puppet-agent/commit/6ecb7733fb378353c380bda4488f898ef2b23a77

joshcooper:
doh! :+1:

PR #457, opened 2015-12-17
(maint) Build puppet-agent with nuget packages

melissa:

PR #466, opened 2015-12-22
(PA-131) Add leatherman as its own component

branan:
This enables us to un-vendor leatherman from all of our individual
projects. Until those projects are ready to consume an external
leatherman install, this will go unused. We can't remove the vendoring
until we have this in place in P-A, though.

### DIFF
--- a/bin/build-leatherman.ps1
+++ b/bin/build-leatherman.ps1
@@ -1,0 +1,53 @@
+### Set variables from command line
+# $arch => Choose 32 or 64-bit build
+# $cores => Set the number of cores to use for parallel builds
+# $leathermanRef => the git repository to build from
+# $leathermanFork => the git ref to build from
+param (
+[int] $arch=64,
+[int] $cores=2,
+[string] $leathermanRef='origin/master',
+[string] $leathermanFork='git://github.com/puppetlabs/leatherman'
+)
+
+$ErrorActionPreference = 'Stop'
+
+$scriptDirectory = (Split-Path -parent $MyInvocation.MyCommand.Definition);
+. $scriptDirectory\build-helpers.ps1
+. $scriptDirectory\windows-env.ps1
+
+Write-Host "arch=$arch, cores=$cores"
+
+Write-Host "Starting leatherman build"
+
+cd $sourceDir
+
+## Download leatherman and setup build directories
+Invoke-External { git clone $leathermanFork leatherman }
+cd leatherman
+Invoke-External { git checkout $leathermanRef }
+mkdir -Force release
+cd release
+
+## Build leatherman
+$cmake_args = @(
+  '-G',
+  "MinGW Makefiles",
+  "-DCMAKE_TOOLCHAIN_FILE=C:/tools/pl-build-tools/pl-build-toolchain.cmake",
+  "-DBOOST_STATIC=ON",
+  "-DCMAKE_PREFIX_PATH=`"$toolsDir\$curlPkg`"",
+  "-DCMAKE_INSTALL_PREFIX=`"$toolsDir\leatherman`"",
+  "-DCURL_STATIC=ON",
+  ".."
+)
+Invoke-External { cmake $cmake_args }
+Invoke-External { mingw32-make -j $cores }
+Invoke-External { mingw32-make install }
+
+Write-Host "leatherman Build completed."
+## Write out the version that was just built.
+#git describe --long | Out-File -FilePath 'bin/VERSION' -Encoding ASCII -Force
+
+## Test the results.
+Write-Host "Starting Tests"
+Invoke-External { mingw32-make test ARGS=-V }
--- a/configs/components/leatherman.json
+++ b/configs/components/leatherman.json
@@ -1,0 +1,1 @@
+{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "0.3.4"}
--- a/configs/components/leatherman.rb
+++ b/configs/components/leatherman.rb
@@ -1,0 +1,99 @@
+component "leatherman" do |pkg, settings, platform|
+  pkg.load_from_json('configs/components/leatherman.json')
+
+  if platform.is_osx?
+    pkg.build_requires "cmake"
+    pkg.build_requires "boost"
+  elsif platform.name =~ /huaweios/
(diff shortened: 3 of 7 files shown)


branan:
@MikaelSmith @melissa have at it!

configs/components/leatherman.rb
```diff
@@ -0,0 +1,93 @@
...
+    cmake = "/opt/pl-build-tools/bin/cmake"
+  end
+
+  pkg.configure do
+    ["#{cmake} \
+        #{toolchain} \
+        -DCMAKE_VERBOSE_MAKEFILE=ON \
+        -DCMAKE_PREFIX_PATH=#{settings[:prefix]} \
+        -DCMAKE_INSTALL_PREFIX=#{settings[:prefix]} \
+        -DLEATHERMAN_SHARED=TRUE \
```
branan: I still have no super strong feelings over whether we want to go with shared here or not

configs/components/leatherman.json
```diff
@@ -0,0 +1 @@
+{"url": "git://github.com/puppetlasb/leatherman.git", "ref": "0.3.2"}
```
branan: Now that we have a tagged leatherman that works standalone, I've pinned this to that tag

bin/build-leatherman.ps1
```diff
@@ -0,0 +1,54 @@
...
+## Download leatherman and setup build directories
+Invoke-External { git clone $leathermanFork leatherman }
+cd leatherman
+Invoke-External { git checkout $leathermanRef }
+Invoke-External { git submodule update --init --recursive }
+mkdir -Force release
+cd release
+
+## Build leatherman
+$cmake_args = @(
```
MikaelSmith: Going with static leatherman here makes sense. I have a PR we can test eventually for LTH-63.

configs/projects/puppet-agent.rb
```diff
@@ -92,6 +92,7 @@
   proj.component "puppet"
   proj.component "facter"
   proj.component "hiera"
+  proj.component "leatherman"
   proj.component "marionette-collective"
   proj.component "cpp-pcp-client"
   proj.component "pxp-agent"
```
melissa: Isn't leatherman a dependency on cpp-cpc-client and pxp-agent as well? If it is, we should include it as a build_requires in those config files just for the sake of clarity

bin/build-leatherman.ps1
```diff
@@ -0,0 +1,54 @@
...
+
+Write-Host "Starting leatherman build"
+
+cd $sourceDir
+
+## Download leatherman and setup build directories
+Invoke-External { git clone $leathermanFork leatherman }
+cd leatherman
+Invoke-External { git checkout $leathermanRef }
+Invoke-External { git submodule update --init --recursive }
```
melissa: Does leatherman also have submodules?

branan: nope, this is just cargo culting

configs/projects/puppet-agent.rb
branan: Yes, but they're still using their vendored version. Until I get around to unvendoring lth from their, they don't have a require on the system version

branan:
@melissa Do you have a good way to test this across a ton of build targets? I can probably throw together a script to test it all in parallel kicked off from my laptop, but I feel like there must be a Better Way

branan:
@melissa submodule line removed

melissa:
@branan I would suggest testing with http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/ I need to modify it to handle windows builds, but that will happen momentarily

configs/projects/puppet-agent.rb
melissa: Derp, that's what you say in your commit message. I should have read that first...

branan:
Sweet. I've kicked that, and am running an AIX build test from my laptop

configs/projects/puppet-agent.rb
MikaelSmith: It wouldn't hurt to include them, it'd ensure the correct ordering and - as there are no changes to Facter - may avoid needing future changes to puppet-agent to enable using unvendored Leatherman.

configs/components/leatherman.rb
MikaelSmith: This may be potentially helpful for Boost.Log supporting the Windows EventLog, but it hasn't been tested. Otherwise I have no particular arguments either. Maybe compare binary sizes and start-up speed/memory use to see if we get any benefits.

MikaelSmith:
:+1: on this, although it wouldn't hurt to put the dependency in cpp-pcp-client and pxp-agent.

branan:
This is blocked on us either

* Resolving the unit test ordering issue that's causing failures on Squeeze and Lucid
* Adding some workarounds here to skip the test step on those platforms.

I'd rather us fix the tests if we can, obviously :)

branan:
We also need to roll a bugfix release of leatherman - I have one fix in, and I'd like to get a couple  of Michael's changes in too so we're not kicking lth releases out every day next week

configs/components/leatherman.rb
Iristyle: I think the Windows event log situation is definitely worth validating.

MikaelSmith:
https://github.com/puppetlabs/leatherman/pull/123 should resolve Squeeze and Lucid failures.

MikaelSmith:
Needs a rebase.

branan:
I'm kicking off a windows build (since that's where the conflict was) locally to verify it all looks good, then I'll push up my rebase and kick the job

branan:
It looks like the merge of #457 is causing some weirdness for me. Working on it...

branan:
http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/251/ kicked off to test this. Merge conflict will be resolved when I remove my tmp commit for the leatherman-requiring-facter

branan:
OK, everything is looking green in vanagon_generic_job, after a few runs. (http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/253/)

We still need to re-tag leatherman, then I can rip out my extra commits here and we can finally call this thing "done"

branan:
Leatherman re-tagged, and extra commits ripped out. This is ready IMO

MikaelSmith:
:tada:

PR #467, opened 2015-12-23
(PA-100) Update version of libxml2 built in puppet-agent

McdonaldSeanp:
Per http://www.xmlsoft.org/news.html, we should update
libxml2 2.9.3 that will be shipped with puppet-agent
(currently only in master). This commit updates the
md5 sum of and version number for the libxml2
component


McdonaldSeanp:
This was tested in jenkins here: http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/249/

and the 2.9.3 source is on build sources

stahnma:
:+1:

PR #469, opened 2015-12-24
(PA-138) Build native software on windows with secure components

melissa:
Prior to this commit, we were building facter, pxp-agent, and
cpp-pcp-client with an outdated version of openssl and with a ruby that
was linked against the same outdated version of openssl.

Given that we are building everything else with a much more recent
version of openssl, we really really should continue to use that.

This commit also updates the paths that facter searches when linking.
We need to ensure all our dependency directories are included in the
search path, including openssl and ruby. This ensures facter will find
and correctly link against those libraries.

Conflicts:
	bin/build-cpp-pcp-client.ps1
	bin/build-facter.ps1
	bin/windows-toolset.ps1

### DIFF
--- a/bin/build-cpp-pcp-client.ps1
+++ b/bin/build-cpp-pcp-client.ps1
@@ -28,7 +28,7 @@ $cmake_args = @(
   "MinGW Makefiles",
   "-DCMAKE_TOOLCHAIN_FILE=C:/tools/pl-build-tools/pl-build-toolchain.cmake",
   "-DBOOST_STATIC=ON",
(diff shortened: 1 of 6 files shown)


melissa:
Passed with https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/6/

melissa:
Nevermind, this won't make future merge ups to master less painful.

PR #472, opened 2016-01-05
(maint) Set GEM_HOME in project

melissa:
For windows, there is no reason we have GEM_HOME defined in the cygwin
path format. We shouldn't be converting it to a windows style path every
time we need it. Instead, we make sure it's a windows style patch from
the get-go.


McdonaldSeanp:
:+1:

PR #473, opened 2016-01-05
(maint) Remove `env` from Leatherman tests

melissa:
This was originally put here to deal with environment variables on
solaris. We don't actually require it though.

This is currently only causing problems on windows when we deal with
different environments. `env` isn't available in the windows
environment, only in cygwin. So this fails when we need to exclude
everything cygwin from the path.


melissa:
cc/ @branan I tested this on solaris-10-i386 with no problems.

MikaelSmith:
:+1: Looks sane to me.

PR #474, opened 2016-01-05
(RE-6013) Ensure our projects build on windows

melissa:


configs/components/cpp-pcp-client.rb
```diff
@@ -19,6 +23,25 @@
...
     cmake = "/opt/pl-build-tools/i386-pc-solaris2.#{platform.os_version}/bin/cmake"
     toolchain = "-DCMAKE_TOOLCHAIN_FILE=/opt/pl-build-tools/#{settings[:platform_triple]}/pl-build-toolchain.cmake"
+  elsif platform.is_windows?
+    pkg.build_requires "cmake"
+    pkg.build_requires "pl-toolchain-#{platform.architecture}"
+    pkg.build_requires "pl-boost-#{platform.architecture}"
+
+    arch = platform.architecture == "x64" ? "64" : "32"
+    make = "#{platform.drive_root}/tools/mingw#{arch}/bin/mingw32-make"
+    prefix = platform.convert_to_windows_path(settings[:prefix])
```
haus: This should probably happen in the project file, as I presume it applies to most components.

haus:
In general it seems like all of the windows specific path munging that is repeated in each component could be done once in the project config.

configs/components/cpp-pcp-client.rb
melissa: If we specify windows style pathing (`C:/`), the makefile generated will append the current working directory to the path name, so you end up with something like `/home/melissa/puppet-agent/c:/Program Files/Puppet Labs/Puppet`

haus: why would project vs component affect how this method behaves?

melissa: We could create a new setting in the project file, `windows_prefix = platform.convert_to_windows_path(settings[:prefix])`, but we can't have vanagon try to create the directory or any directories that inherit from it. Is that what you're thinking? We still have to do the special prefix casing though

haus: I was thinking more along the lines of `if windows? then proj.add_setting(:prefix, convert_to_windows_path(proj.prefix))` in puppet-agent.rb.

melissa: https://github.com/puppetlabs/puppet-agent/blob/master/configs/projects/puppet-agent.rb#L140 will cause the makefile to not populate correctly :(

haus: that sounds like a pretty sharp edge for a vanagon user to cut themself on. is it ticketed so that the conditionals can get cleaned up someday?

melissa: https://tickets.puppetlabs.com/browse/RE-6215

haus: :+1:

McdonaldSeanp:
This all seems pretty legitimate to me :+1:

MikaelSmith:
In the vein of @haus's comment, I've done some cleanup by moving common component logic into the project file for https://github.com/puppetlabs/puppet-client-tools-vanagon/pull/14. You also don't need to set most of the environment variables for CMake projects, they really just need mingw at the front of the PATH.

configs/components/cpp-pcp-client.rb
```diff
@@ -2,9 +2,13 @@
   pkg.load_from_json('configs/components/cpp-pcp-client.json')
   cmake = "/opt/pl-build-tools/bin/cmake"
   toolchain = "-DCMAKE_TOOLCHAIN_FILE=/opt/pl-build-tools/pl-build-toolchain.cmake"
-  pkg.environment "PATH" => "#{settings[:bindir]}:/opt/pl-build-tools/bin:$$PATH"
+  pkg.environment "PATH" => "#{settings[:bindir]}:#{settings[:tools_root]}/bin:$$PATH"
```
MikaelSmith: Not sure what you need to pick up from `#{settings[:tools_root]}/bin`.

configs/components/cpp-pcp-client.rb
MikaelSmith: The only other issue with using `C:/` everywhere is that it can't go into PATH under Cygwin (but that could be the special case and would be much less prevalent).

configs/components/cpp-pcp-client.rb
melissa: For *nix, this is where cmake is

melissa:
@MikaelSmith I totally agree that we'll need to refactor some of this. That's true of some of the code that's already been merged in. Some of the refactoring I'd like to do is dependent on work that's currently in progress.

configs/components/cpp-pcp-client.rb
MikaelSmith: We use the full path for cmake at line 3.

melissa: I'm not sure. This is carried forward from the previous setup. I'd rather keep it around then take it out at this point in time.

MikaelSmith:
I'm :+1:

configs/components/facter.rb
```diff
@@ -56,6 +56,12 @@
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-boost-1.58.0-1.aix#{platform.os_version}.ppc.rpm"
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-yaml-cpp-0.5.1-1.aix#{platform.os_version}.ppc.rpm"
     pkg.build_requires "runtime"
+  elsif platform.is_windows?
+    pkg.build_requires "cmake"
+    pkg.build_requires "pl-toolchain-#{platform.architecture}"
+    pkg.build_requires "pl-boost-#{platform.architecture}"
+    pkg.build_requires "pl-yaml-cpp-#{platform.architecture}"
+    pkg.build_requires "runtime"
```
ferventcoder: :+1:

ferventcoder:
I can't tell you whether the arguments for building C++ are correct or not, but the rest looks fine.

configs/components/pxp-agent.rb
```diff
@@ -30,10 +55,10 @@
...
       #{toolchain} \
           -DCMAKE_VERBOSE_MAKEFILE=ON \
-          -DCMAKE_PREFIX_PATH=#{settings[:prefix]} \
-          -DCMAKE_INSTALL_PREFIX=#{settings[:prefix]} \
-          -DCMAKE_SYSTEM_PREFIX_PATH=#{settings[:prefix]} \
-          -DMODULES_INSTALL_PATH=#{File.join(settings[:install_root], 'pxp-agent', 'modules')} \
+          -DCMAKE_PREFIX_PATH=#{prefix} \
+          -DCMAKE_INSTALL_PREFIX=#{prefix} \
+          -DCMAKE_SYSTEM_PREFIX_PATH=#{prefix} \
+          -DMODULES_INSTALL_PATH=#{File.join(install_root, 'pxp-agent', 'modules')} \
```
MikaelSmith: This should probably use `C:/`. Is it?

configs/components/pxp-agent.rb
```diff
@@ -41,16 +66,21 @@
...
 
   pkg.install do
-    ["#{platform[:make]} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
+    ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
   end
 
+  # Due to windows, when we run the configure, we need this variable to be set
+  # in accordance with windows pathing. However, when we create the directory
+  # below, we need it in accordance with cygwin pathing.
+  install_root = settings[:prefix] if platform.is_windows?
```
MikaelSmith: Why does it use `settings[:prefix]` here? I'm not sure I understand what's happening.

configs/components/pxp-agent.rb
melissa: Oh dang, good catch. I was setting `install_root` for windows, but it looks like that was lost.

configs/components/pxp-agent.rb
melissa: This is to make sure we go back to `/cygdrive/c` rather then `C:`. For windows purposes, install_root is equivalent to prefix

MikaelSmith: That doesn't appear to be true: https://github.com/puppetlabs/puppet-agent/blob/feature/master/windows-vanagon-development/configs/projects/puppet-agent.rb#L10-L12

melissa: https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md Ah, this is why we're setting install_root to prefix on windows. We're installing pxp-agent to C:/Program Files/Puppet Labs/Puppet on windows and /opt/puppetlabs on *nix. So, that's why this is set this way

configs/components/pxp-agent.rb
melissa: Fixed

configs/components/pxp-agent.rb
```diff
@@ -19,6 +25,26 @@
...
     special_flags = "-DCMAKE_CXX_FLAGS_RELEASE='-O2 -DNDEBUG'"
+  elsif platform.is_windows?
+    pkg.build_requires "cmake"
+    pkg.build_requires "pl-toolchain-#{platform.architecture}"
+    pkg.build_requires "pl-boost-#{platform.architecture}"
+
+    arch = platform.architecture == "x64" ? "64" : "32"
+    make = "#{platform.drive_root}/tools/mingw#{arch}/bin/mingw32-make"
+    prefix = platform.convert_to_windows_path(settings[:prefix])
+    install_root = platform.convert_to_windows_path(settings[:prefix])
```
MikaelSmith: Shouldn't this be `:install_root`?

configs/components/pxp-agent.rb
MikaelSmith: Can you add a comment to that effect?

configs/components/cpp-pcp-client.rb
```diff
@@ -1,10 +1,14 @@
...
   pkg.load_from_json('configs/components/cpp-pcp-client.json')
+
   cmake = "/opt/pl-build-tools/bin/cmake"
   toolchain = "-DCMAKE_TOOLCHAIN_FILE=/opt/pl-build-tools/pl-build-toolchain.cmake"
-  pkg.environment "PATH" => "#{settings[:bindir]}:/opt/pl-build-tools/bin:$$PATH"
+  make = platform[:make]
+
+  unless platform.is_windows?
+    pkg.environment "PATH" => "#{settings[:bindir]}:/opt/pl-build-tools/bin:$$PATH"
+  end
```
ferventcoder: No windows path overloads?

configs/components/facter.rb
```diff
@@ -134,6 +144,13 @@
 
     # FACT-1156: If we build with -O3, solaris segfaults due to something in std::vector
     special_flags = "-DCMAKE_CXX_FLAGS_RELEASE='-O2 -DNDEBUG'"
+  elsif platform.is_windows?
+    make = "#{settings[:gcc_bindir]}/mingw32-make"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/Wbem:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0:/cygdrive/c/pstools"
```
ferventcoder: Is there a way to append to the current path versus overriding it?

melissa: Yeah, that would be `$$PATH`. We don't get that luxury here though. Since we're building for MinGW Makefiles in a cygwin environment, we can't have cygwin sh.exe on the path. Since it's there by default in the cygwin environment, we need to overwrite the entire path :( I guess I could look at just removing the cygwin specific paths, but that seemed too complicated to do

configs/components/cpp-pcp-client.rb
melissa: What do you mean? I'm setting path for windows on line 32, but I can move it here if that's easier to read?

configs/components/facter.rb
MikaelSmith: The path could probably be less exhaustive.

melissa: Yeah, probably... I just copied the path from my system and removed the cygwin directories...

melissa:
Also, I just added ruby projects to this PR because I was bored...

configs/components/facter.rb
MikaelSmith: You should be able to safely strip it down to `$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows`. Wbem and pstools aren't needed. I guess it makes sense to keep `/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0` in case something calls powershell.

MikaelSmith: Maybe generate this in the project file, since all CMake projects use the same PATH?

PR #475, opened 2016-01-05
(PA-140) Specify builders for AIX targets

stahnma:
This commit enables the vanagon hardware engine for AIX builds by
specifying a specific builder to use. The hardware engine produces a
lock so that other builds/tests do not trample on the instance being
used during the build. Right now we're only specifying one host, but
this could easily be multiple if we had the builders available.


McdonaldSeanp:
:+1:

PR #478, opened 2016-01-06
(PA-40) Remove unneeded binaries from dmidecode

stahnma:
We build our own dmidecode. This ships 3 items in the bin directory, but
we only need dmidecode for the puppet-agent package. So, this removes
vpddecode and biosdecode.


stahnma:
@shrug since you were curious, you can build just dmdecode,but when you run the install target, it builds the rest. You can't just install dmidecode using their Makefile.

shrug:
Delightful

:+1:

PR #479, opened 2016-01-07
(RE-6194) Use updated ruby builds

underscorgan:
Ruby was rebuilt with a newer version of openssl. Use those for windows
puppet-agent builds.


shrug:
:+1:

PR #480, opened 2016-01-08
(PA-142) Add ubuntu 15.10 to foss whitelist

joshcooper:
Add ubuntu 15.10 to foss whitelist for shipping to nightlies.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -28,6 +28,8 @@ foss_platforms:
   - ubuntu-14.04-i386
   - ubuntu-15.04-amd64
   - ubuntu-15.04-i386
+  - ubuntu-15.10-amd64
+  - ubuntu-15.10-i386
   - windows-2012-x86
   - windows-2012-x64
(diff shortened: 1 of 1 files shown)


roidelapluie:
Is it also needed to update "deb_targets" in the same file?

joshcooper:
good question @roidelapluie, @haus @shrug thoughts?

joshcooper:
ping @stahnma

stahnma:
@underscorgan @melissa what did we decide on this?

melissa:
I honestly have no idea what those targets are used for. The only reason they should be needed are the promotion pipelines, but I would think the change here takes care of that. In light of my uncertainty, I'd say we add it just to be safe.

haus:
deb_targets is used for PE ship and promotion, but as 15.10 is not an LTS it is not required (https://github.com/puppetlabs/packaging/blob/334210e1c2f4a9ae983f801222230fbc88fd73bc/tasks/pe_ship.rake#L93).

PR #485, opened 2016-01-11
(maint) Remove obsolete definitions for huaweios

ScottGarman:
This removes obsolete code referring to HuaweiOS when it was targeted for
WRL/RPM.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -27,7 +27,7 @@ component 'augeas' do |pkg, settings, platform|
 
   if platform.is_rpm? && !platform.is_aix?
     pkg.build_requires 'readline-devel'
-    if platform.is_cisco_wrlinux? || platform.is_huaweios?
+    if platform.is_cisco_wrlinux?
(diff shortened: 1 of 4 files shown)


mckern:
@ScottGarman are there any packages we should circle around and clean up as well or are we hanging on to those as historical artifacts?

:+1: on this.

shermdog:
Looks like this should also be removed as part of this PR:  https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/leatherman.rb#L7-L10

stahnma:
no code is best code

ScottGarman:
@shermdog that code is in puppet-agent master, and this PR is targeted at stable, so I'll have to create a separate PR to cull that bit.

@mckern I want to keep the WLR artifacts around for just a bit longer; demos that Huawei are doing could need to fall back to the WRL based environment until we've released the Debian based one.

mckern:
@ScottGarman that makes sense. "Historical artifacts" it is!

PR #486, opened 2016-01-12
(RE-6143) Use ruby 2.1.8 with windows

underscorgan:
To address CVE-2015-7551 we need to update from ruby 2.1.7 to ruby
2.1.8.


underscorgan:
Manually built for x86 and x64. Ruby version was correct. Was able to run facter, puppet, and hiera.

melissa:
:+1:

PR #487, opened 2016-01-13
(maint) Update paths to reduce windows specific variation

melissa:
This pull request updates the current windows work in the windows feature branch to reduce the windows specific branching we have, and also to conform with recent updates to vanagon.

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -6,9 +6,6 @@ component 'curl' do |pkg, settings, platform|
   pkg.build_requires "openssl"
   pkg.build_requires "puppet-ca-bundle"
 
-  prefix = settings[:prefix]
-  pem_file = "#{settings[:prefix]}/ssl/cert.pem"
-
   if platform.is_windows?
     pkg.build_requires "pl-zlib-#{platform.architecture}"
     pkg.build_requires "runtime"
@@ -17,19 +14,16 @@ component 'curl' do |pkg, settings, platform|
     pkg.environment "CYGWIN" => settings[:cygwin]
     pkg.environment "CC" => settings[:cc]
     pkg.environment "CXX" => settings[:cxx]
-
-    prefix = platform.convert_to_windows_path(settings[:prefix])
-    pem_file = platform.convert_to_windows_path("#{settings[:prefix]}/ssl/cert.pem")
   end
 
   pkg.configure do
     ["LDFLAGS='#{settings[:ldflags]}' \
-     ./configure --prefix=#{prefix} \
-        --with-ssl=#{prefix} \
+     ./configure --prefix=#{settings[:prefix]} \
+        --with-ssl=#{settings[:prefix]} \
         --enable-threaded-resolver \
         --disable-ldap \
         --disable-ldaps \
-        --with-ca-bundle=#{pem_file} \
+        --with-ca-bundle=#{settings[:prefix]}/ssl/cert.pem \
         #{settings[:host]}"]
   end
 
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -7,13 +7,10 @@ component 'curl' do |pkg, settings, platform|
   pkg.build_requires "puppet-ca-bundle"
 
   if platform.is_windows?
-    pkg.build_requires "pl-zlib-#{platform.architecture}"
     pkg.build_requires "runtime"
 
-    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
-    pkg.environment "CC" => settings[:cc]
-    pkg.environment "CXX" => settings[:cxx]
   end
 
   pkg.configure do
--- a/configs/components/openssl.rb
+++ b/configs/components/openssl.rb
@@ -60,8 +60,6 @@ component "openssl" do |pkg, settings, platform|
     pkg.build_requires "runtime"
   end
 
-  prefix = settings[:prefix]
-
   if platform.is_osx?
     pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH:/usr/local/bin"
     target = 'darwin64-x86_64-cc'
@@ -86,12 +84,11 @@ component "openssl" do |pkg, settings, platform|
     pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
   elsif platform.is_windows?
     target = platform.architecture == "x64" ? "mingw64" : "mingw"
-    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
     pkg.environment "CC" => settings[:cc]
     pkg.environment "CXX" => settings[:cxx]
     pkg.environment "MAKE" => platform[:make]
-    prefix = platform.convert_to_windows_path(settings[:prefix])
     cflags = settings[:cflags]
     ldflags = settings[:ldflags]
   else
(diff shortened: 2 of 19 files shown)


configs/components/curl.rb
```diff
@@ -6,30 +6,21 @@
...
-  pem_file = "#{settings[:prefix]}/ssl/cert.pem"
-
   if platform.is_windows?
-    pkg.build_requires "pl-zlib-#{platform.architecture}"
     pkg.build_requires "runtime"
 
-    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
-    pkg.environment "CC" => settings[:cc]
```
McdonaldSeanp: There is no longer a need to set the compiler environment? (i'm mostly just curious)

melissa: Yeah, it builds fine. Notice the compiler is the first thing on the path on line 12

McdonaldSeanp: Ah, I see

McdonaldSeanp:
:+1: Seems pretty legitimate to me

underscorgan:
@melissa has this been tested?

underscorgan:
assuming tests work, :100:

configs/components/openssl.rb
```diff
@@ -1,7 +1,28 @@
 component "openssl" do |pkg, settings, platform|
   pkg.version "1.0.2e"
-  pkg.md5sum "5262bfa25b60ed9de9f28d5d52d77fc5"
-  pkg.url "http://buildsources.delivery.puppetlabs.net/openssl-#{pkg.get_version}.tar.gz"
+  if platform.is_windows?
```
MikaelSmith: This is still really weird to me. I wish I knew why it changed. Did it start happening with 1.0.2e?

configs/components/openssl.rb
```diff
@@ -65,12 +84,9 @@
     pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
   elsif platform.is_windows?
     target = platform.architecture == "x64" ? "mingw64" : "mingw"
-    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
     pkg.environment "CC" => settings[:cc]
```
MikaelSmith: Do we still need CC here? PATH should take care of it here as well.

configs/components/ruby.rb
```diff
@@ -113,22 +113,16 @@
...
   if platform.is_windows?
     pkg.build_requires "pl-gdbm-#{platform.architecture}"
     pkg.build_requires "pl-iconv-#{platform.architecture}"
     pkg.build_requires "pl-libffi-#{platform.architecture}"
     pkg.build_requires "pl-pdcurses-#{platform.architecture}"
-    pkg.environment "PATH" => "#{settings[:gcc_bindir]}:#{settings[:tools_root]}/bin:#{settings[:bindir]}:$$PATH"
+
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:bindir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
     pkg.environment "CC" => settings[:cc]
```
MikaelSmith: Do we need CC?

configs/components/openssl.rb
melissa: Nope, it started when I switched all the paths from `/cygdrive/c` to `C:`

configs/components/openssl.rb
melissa: Yeah, I think we do. I tried it without CC, but got failures when some stuff later in the build couldn't find gcc :(

configs/projects/puppet-agent.rb
```diff
@@ -71,7 +70,8 @@
     end
   elsif platform.is_windows?
     # For windows, we need to ensure we are building for mingw not cygwin
-    host = "--host #{platform.platform_triple}"
+    platform_triple = platform.platform_triple
+    host = "--host #{platform_triple}"
```
MikaelSmith: Why change this line? It looked fine before.

configs/components/ruby.rb
melissa: I didn't test removing it for ruby. Mainly because other platforms are also setting it. I can test to see if ruby builds without it though

configs/components/curl.rb
melissa: As we found out with some of the other projects though, this logic doesn't always hold, so it's always a good question to ask

configs/projects/puppet-agent.rb
```diff
@@ -106,17 +106,17 @@
   # Platform specific
   if platform.is_windows?
     arch = platform.architecture == "x64" ? "64" : "32"
-    proj.setting(:cflags, "-IC:/tools/pl-build-tools/include -IC:/tools/mingw#{arch}/include -I#{platform.convert_to_windows_path(proj.includedir)}")
-    proj.setting(:ldflags, "-LC:/tools/pl-build-tools/lib -LC:/tools/mingw#{arch}/lib -L#{platform.convert_to_windows_path(proj.libdir)}")
-    proj.setting(:gcc_bindir, "#{platform.drive_root}/tools/mingw#{arch}/bin")
+    proj.setting(:gcc_root, "C:/tools/mingw#{arch}")
+    proj.setting(:gcc_bindir, "#{proj.gcc_root}/bin")
+    proj.setting(:cc, "#{proj.gcc_bindir}/gcc")
+    proj.setting(:cxx, "#{proj.gcc_bindir}/g++")
```
MikaelSmith: Do we need this? I don't think we use it anywhere anymore.

MikaelSmith:
\o/

configs/projects/puppet-agent.rb
melissa: Yeah, I figured it would be better to be consistent with the other projects. This ensures we have `settings[:platform_tripple] available, and don't have to reach out to 'platform.platform_triple`

configs/components/openssl.rb
MikaelSmith: Then it must have to do with how unpacking is done...

configs/projects/puppet-agent.rb
melissa: But, it was working before, so it's definitely not strictly necessary. I'm not really sure why it was working before...

configs/components/ruby.rb
MikaelSmith: Not that important then, seems fine to leave it.

configs/components/curl.rb
MikaelSmith: Safe with CMake, may not be safe with some autotools-based builds.

configs/projects/puppet-agent.rb
melissa: We're definitely using `settings[:cc]` in this PR (see openssl). We might not need CXX, but I feel like it's nice to keep it around just in case

MikaelSmith: CMake picks it up from PATH, doesn't need CXX specified explicitly. But it doesn't hurt, so up to you.

configs/projects/puppet-agent.rb
melissa: Oh, wait, so `host` is and was being set correctly. However, if we ever want to reference `platform_triple` in the same way solaris and aix are accessing it (i.e., `settings[:platform_triple]`), then this is how we do that. We're not currently using that variable anywhere other then `settings[:host]`. It seemed cleaner to set it now rather then wait until we need it.

configs/components/openssl.rb
melissa: Apparently I was wrong. I just tested this again, and it was fine without it...

configs/projects/puppet-agent.rb
melissa: We probably don't need to set these at all, but I like having them around just in case... What do other people think? We're not using `:cc` or `:cxx` anywhere currently. Should I take them out? We can always add them back in if we decide we don't need them

melissa: kk, I removed them.

MikaelSmith:
So fresh and so clean

melissa:
Tested, it definitely builds.

PR #490, opened 2016-01-15
(PE-13590) Update pl-gcc build to fix solaris 10 linking

shrug:
The version of pl-gcc that is currently used in the solaris 10
build does not properly set the library rpath, leading to issues
if the user has an incompatible build of GCC in their default system
library path.

This updates to a more recent build that sets the rpath correctly.
The issue was fixed in the pl-build-tools repo a while back, but
puppet-agent was never updated to use the newer build.

The version of GCC itself does not changed, just the linker flags passed
to the build.


melissa:
:+1:

PR #492, opened 2016-01-17
ppc64le support

pyther:
With this change I can successfully build puppet-agent on RHEL 7.2 ppc64le.

I suspect there is a better way to implement the changes libxslt.rb and virt-what.rb. I'll gladly update this patch set.

I plan on submitting upstream bug reports requesting config.guess gets updated in virt-what and libxslt.

### DIFF
--- a/configs/platforms/el-7-ppc64le.rb
+++ b/configs/platforms/el-7-ppc64le.rb
@@ -1,0 +1,8 @@
+platform "el-7-ppc64le" do |plat|
+  plat.servicedir "/usr/lib/systemd/system"
+  plat.defaultdir "/etc/sysconfig"
+  plat.servicetype "systemd"
+
+  plat.provision_with "yum install --assumeyes autoconf automake createrepo rsync gcc make rpmdevtools rpm-libs yum-utils rpm-sign;"
+  plat.install_build_dependencies_with "yum install --assumeyes"
+end
--- a/configs/components/libxslt.rb
+++ b/configs/components/libxslt.rb
@@ -25,8 +25,14 @@ component "libxslt" do |pkg, settings, platform|
     pkg.build_requires "pl-gcc-#{platform.architecture}"
   end
 
+  if platform.is_linux?
+    if platform.architecture =~ /ppc64le$/
+      target = 'powerpc64le-unknown-linux-gnu'
+    end
+  end
+
   pkg.configure do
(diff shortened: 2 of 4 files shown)


stahnma:
@pyther thanks! We'll be checking this out a bit. We don't have a system that is ppcle to test right now, so we'll basically just make sure nothing else broke with this changeset.

stahnma:
Tested building on 4 linux platforms, 2 Solaris SPARCS w/ cross compiling and nothing broke. :+1:

configs/platforms/el-7-ppc64le.rb
```diff
@@ -0,0 +1,9 @@
+platform "el-7-ppc64le" do |plat|
+  plat.servicedir "/usr/lib/systemd/system"
+  plat.defaultdir "/etc/sysconfig"
+  plat.servicetype "systemd"
+
+  plat.provision_with "yum install --assumeyes autoconf automake createrepo rsync gcc make rpmdevtools rpm-libs yum-utils rpm-sign;echo '[pl-build-tools]\nname=pl-build-tools\ngpgcheck=0\nbaseurl=http://pl-build-tools.delivery.puppetlabs.net/yum/el/6/$basearch' > /etc/yum.repos.d/pl-build-tools.repo"
```
kylog: Is the `6` in `yum/el/6` correct given this is for el7?

configs/platforms/el-7-ppc64le.rb
```diff
@@ -0,0 +1,9 @@
...
+  plat.install_build_dependencies_with "yum install --assumeyes"
+  plat.vmpooler_template "centos-7-ppc64le"
```
kylog: Is this an actual vmpooler template?

stahnma: No. We could remove that.

configs/platforms/el-7-ppc64le.rb
stahnma: No, but that was copied from our EL7 definition :/

kylog: Hah! That's funny. Well, glad, I noticed :)

pyther:
I removed `plat.vmpooler_template` option and the yum repo definition as the el-7-s390.rb template does not include that.

stahnma:
:+1:  still from me.

kylog:
:+1: Thanks @pyther !

PR #494, opened 2016-01-19
(PCP-209) Remove leatherman git submodule from pxp-agent

parisiale:
With this commit, we add leatherman's path to cpp-pcp-client's
CMAKE_PREFIX_PATH and we require leatherman's package when building
pxp-agent.

Those changes are required after removing leatherman's git submodule
from pxp-agent. Note that cpp-pcp-client's submodule has been removed as
well but no change in puppet-agent's build scripts should be necessary
since cpp-pcp-client dependency is managed directly.

### DIFF
--- a/bin/build-pxp-agent.ps1
+++ b/bin/build-pxp-agent.ps1
@@ -32,7 +32,7 @@ $cmake_args = @(
   "-DCMAKE_TOOLCHAIN_FILE=C:/tools/pl-build-tools/pl-build-toolchain.cmake",
   "-DBOOST_STATIC=ON",
   "-DCMAKE_INSTALL_PREFIX=`"$sourceDir`"",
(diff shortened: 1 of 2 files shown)


MikaelSmith:
:+1:

MikaelSmith:
Shouldn't pxp-agent have a dependency on the cpp-pcp-client component? @melissa

melissa:
Does pxp-agent require cpp-pcp-client to build? If so, then yes

MikaelSmith:
Looks like it was missed. @parisiale mind throwing in an extra maint commit for `pkg.build_requires 'cpp-pcp-client'`?

melissa:
:+1: nice catch. Looks good to me a long as everything actually builds

MikaelSmith:
Builds for me.

PR #504, opened 2016-01-22
Maint/vanagon for the win/puppet specifications

melissa:


configs/components/facter.rb
```diff
@@ -197,11 +197,6 @@
 
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
 
-  if platform.is_windows?
-    pkg.directory File.join(settings[:sysconfdir], "facter", "facts.d")
-  else
-    pkg.link "#{settings[:bindir]}/facter", "#{settings[:link_bindir]}/facter"
-    pkg.directory File.join('/opt/puppetlabs', 'facter')
```
underscorgan: We don't need these two lines for non-windows anymore?

melissa: On non-windows `settings[:install_root]` corresponds to `/opt/puppetlabs`. I figured it'd be easier to use the variable rather than hard-coding `/opt/puppetlabs`. That also lets us use that call on windows rather then having a similar-but-different facts.d directory defined

melissa: Oh, but the two with `pkg.directory File.join('/opt/puppetlabs', 'facter')` and `pkg.directory File.join('/opt/puppetlabs', 'facter', 'facts.d')` were unnecessarily redundant and `pkg.link "#{settings[:bindir]}/facter", "#{settings[:link_bindir]}/facter"` is still getting called on non-windows

underscorgan: ahhh, ok, cool. Thanks for the explanation :)

underscorgan:
this is tested for windows + some non-windows?

melissa:
Yup, but the only non-windows platform I've tested it on is el-7

underscorgan:
:+1:

configs/components/facter.rb
melissa: :+1:

MikaelSmith:
:+1:

PR #505, opened 2016-01-25
(DO NOT MERGE) (maint) Add a rake task to verify tagged components

joshcooper:
This commit adds a `release:tag_check` rake task to verify components
are tagged correctly. You'll want to pull the latest puppet-agent before
running the task:

    $ git pull upstream stable
    $ bundle exec rake release:tag_check

If a component refers to a non-existent tag, an error is reported:

    hiera 3.0.7:                                                      [ FAIL ]
      Tag doesn't exist

If a component's tag is not in the standard `refs/tags/1.2.3` form, an
error is reported:

    puppet 4.3.2:                                                     [ FAIL ]
      Tag '4.3.2' is not of the form 'refs/tags/4.3.2'

If a component is not pinned to a tag, an error is reported:

    pxp-agent stable:                                                 [ FAIL ]
      Tag 'stable' is invalid

If a component's tag points to a commit and the commit is not the head
of the stable branch, a warning is reported, along with the SHAs of the
tag and branch:

    cpp-pcp-client 1.0.1:                                             [ WARN ]
      Additional commits on branch 'stable'
        tag    : 9c4db665c5 (1.0.1)
          └->  : 0a90ea7849
        branch : bee31d1b81 (stable)

If a component is missing a URL, an error is reported:

    nssm:                                                             [ FAIL ]
      URL is missing

If a component has a URL, but no `ref`, such as nssm, then the component
is skipped:

    nssm <none>:                                                      [ SKIP ]

Otherwise, if a component's tag points to the head of the branch, then
OK is printed:

    facter 3.1.4:                                                     [ OK  ]

Note some repos don't follow the stable branching scheme, e.g. mco uses
2.8.x, so the rake task compares against the equivalent "stable" branch
for those repos.

The puppet-win32-ruby repo is special since it is effectively two
components, for x86 and x64 ruby, and the rake task will check both.

Github will rate limit unauthenticated requests, so you may need to
create a personal github access token, and set the `GITHUB_AUTH_TOKEN`
environment variable to the value of the token.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,6 +5,8 @@ begin
 rescue LoadError
 end
 
+load File.join(RAKE_ROOT, 'tasks', 'tag-checker.rake')
+
 build_defs_file = File.join(RAKE_ROOT, 'ext', 'build_defaults.yaml')
 if File.exist?(build_defs_file)
   begin
--- a/tasks/tag-checker.rake
+++ b/tasks/tag-checker.rake
@@ -1,0 +1,216 @@
+# -*- coding: utf-8 -*-
+require 'json'
+require 'net/http'
+
+TAGS_REGEX = /^refs\/tags\/([A-Za-z\d.-]+)$/
+
+GREEN   = "\e[0;32m"
+YELLOW  = "\e[0;33m"
+HRED    = "\e[1;31m"
+RESET   = "\e[0m"
+
+STATUSES = {
+  :ok   => "[  #{GREEN}OK#{RESET}  ]",
+  :fail => "[ #{HRED}FAIL#{RESET} ]",
+  :skip => "[ #{GREEN}SKIP#{RESET} ]",
+  :warn => "[ #{YELLOW}WARN#{RESET} ]"
+}
+
+# Print a status line of the form:
+#  <name> <version>:                                    [  OK  ]
+#
+# If <msg> is provided, writes it indented on a newline
+def print_status(name, version, status, msg = nil)
+  prefix = "#{name} #{version}: ".ljust(66)
+  puts prefix + STATUSES[status]
+  puts "  #{msg}" if msg
+end
+
+# Wrap github API calls
+#
+class Github
+  include Singleton
+
+  def initialize
+    @server = 'api.github.com'
+    @port = 443
+  end
+
+  # Get the SHA that ref points to
+  def get_sha(repo, ref)
+    object = get_json("repos/puppetlabs/#{repo}/git/#{ref}")
+    object ? object['object']['sha'] : nil
+  end
+
+  # Make a github HTTP request, returning the response as JSON.
+  def get_json(path)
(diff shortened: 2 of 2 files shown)


joshcooper:
Running this against what is effectively puppet-agent 1.3.4 shows that some components have additional commits beyond the tags puppet-agent is using:

```
$ bundle exec rake release:tag_check
cpp-pcp-client 1.0.1:                                             [ WARN ]
  Additional commits on branch 'stable':
    tag    : 9c4db665c5 (1.0.1)
      └->  : 0a90ea7849
    branch : bee31d1b81 (stable)
facter 3.1.4:                                                     [  OK  ]
hiera 3.0.6:                                                      [  OK  ]
mco 2.8.7:                                                        [  OK  ]
nssm <none>:                                                      [ SKIP ]
puppet-ca 1.0.7:                                                  [ WARN ]
  Additional commits on branch 'master':
    tag    : ab44d6c82a (1.0.7)
      └->  : d46a3c87cf
    branch : a1a7bce0d8 (master)
puppet 4.3.2:                                                     [  OK  ]
pxp-agent 1.0.2:                                                  [ WARN ]
  Additional commits on branch 'stable':
    tag    : 216d7751e9 (1.0.2)
      └->  : 0c40d11c3c
    branch : 1b12224ca6 (stable)
windows_puppet 4.3.2:                                             [  OK  ]
windows_ruby (x86) 2.1.8.0-x86:                                   [  OK  ]
windows_ruby (x64) 2.1.8.0-x64:                                   [  OK  ]
```

I believe pxp-agent and cpp-pcp-client are ok, and that the puppet-ca changes post 1.0.7 are not necessary for release. But we should sort this out after the release.

/cc @McdonaldSeanp @khussey @barlowst @kylog

mckern:
First thoughts are that this is a lot of Rake to add to this repo. We don't use a `tasks/` directory for this repo, and we (Release Engineering) generally have been trying to move away from special-casing new utilities in Rakefiles. Is this functionality that could be better integrated into Vanagon itself or into the library in the Packaging repo? It looks like it's most of a library already.

McdonaldSeanp:
I agree with @mckern here, this should likely go to either vanagon or the packaging repo. My vote is that since this is pretty vanagon specific it should go to there.

stahnma:
@McdonaldSeanp can you pull this into a vanagon PR? This functionality is great.

joshcooper:
I'm happy if you want to take this over and I'll close this out.

One note about moving the logic to vanagon. This PR has logic that not all component branches follow the stable/master paradigm. For example, mco uses 2.8.x. Also PFTW contains essentially 2 components, whose branches are 2.1.x-x86 and 2.1.x-x64, and we'll want to check them both.

So I don't think we want vanagon to know that, but it will somehow need to pull this information from the puppet-agent project when deciding how to compare component tags against their respective branches.

PR #506, opened 2016-01-25
(RE-6286) Set -T instead of -t for remote_ssh_command

jcoconnor:
This fix applied to Vanagon is also required for the windows-build.rb script.
While the existing ssh-remote commands worked with the current live win-2012r2 vmpooler template, and updated ssh on the new template to be introduced (see [RE-6093](https://tickets.puppetlabs.com/browse/RE-6093)) causes a hang.

Changing the "ssh -tt" option to "ssh -T" resolves a problem with the remote-ssh commands which caused the build to hang.

All remote-ssh commands are set to echo out to the build log which also aids in debugging.

### DIFF
--- a/bin/build-windows.rb
+++ b/bin/build-windows.rb
@@ -72,7 +72,7 @@ def clone_and_rynsc_private_repo(fork, ref, hostname, ssh_key)
 end
 
 # Set up the environment so I don't keep crying
-ssh_command = "ssh #{ssh_key} -tt -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null Administrator@#{hostname}"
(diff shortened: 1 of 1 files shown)


rick:
I notice that `clone_and_rynsc_private_repo` doesn't/didn't use `ssh -tt`. Presuming that method is different in some fundamental way, but figured it was worth mentioning...

Iristyle:
Please retarget at stable, and I'll merge up after.  Thanks!

Iristyle:
Based on what we talked about in Vanagon Windows sync-up, please add a few more notes to commit msg:

* Difference between `-T` and `-tt`
* Why is this change necessary (i.e. upgrade from Vanagon X to Vanagon Y)

jcoconnor:
Superseded by https://github.com/puppetlabs/puppet-agent/pull/510

PR #508, opened 2016-01-25
(feature) Merge windows feature branch into master

melissa:
This is blocked on a vanagon release containing the nuget automation and will require a vanagon dependency bump/


configs/projects/puppet-agent.rb
```diff
@@ -47,9 +65,14 @@
...
       end
     end
+  elsif platform.is_windows?
+    # For windows, we need to ensure we are building for mingw not cygwin
+    platform_triple = platform.platform_triple
+    host = "--host #{platform_triple}"
   end
 
-  proj.setting(:gem_install, "#{proj.host_gem} install --no-rdoc --no-ri --local ")
+  proj.setting(:gem_install, "#{proj.host_gem} install --no-rdoc --no-ri --local --bindir #{proj.bindir} ")
```
melissa: Should I make this change only on windows?

configs/components/puppet.rb
```diff
@@ -96,9 +98,26 @@
     pkg.install_file("../osx_paths.txt", "/etc/paths.d/puppet-agent")
   end
 
+  if platform.is_windows?
+    pkg.environment "FACTERDIR" => settings[:prefix]
```
melissa: This is only needed until https://github.com/puppetlabs/facter/blob/8886a003ac6459e56d78cd3d2d98c6d1e8b1898b/lib/facter.rb.in#L32 is updated. Right now, it doesn't work properly.

cc/ @MikaelSmith

melissa:
https://github.com/puppetlabs/puppet-agent/pull/509 a few related/mainly cosmetic updates.

configs/components/puppet.rb
MikaelSmith: https://tickets.puppetlabs.com/browse/FACT-1294

configs/projects/puppet-agent.rb
melissa: https://github.com/puppetlabs/puppet-agent/pull/509 makes this change windows specific

stahnma:
Pending a vanagon release. (with the nuget feature support)

jcoconnor:
When are we planning the Vanagon release ?

melissa:
@jcoconnor we have a ticket in this sprint.

MikaelSmith:
Vanagon 0.5.1 went out with the latest Windows support.

MikaelSmith:
Something seems to misbehave with Solaris SPARC installing Puppet: http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/286/PLATFORM=solaris-10-sparc,label=vanagon/consoleText

configs/components/puppet.rb
```diff
@@ -96,9 +102,33 @@
...
+  else
+    vardir = File.join(settings[:prefix], 'cache')
+    configdir = settings[:puppet_configdir]
+  end
   pkg.install do
     [
-      "#{settings[:host_ruby]} install.rb --ruby=#{File.join(settings[:bindir], 'ruby')} --no-check-prereqs --bindir=#{settings[:bindir]} --configdir=#{settings[:puppet_configdir]} --sitelibdir=#{settings[:ruby_vendordir]} --codedir=#{settings[:puppet_codedir]} --configs --quick --man --mandir=#{settings[:mandir]}",
+      "#{settings[:host_ruby]} install.rb \
+        --ruby=#{File.join(settings[:bindir], 'ruby')} \
+        --check-prereqs \
```
MikaelSmith: Oh, this changed from no-check-prereqs to check-prereqs. That probably doesn't work for cross-compiles.

MikaelSmith:
Build of this branch passed at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/289/.

melissa:
I'd like to do at least another pass though this code to ensure nothing strange got on accidentally. Cc/ @shrug @underscorgan etc could you take a look through this?

configs/components/puppet.rb
```diff
@@ -59,6 +63,83 @@
     pkg.install_service "ext/solaris/smf/puppet.xml", "ext/solaris/smf/puppet", service_type: "network"
   when "aix"
     pkg.install_service "resources/aix/puppet.service", nil, "puppet"
+  when "windows"
+    description = "Periodically fetches and applies configurations from a Puppet master server."
+    service_hash = [
```
Iristyle: I will reiterate that I think using Ruby hashes to templatize Wix XML (in this case the `ServiceInstall` element) is a problematic idea:

* The abstraction is leaky - note that XMLisms have surfaced in the `:condition` as `CDATA`, and elsewhere as entity escapes like `&quot;`.  If the goal was to insulate from the Wix details, I don't think this approach succeeds.  Now consumers must be concerned with understanding the rules of this abstraction in addition to those of Wix / XML.  I would rather see these details remain in their native Wix XML files as its a clearer demarcation.
* The XML template built at  https://github.com/puppetlabs/vanagon/blob/d03a46ededb45d3df20566b35da2b18e149cb21b/resources/windows/wix/service.component.wxs.erb has a number of hardcoded elements / doesn't fully capture the flexibility of the underlying `ServiceInstall` type at http://wixtoolset.org/documentation/manual/v3/xsd/wix/serviceinstall.html.  Specifically, `Account`, `Password`, `Interactive`, `Type`, `ErrorControl` and `Vital` are all hard-coded.  `EraseDescription`, `LoadOrderGroup` are missing, as is the support for any of the child elements.  I'm concerned now that instead of having a solution that's flexible, there's a tight coupling to enable new functionality, blocked on the release of Vanagon gem updates.  That is a major concern IMHO.
* Our new abstraction doesn't come with the same documentation as the underlying abstraction - I'd rather be able to point directly at the Wix docs (which are quite good), rather than figuring out if an issue has been encountered in this higher level abstraction (such as properly specifying a value for interpolation, etc)
* I'm concerned that this makes manual testing more difficult - i.e. I'd rather be able to simply assemble a static set of Wix artifacts to manually test package changes than require bits to be codegen'd and glued together.
* We know we need to have Wix XML in this repo specific to the project, so there's no real savings here in terms of freeing ourselves from the Wix burden.  I think it ends up making things a little more confusing to have some bits specified like this, and other bits specified elsewhere, etc.  The assembly process is more difficult to cognitively understand

I do understand that we would like to make the pieces more modular / easier to compose - but can we perhaps do that by including / assembling plain old Wix snippets?  Or some other alternative solution?  Is my understanding correct that this implementation is partially driven by the need to leverage the Vanagon `install_service` API?  Are there other details that I'm missing?

/cc @puppetlabs/windows

jcoconnor: @Iristyle We did consider the alternative approach - See https://github.com/puppetlabs/puppet-agent/pull/514/files. It sounds like you think this might be safer.

melissa: @MikaelSmith has opinions on this too. He was suggesting we move this data into a seperate file that gets passed into the `install_service` method.

Iristyle: Yeah, I think a separate file is leaning in the right direction - then Vanagon can be blissfully unaware of these Wix implementation details.

melissa: I think also @ferventcoder had opinions on this?

melissa: @Iristyle if we move these hashed to a file, vanagon still has to be aware of wix and how to handle services... Are you saying that rather than have a template we should hard code the services into files? I think Michael's suggestion was to move these hashes into a file, and then read in the hash and pass it to the `install_service` command as it currently happening

MikaelSmith: My suggestions were about project organization. It sounds like there are functional implications to how we encode the data for WiX files, and I don't have much background on that.

joshcooper: I share @Iristyle concerns about providing an abstraction over WiX for a few reasons. My main concern is it will force us to implement everything in order to ship anything.

I'd be a big +1 on passing WiX files to vanagon, rather than vanagon knowing how to generate them (using ERB templates or otherwise). That way we can get MSIs building for projects we already have, validated in CI, and shipped. There will be some copy and paste for new projects, but it's fairly boiler plate code.

Then over time, continue working on splitting functionality into component repos, e.g. facter's WiX component should know to create `facts.d`, and refactoring common functionality where it makes sense.

I realize I'm late to the party and the ship may have already sailed, but I think it would be much less complicated to do this in stages as opposed to trying to create an abstraction over WiX, which is itself an abstraction over Windows Installer. Unfortunately both are difficult to work with, even for simple things.

melissa: totes not late to the party, that's why we have this PR to merge the feature branch into master! These concerns are great and I'm glad we're discussing them before we get the feature branch merged in.

melissa:
Re-run with http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/300/ to include the most recent changes. The only failures we're seeing are because we need to use 7zip to install nssm. That addition has yet to be released.

melissa:
Closed in favor of #553

PR #510, opened 2016-01-25
(RE-6286) Set -T instead of -t for remote_ssh cmd

jcoconnor:
This fix is a side-effect of the Windows/Vanagon build upgrade work,
where the new "win-2012r2-x86_64" pooler template caused the existing
pxp-agent/puppet-agent builds to hand in the pipeline (see RE-6093).
The hangs only occurred when running from Jenkins and were thought to be
related to the ssh/TERM handling as the actual hang occurred during a
remote-ssh command to install the cmake package on the target pooler
windows machine.

The updated pooler template had some additional cygwin packages added
which required a general cygwin package refresh. This caused the pooler
machine ssh version to update from Version "SSH-2.0-OpenSSH_6.5"  to
Version "SSH-2.0-OpenSSH_7.1"

While it is not clear why this should cause the hang, it is assumed that
the problem was related to Pseudo-tty handling issues when running under
Jenkins as a similar issue had been noted with the Cisco build under
Vanagon (RE-5314).  This particular fix changed the "ssh -t" option to
"ssh -T". The "-T" option disables pty allocation.
This (build-windows.rb) script actually used "ssh -tt" which forced pty
allocation even if ssh has no local tty.

Once applied, this change appears to resolve the issue and the script
works satisfactorily under both the existing live template
"win-2012r2-x86_64" and the new one "win-2012r2-x86_64.make"

### DIFF
--- a/bin/build-windows.rb
+++ b/bin/build-windows.rb
@@ -85,41 +85,41 @@ fail "Unable to connect to the host. Is is possible that you aren't on VPN or co
 
 Kernel.system("set -vx;#{ssh_command} 'source .bash_profile ; echo $PATH'")
 
-Kernel.system("#{scp_command} #{File.join(SCRIPT_ROOT, 'build-helpers.ps1')} Administrator@#{hostname}:/home/Administrator/")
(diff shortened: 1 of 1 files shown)


Iristyle:
Merged up to master in https://github.com/puppetlabs/puppet-agent/commit/7bbc2cde246acac6051c54c91bfcc061faea28d3

PR #511, opened 2016-01-26
(FACT-1321) Install windows batch scripts for facter

melissa:
When running facter on windows, we need a few batch scripts. This commit
adds them

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -197,6 +197,12 @@ component "facter" do |pkg, settings, platform|
 
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
 
+  if platform.is_windows?
+    pkg.install_file "ext/windows/facter.bat", "#{settings[:bindir]}/facter.bat"
+    pkg.install_file "ext/windows/facter_interactive.bat", "#{settings[:bindir]}/facter_interactive.bat"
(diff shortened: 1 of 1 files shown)


configs/components/facter.rb
```diff
@@ -197,6 +203,12 @@
 
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
 
+  if platform.is_windows?
+    pkg.install_file "../facter.bat", "#{settings[:bindir]}/facter.bat"
+    pkg.install_file "../facter_interactive.bat", "#{settings[:bindir]}/facter_interactive.bat"
+    pkg.install_file "../run_facter_interactive.bat", "#{settings[:bindir]}/run_facter_interactive.bat"
```
ferventcoder: this is an interesting path choice.

melissa: Why? That's where the resources were (previously). Because I added them as a sources (at the top of this file), they were a directory above the current working directory. I decided this wasn't how I wanted to deal with these files though.

melissa:
Depends on https://github.com/puppetlabs/facter/pull/1263

shrug:
:+1:

configs/components/facter.rb
ferventcoder: Just an observation. Would the working directory ever have an issue with this?

melissa:
Thank you for merging this, but it doesn't work yet since the PR this is dependent on hasn't been merged yet :(

PR #512, opened 2016-01-26
(PA-187) Install environment.bat on windows

melissa:
In puppet-agent on windows, we need to install environment.bat so our
batch files can have access to a cohesive group of environment
variables, etc. This commit adds that.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -1,6 +1,10 @@
 component "puppet" do |pkg, settings, platform|
   pkg.load_from_json("configs/components/puppet.json")
 
+  if platform.is_windows?
+    pkg.add_source("file://resources/files/windows/environment.bat", sum: "c698da37e559935e904f8690dd5c26fa")
+  end
+
   pkg.build_requires "ruby"
   pkg.build_requires "facter"
   pkg.build_requires "hiera"
@@ -142,6 +146,9 @@ component "puppet" do |pkg, settings, platform|
 
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
 
+  if platform.is_windows?
+    pkg.install_file "../environment.bat", "#{settings[:bindir]}/environment.bat"
+  end
 
   pkg.configfile File.join(configdir, 'puppet.conf')
   pkg.configfile File.join(configdir, 'auth.conf')
--- a/resources/files/windows/environment.bat
+++ b/resources/files/windows/environment.bat
@@ -1,0 +1,19 @@
+@ECHO OFF
+REM This is the parent directory of the directory containing this script.
+SET "PL_BASEDIR=%~dp0.."
+REM Avoid the nasty \..\ littering the paths.
+SET "PL_BASEDIR=%PL_BASEDIR:\bin\..=%"
+
+REM Set a fact so we can easily source the environment.bat file in the future.
+SET "FACTER_env_windows_installdir=%PL_BASEDIR%"
+SET "FACTERDIR=%PL_BASEDIR%"
+
+SET "PATH=%PL_BASEDIR%\bin;%PATH%"
+
+REM Enable rubygems support
+SET RUBYOPT=rubygems
+REM Now return to the caller.
+
+REM Set SSL variables to ensure trusted locations are used
+SET "SSL_CERT_FILE=%PL_BASEDIR%\ssl\cert.pem"
+SET "SSL_CERT_DIR=%PL_BASEDIR%\ssl\certs"


configs/components/puppet.rb
```diff
@@ -1,6 +1,10 @@
 component "puppet" do |pkg, settings, platform|
   pkg.load_from_json("configs/components/puppet.json")
 
+  if platform.is_windows?
+    pkg.add_source("file://resources/files/windows/environment.bat", sum: "140cc75b2dabfb015468ecdafc533090")
```
melissa: I would rather install this file as a part of the puppet-agent project file, but I wasn't sure if I could do that. This was the next best option that I could think of.

resources/files/windows/environment.bat
```diff
@@ -0,0 +1,35 @@
...
+SET RUBYOPT=rubygems
+REM Now return to the caller.
+
+REM Set SSL variables to ensure trusted locations are used
+SET SSL_CERT_FILE=%SystemRoot%\system32\ssl\cert.pem
+SET SSL_CERT_DIR=%SystemRoot%\system32\ssl\certs
+
+REM Set MCO specific variables
+SET SERVER_CONFIG=%PROGRAMDATA%\server.cfg
+SET CLIENT_CONFIG=%PROGRAMDATA%\client.cfg
```
melissa: These will have to be updated to where we put the mco configs, which will probably be back where they were before

ScottGarman:
:+1:

shrug:
:+1:

resources/files/windows/environment.bat
```diff
@@ -0,0 +1,35 @@
...
+REM Set SSL variables to ensure trusted locations are used
+SET SSL_CERT_FILE=%SystemRoot%\system32\ssl\cert.pem
+SET SSL_CERT_DIR=%SystemRoot%\system32\ssl\certs
+
+REM Set MCO specific variables
+SET SERVER_CONFIG=%PROGRAMDATA%\PuppetLabs\mcollective\etc\server.cfg
+SET CLIENT_CONFIG=%PROGRAMDATA%\PuppetLabs\mcollective\etc\client.cfg
+SET MCOLLECTIVED=%PL_BASEDIR%\bin\mcollectived
+SET MC_STARTTYPE=manual
+REM SET MC_STARTTYPE=auto
```
melissa: @richardc as per our conversation, these point to the original config locations

melissa: Hopefully these settings are correct...

melissa:
I think I need to quote all of the things here...

MikaelSmith:
Looks good to me, the only input I'd wait on is a thumbs-up from @richardc on the MCO settings.

resources/files/windows/environment.bat
```diff
@@ -0,0 +1,35 @@
...
+REM Get the file name we were originally called as.  e.g. puppet.bat or puppet
+REM or facter.bat or facter.  ~n means: will return the file name only of
+SET SCRIPT_TEMP=%~n1
+REM Strip off the extension of the script name.  We need to do this to know
+REM what to pass to ruby -S
+SET SCRIPT_NAME=%SCRIPT_TEMP:.bat=%
+REM Shift off the original command name we we were called
+SHIFT
+
+SET PATH=%PL_BASEDIR%\bin;%PATH%
```
MikaelSmith: Most of the PL_BASEDIR stuff seems unneeded (everything except `FACTER_env_windows_installdir`). I'm not sure about MCOLLECTIVED. Also, aren't SERVER_CONFIG and CLIENT_CONFIG the defaults for mco on Windows?

melissa:
@MikaelSmith updated!

resources/files/windows/environment.bat
```diff
@@ -0,0 +1,19 @@
...
+
+SET "PATH=%PL_BASEDIR%\bin;%PATH%"
+
+REM Enable rubygems support
+SET RUBYOPT=rubygems
+REM Now return to the caller.
+
+REM Set SSL variables to ensure trusted locations are used
+SET "SSL_CERT_FILE=%PL_BASEDIR%\ssl\cert.pem"
+SET "SSL_CERT_DIR=%PL_BASEDIR%\ssl\certs"
```
melissa: I'm not positive these are the final location of the certs, but it's where we're installing them currently in the openssl component. We may want to move them at a later date, in which case we would need to update this.

MikaelSmith: If we keep environment.bat, we'd probably want to make this file configured to the MSI install location. I'm more and more tempted to throw them all in `C:/Windows/` somewhere so we can have a single common location for all our applications to look.

MikaelSmith:
I'm fine merging this as-is and cleaning up configuration as we answer some of the outstanding questions.

melissa:
For posterity's sake, this is the environment.bat file installed by pftw
```
@ECHO OFF
REM This is the parent directory of the directory containing this script.
SET PL_BASEDIR=%~dp0..
REM Avoid the nasty \..\ littering the paths.
SET PL_BASEDIR=%PL_BASEDIR:\bin\..=%

REM Set a fact so we can easily source the environment.bat file in the future.
SET FACTER_env_windows_installdir=%PL_BASEDIR%

REM Get the file name we were originally called as.  e.g. puppet.bat or puppet
REM or facter.bat or facter.  ~n means: will return the file name only of
SET SCRIPT_TEMP=%~n1
REM Strip off the extension of the script name.  We need to do this to know
REM what to pass to ruby -S
SET SCRIPT_NAME=%SCRIPT_TEMP:.bat=%
REM Shift off the original command name we we were called
SHIFT

SET PUPPET_DIR=%PL_BASEDIR%\puppet
REM Facter will load FACTER_ env vars as facts, so don't use FACTER_DIR
SET FACTERDIR=%PL_BASEDIR%\facter
SET HIERA_DIR=%PL_BASEDIR%\hiera
SET MCOLLECTIVE_DIR=%PL_BASEDIR%\mcollective

SET PATH=%PUPPET_DIR%\bin;%FACTERDIR%\bin;%HIERA_DIR%\bin;%MCOLLECTIVE_DIR%\bin;%PL_BASEDIR%\bin;%PL_BASEDIR%\sys\ruby\bin;%PL_BASEDIR%\sys\tools\bin;%PATH%

REM Set the RUBY LOAD_PATH using the RUBYLIB environment variable
SET RUBYLIB=%PUPPET_DIR%\lib;%FACTERDIR%\lib;%HIERA_DIR%\lib;%MCOLLECTIVE_DIR%\lib;%RUBYLIB%;

REM Translate all slashes to / style to avoid issue #11930
SET RUBYLIB=%RUBYLIB:\=/%

REM Enable rubygems support
SET RUBYOPT=rubygems
REM Now return to the caller.

REM Set SSL variables to ensure trusted locations are used
SET SSL_CERT_FILE=%SystemRoot%\system32\ssl\cert.pem
SET SSL_CERT_DIR=%SystemRoot%\system32\ssl\certs
```

PR #513, opened 2016-01-26
remove references to aardwolf branch

pyther:
on Jan 22 2015, Melissa stated:
> The aardwolf branch is no longer is use. Releases in the current series are cut from the stable branch.

As someone that was trying to build puppet-agent, the references to the aardwolf branch mislead me. 

I figured this patch could start the discussion to updating the readme removing the references to aardwolf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,24 +83,16 @@ To get started creating nuget packages, see
 
 Branches in puppet-agent
 ---
-There are two types of branches in puppet-agent
-* release branches (e.g. aardwolf)
-* tracking branches (e.g. master and stable)
 
-The fundamental difference between the two is their relation with CI pipelines, especially wrt the component configurations:
-* on a release branch:
-  * all components reference component tags *never* SHAs
-* on a tracking branch:
+Tracking branch (master + stable):
   * some components may reference tags if they’re slow moving (ruby, openssl)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,14 +101,6 @@ Note that for all merges from master or stable, the merge should pick up:
 But never:
 * changes that bumped to a SHA inside config/components
 
-Here's a sample snippet used for a stable -> aardwolf merge:
```
stahnma: Should we keep this as a merge process from stable => master? @kylog ?

kylog: Yeah, good call. This could be used as the template for stable->master merges (replacing aardwolf with master throughout of course), since those merges will have the same exceptions for changes to various items in `configs/components`.

pyther: Should I attempt to update this language? I'm not entirely sure I'm qualified to do that.

kylog: Thanks for jumping back in @pyther; this kinda slipped off my radar. I think you could just retain this stanza, but s/aardwolf/master/g it. Alternately, leave it in and someone else can do that bit :)

pyther:
@kylog I pushed an update which I think does the trick.

shrug:
:+1:  lgtm

PR #515, opened 2016-01-26
(maint) Add elevate.exe to ruby on windows

melissa:
This is an executable compiled by Josh Cooper and used to run our
projects interactively

We need elevate.exe.config to allow elevate.exe to run against .NET 4.
See
puppetlabs/puppet-win32-ruby@70847bd
and PUP-1951 for more information.

elevate.exe is used to run our project with elevated permissions. This
doesn't work in a cygwin environment, but otherwise allows facter and
puppet to request UAC elevation and run in a command prompt window with
administrator privileges.

### DIFF
--- a/configs/components/ruby.rb
+++ b/configs/components/ruby.rb
@@ -3,6 +3,11 @@ component "ruby" do |pkg, settings, platform|
   pkg.md5sum "091b62f0a9796a3c55de2a228a0e6ef3"
   pkg.url "http://buildsources.delivery.puppetlabs.net/ruby-#{pkg.get_version}.tar.gz"
 
+  if platform.is_windows?
+    pkg.add_source "http://buildsources.delivery.puppetlabs.net/windows/elevate/elevate.exe", sum: "bd81807a5c13da32dd2a7157f66fa55d"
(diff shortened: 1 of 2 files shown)


stahnma:
Can we have a commit message that tells us why we want this (beyond that it's an executable compiled by Josh?)

melissa:
@stahnma good point. Does this look better?

configs/components/ruby.rb
```diff
@@ -3,6 +3,11 @@
   pkg.md5sum "091b62f0a9796a3c55de2a228a0e6ef3"
   pkg.url "http://buildsources.delivery.puppetlabs.net/ruby-#{pkg.get_version}.tar.gz"
 
+  if platform.is_windows?
+    pkg.add_source "http://buildsources.delivery.puppetlabs.net/windows/elevate/elevate.exe", sum: "bd81807a5c13da32dd2a7157f66fa55d"
```
shrug: Where does this live? Should it be it;s own component? or is that overkill?

shrug:
:+1:

configs/components/ruby.rb
melissa: I was thinking about it, but it's a part of ruby in `puppet-win32-ruby`, so I figured this would be easiest. We definitely can make it it's own component though.

PR #516, opened 2016-01-26
(QENG-3433) Add script to control promotions into pe.

waynr:


waynr:
This will be accompanied shortly by a PR to puppetlabs/ci-job-configs that makes use of this script.

stahnma:
Can we have commit message here that explains what is happening and why? Right now, I understand nothing from reading this.

waynr:
@stahnma I am closing this now, it was proposed as a way for developers to control promotion of puppet-agent artifacts into PE from the relative comfort of their git repo but @joshcooper said that he would prefer this behavior to remain within the domain of the job configuration repository.

PR #519, opened 2016-01-28
(RE-6236) Templatize wix service files

melissa:


melissa:
Dependent on https://github.com/puppetlabs/vanagon/pull/270

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,28 @@
   when "aix"
     pkg.install_service "resources/aix/mcollective.service", nil, "mcollective"
   when "windows"
-    puts "Service files not enabled on windows"
+    service_hash = [{
+      :directory_ref => "bindir",
```
melissa: This will be the bindir directory id that's used to define the bindir in the directory wix file

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,28 @@
...
+      :id => "MCOService",
+      :guid => "7601FCEA-90B3-CC69-6A69-4087FBC7292D",
+      :win64 => true,
+      :file => {
+        :id => "RubyWExe",
```
melissa: This will be the rubyw.exe id defined in wix

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,28 @@
...
+        :source => 'stagedir\\sys\\ruby\\bin\\rubyw.exe',
```
melissa: This also will be a variable, but I haven't figure out what that's going to look like yet

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,28 @@
...
+      },
+      :serviceinstall => {
+        :id => "MCOServiceInstaller",
+        :description => "Puppet Labs server orchestration framework",
+        :displayname => "Marionette Collective Server",
+        :name => "mcollective",
+        :start => "demand",
+        :arguments => '-I&quot;[INSTALLDIR]puppet\\lib;[INSTALLDIR]facter\\lib;[INSTALLDIR]hiera\\lib;[INSTALLDIR]mcollective\\lib&quot; -rubygems &quot;[INSTALLDIR]mcollective\\bin\\mcollectived&quot; --daemonize',
```
melissa: This also needs to be made into a variable, but I haven't decided what that's going to look like yet

configs/components/puppet.rb
```diff
@@ -60,7 +60,79 @@
...
+        :guid => "639ECD7F-6186-43D5-9E1A-FC0278DBEE15",
+        :win64 => true,
+        :conditions => ['<![CDATA[ (PUPPET_AGENT_STARTUP_MODE ~<> "manual") AND (PUPPET_AGENT_STARTUP_MODE ~<> "disabled") ]]>'],
+        :file => {
+          :id => "RubyExeAutomatic",
+          :source => 'stagedir\\sys\\ruby\\bin\\ruby.exe',
+        },
+        :serviceinstall => {
+          :id => "ServiceInstaller",
+          :description => "Periodically fetches and applies configurations from a Puppet master server.",
```
jcoconnor: A suggestion - since we are using the same content here (titles etc), we should really but them up in comment variables, e.g. service_description, display_name etc. etc.
That way, even though we have some of the horribleness of repetition, we can protect against mis-types in the service descriptions etc. etc.

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,33 @@
...
+      # it won't be correctly configured to start right away. The
+      # puppet run that configures mcollective will allow it to start
+      # and set it to automatic
+      :serviceinstall => {
+        :id => "MCOServiceInstaller",
+        :description => "Puppet Labs server orchestration framework",
+        :displayname => "Marionette Collective Server",
+        :name => "mcollective",
+        :start => "demand",
+        :arguments => '-I&quot;[INSTALLDIR]\\lib&quot; -rubygems &quot;[INSTALLDIR]\\bin\\mcollectived&quot; --daemonize',
```
melissa: I want this to be `settings[:libdir]`, but we'll have to figure out what `[INSTALLDIR]` should be, and then sub it out. Maybe default to `settings[:install_root]`, but that not might be sufficient. `settings[:prefix]` could also work.

melissa: As is stands, `settings[:install_root]` expands to `C:/Progra~1/Puppet~1` (Because spaces are difficult).

jcoconnor: [INSTALLDIR] is a Wix/MSI property - it gets set during the actual installation, so the [INSTALLDIR] property should be left as is.

melissa: I'm trying to figure out how much of the path to substitute out for `[INSTALLDIR]`. Since `[INSTALLDIR]` isn't in any of our paths to begin with, we have to manually add it. If we start with a path like `C:\Program Files\Puppet Labs\puppet\lib`, should we have this be `[INSTALLDIR]\puppet\lib` or `[INSTALLDIR]\lib`.

No one is suggesting we remove `[INSTALLDIR`.

jcoconnor: @melissa I checked internally on a recent MSI and ```INSTALLDIR``` is set to: ```ProgramFiles64Folder\Puppet Labs\Puppet```

I wonder would it make sense to walk through the contents of the key tables on the MSI using Instedit tomorrow to try and get a feel for how the directories are setup.

I actually find directory setup on Wix/MSI kind of tricky, so I go back and reference actual cases to see how its done.

John

jcoconnor: This is a the beginning of the actual directory hierarchy for the current puppet agent MSI:

```
directory                            Parent directory     Target Directory
--------------------------------------------------------------------------------------------------------------------
CommonAppDataFolder                  TARGETDIR             CommonAppDataFolder
ProgramFiles64Folder                 TARGETDIR             ProgramFiles64Folder
ProgramMenuFolder                    TARGETDIR             ProgramMenuFolder
TARGETDIR                            TARGETDIR
PuppetLabsAppData                    CommonAppDataFolder   CommonAppDataFolder\PuppetLabs
CodeDir                              PuppetLabsAppData     CommonAppDataFolder\PuppetLabs\code
FacterAppData                        PuppetLabsAppData     CommonAppDataFolder\PuppetLabs\facter
MCOAppData                           PuppetLabsAppData     CommonAppDataFolder\PuppetLabs\mcollective
PuppetAppData                        PuppetLabsAppData     CommonAppDataFolder\PuppetLabs\puppet
EnvironmentsDir                      CodeDir               CommonAppDataFolder\PuppetLabs\code\environments
HieraDataDir                         CodeDir               CommonAppDataFolder\PuppetLabs\code\hieradata
CodeModulesDir                       CodeDir               CommonAppDataFolder\PuppetLabs\code\modules
ProductionDir                        EnvironmentsDir       CommonAppDataFolder\PuppetLabs\code\environments\production
ManifestsDir                         ProductionDir         CommonAppDataFolder\PuppetLabs\code\environments\production\manifests
ModulesDir                           ProductionDir         CommonAppDataFolder\PuppetLabs\code\environments\production\modules
FactsDotD                            FacterAppData         CommonAppDataFolder\PuppetLabs\facter\facts.d
MCOConfDir                           MCOAppData            CommonAppDataFolder\PuppetLabs\mcollective\etc
MCOVarDir                            MCOAppData            CommonAppDataFolder\PuppetLabs\mcollective\var
MCOLogDir                            MCOVarDir             CommonAppDataFolder\PuppetLabs\mcollective\var\log
PuppetConfDir                        PuppetAppData         CommonAppDataFolder\PuppetLabs\puppet\etc
PuppetVarDir                         PuppetAppData         CommonAppDataFolder\PuppetLabs\puppet\var
PuppetLabs                           ProgramFiles64Folder  ProgramFiles64Folder\Puppet Labs
INSTALLDIR                           PuppetLabs            ProgramFiles64Folder\Puppet Labs\Puppet
bin                                  INSTALLDIR            ProgramFiles64Folder\Puppet Labs\Puppet\bin
dirC24A635189E52DF456E2FBC3F74A024A  INSTALLDIR            ProgramFiles64Folder\Puppet Labs\Puppet\bin
dir59A60B7031D17F6FAE34CBA46C04A779  INSTALLDIR            ProgramFiles64Folder\Puppet Labs\Puppet\facter

```

melissa: Is that how we want to continue in the future? Would it make more sense to set `[INSTALLDIR]` to `ProgramFiles64Folder\Puppet Labs`?

jcoconnor: Will try and dig out a decent example that explains how this works as it might help.

glennsarti: Also INSTALLDIR is a public MSI property and can be changed at install time.  There's no guarantee that INSTALLDIR is always "Prog. Files\Puppet Lans"

https://msdn.microsoft.com/en-us/library/windows/desktop/aa370912(v=vs.85).aspx

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,33 @@
...
+        :source => settings[:rubyw],
+      },
+      # This service is installed with start set to demand because
+      # it won't be correctly configured to start right away. The
+      # puppet run that configures mcollective will allow it to start
+      # and set it to automatic
+      :serviceinstall => {
+        :id => "MCOServiceInstaller",
+        :description => "Puppet Labs server orchestration framework",
+        :displayname => "Marionette Collective Server",
```
glennsarti: Shouldn't the display name be "Marionette Collective Service" or "Marionette Collective Agent"?.  The MCO Server runs elsewhere.

configs/components/marionette-collective.rb
melissa: Soo, isn't that the point of this question? What is the directory base we want to give users the ability to rip out and replace? `C:\Program Files\Puppet Labs` or `C:\Program Files\Puppet Labs\puppet`. We've traditionally had this be `C:\Program Files\Puppet Labs\Puppet`. Can I change the default on people? Will people have automation expecting to be able to set it to `/some/randome/place/bin/facter.exe` and then be surprised that it's actually `/some/randome/place/puppet/bin/facter.exe`? I know it's a required property, I know the user can change it, I know it gets set in the main wxs file. What's the path I should have after `[INSTALLDIR]`? `\\lib` or `\\puppet\\lib`?

configs/components/marionette-collective.rb
melissa: https://github.com/puppetlabs/puppet_for_the_win/blob/5e59d0e713f199253f86419ed17ca057fb9a7304/wix/puppet.wxs#L229 I don't think I have the authority to change the name

configs/components/marionette-collective.rb
melissa: Okay, let's pause this conversation, and decide the answer in https://github.com/puppetlabs/vanagon/pull/271/files#r51495646. We need that to work for more projects then just puppet-agent, and right now it's assuming a lot about the install location. These assumptions will not hold across other projects.

jcoconnor: @melissa 
This is the actual line in the original puppet.wxs file:
```Arguments="-I&quot;[INSTALLDIR]puppet\lib;[INSTALLDIR]facter\lib;[INSTALLDIR]hiera\lib;[INSTALLDIR]mcollective\lib&quot; -rubygems &quot;[INSTALLDIR]mcollective\bin\mcollectived&quot; --daemonize"```

whereas, the code above transforms to:
```Arguments="-I&quot;[INSTALLDIR]\lib&quot; -rubygems &quot;[INSTALLDIR]\bin\mcollectived&quot; --daemonize"```

So:
- There are some additional paths that need to be added (unless some of the other .bat work you are doing over-wides this).
- You need to use [INSTALLDIR]mcollective\bin to refer correctly to the mcollective binary directory.

File: configs/components/pxp-agent.rb [hunk 8]
```diff
@@ -93,7 +93,51 @@
...
+        :displayname => "Puppet PXP Agent",
+        :name => "pxp-agent",
+        :start => "demand",
+      },
+      # Various registry keys documented at https://nssm.cc/usage
+      :registries => {
+        :root => "HKLM",
+        :key => "SYSTEM\\CurrentControlSet\\Services\\pxp-agent\\Parameters",
+        :values => [
+          { :name => "AppDirectory", :value => "[INSTALLDIR]\\bin", :type => "expandable" },
```
jcoconnor: @melissa Need to change the :value definition to 
```:value => "[INSTALLDIR]pxp-agent\bin"

jcoconnor:
@melissa This looks almost ready for merge once the 3 comments above (1 for MCO and 2 for pxp-agent) are addressed.
:+1:

configs/components/marionette-collective.rb
melissa: @jcoconnor remember we're changing the pathing layout? This corresponds to the new layout. All the binaries are stored in the same directory `C:/Program Files/Puppet Labs/Puppet/bin`. We no longer have all the individual bindirs per project.

configs/components/pxp-agent.rb
melissa: @jcoconnor we no longer have individual binary directories for each project. All binaries live under `C:/Program Files/Puppet Labs/Puppet/bin`.

melissa:
@jcoconnor I addressed your two comments about the bindir location, but I don't see a second comment for pxp-agent

configs/components/pxp-agent.rb
jcoconnor: @melissa Ok - will cross check with the new layout, but if thats the case, you can ignore these comments.

jcoconnor:
@melissa Hmm - I thought I had put an additional comment for pxp-agent on line 117 similar to the one for line 116 - mustn't have hit the Comment button to actually submit it - apologies.

configs/components/pxp-agent.rb
melissa: To verify, all you have to do is unpack that tarball in the branch I gave you to build puppet-agent quickly. Or, you could just believe me that I know the layout of puppet-agent very very well by this point. Or check out https://github.com/puppetlabs/puppet-specifications/pull/63 which hasn't been merged, but does reflect the current reality of the windows feature branch.

jcoconnor: Hey - I know you know the puppet layout very well.
Did the check the specifications and you are correct about the revised layout.

jcoconnor:
:+1: LGTM

jcoconnor:
@melissa Piece of positive news - I have successfully compiled the generated ```service.puppet-agent.wxs``` file and it appears to compile without errors.

mckern:
I don't know how Windows services work at the same level as I do *nix services, but the Ruby in this doesn't look terribad. If this successfully builds (and @jcoconnor says it does), I'm :+1: for merging it and polishing as needed once it's in.

configs/components/pxp-agent.rb
melissa: Sorry for snapping. It's good to double check and make sure things are consistent.

PR #523, opened 2016-02-01
(maint) Add initial set of wix vars

melissa:
This commit adds in a base set of wix variables for processing templates

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -2,19 +2,46 @@ project "puppet-agent" do |proj|
   platform = proj.get_platform
 
   # Project level settings our components will care about
-  # Windows has its own separate layout
   if platform.is_windows?
-    # Our install root can't have spaces in it, so we take advantage of short cuts.
-    # However, in order to take advantage of these shortcuts, we need to explicitly
-    # create the directories before using any shortcuts
-    proj.directory "C:/Program Files/Puppet Labs"
-    proj.setting(:install_root, "C:/Progra~1/Puppet~1")
-    proj.setting(:sysconfdir, "C:/ProgramData/PuppetLabs")
+    proj.setting(:company_name, "Puppet Labs")
+    proj.setting(:company_id, "PuppetLabs")
+    proj.setting(:common_product_id, "PuppetInstaller")
+    proj.setting(:product_id, "Puppet")
+    proj.setting(:upgrade_code, "2AD3D11C-61B3-4710-B106-B4FDEC5FA358")
+    if platform.architecture == "x64"
+      proj.setting(:product_name, "Puppet Agent (64-bit)")
+      proj.setting(:win64, "yes")
+      proj.setting(:base_dir, "ProgramFiles64Folder")
+      proj.setting(:RememberedInstallDirRegKey, "RememberedInstallDir64")
+    else
+      proj.setting(:product_name, "Puppet Agent")
+      proj.setting(:win64, "no")
+      proj.setting(:base_dir, "ProgramFilesFolder")
+      proj.setting(:RememberedInstallDirRegKey, "RememberedInstallDir")
+    end
+    proj.setting(:links, {
+        :HelpLink => "http://links.puppetlabs.com/customer-support-foss",
+        :CommunityLink => "http://links.puppetlabs.com/windows-installer-starting-out",
+        :ForgeLink => "http://links.puppetlabs.com/forge-windows",
+        :NextStepLink => "http://links.puppetlabs.com/windows-installer-next-steps-foss",
+        :ManualLink => "http://links.puppetlabs.com/windows-manual-foss",
+      })
+    proj.setting(:UI_exitdialogtext, "Manage your first resources on this node, explore the Puppet community and get support using the shortcuts in the Documentation folder of your Start Menu.")
+
+    # Directory IDs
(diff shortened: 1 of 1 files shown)


configs/projects/puppet-agent.rb
```diff
@@ -98,6 +98,36 @@
     proj.identifier "com.puppetlabs"
   end
 
+  # WIX settings
+  if platform.is_windows?
+    proj.setting(:company_name, "Puppet Labs")
+    proj.setting(:company_word, "PuppetLabs")
+    proj.setting(:common_product_name, "PuppetInstaller")
+    proj.setting(:upgrade_code, "2AD3D11C-61B3-4710-B106-B4FDEC5FA358")
```
stahnma: Where does this come from?

melissa: https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/include/puppet.wxi#L35

configs/projects/puppet-agent.rb
```diff
@@ -2,19 +2,46 @@
...
+    proj.setting(:UI_exitdialogtext, "Manage your first resources on this node, explore the Puppet community and get support using the shortcuts in the Documentation folder of your Start Menu.")
+
+    # Directory IDs
+    proj.setting(:bindir_id, "bindir")
+
+    # We build for windows not in the final destination, but in the paths that correspond
+    # to the directory ids expected by WIX. This will allow for a portable installation (ideally).
+    proj.setting(:install_root, File.join("C:", proj.base_dir, proj.company_id))
+    proj.setting(:prefix, File.join(proj.install_root, proj.product_id))
+    proj.setting(:sysconfdir, File.join("C:", "CommonAppDataFolder", proj.company_id))
```
melissa: `<Directory Id="CommonAppDataFolder" Name="CommonAppData">`

@jpogran can this be like `Program Files` where we only specify the ID?

configs/projects/puppet-agent.rb
```diff
@@ -2,19 +2,46 @@
...
+        :ManualLink => "http://links.puppetlabs.com/windows-manual-foss",
+      })
+    proj.setting(:UI_exitdialogtext, "Manage your first resources on this node, explore the Puppet community and get support using the shortcuts in the Documentation folder of your Start Menu.")
+
+    # Directory IDs
+    proj.setting(:bindir_id, "bindir")
+
+    # We build for windows not in the final destination, but in the paths that correspond
+    # to the directory ids expected by WIX. This will allow for a portable installation (ideally).
+    proj.setting(:install_root, File.join("C:", proj.base_dir, proj.company_id))
```
melissa: This should end with `company_name`, but I can't get it to build with spaces still

configs/projects/puppet-agent.rb
```diff
@@ -180,11 +206,19 @@
     proj.component "ruby-selinux"
   end
 
-  proj.directory proj.install_root
-  proj.directory proj.prefix
-  proj.directory proj.sysconfdir
+
+  if platform.is_windows?
+    proj.directory proj.install_root, id: proj.company_id, name: proj.company_name
```
MikaelSmith: These look like new options for Vanagon. They'll just be ignored when calling directory?

configs/projects/puppet-agent.rb
MikaelSmith: So we need a function to generate the short version of the name, and to create the directory as before? Or going a different route for MSI staging?

configs/projects/puppet-agent.rb
melissa: I put them up as an example for @McdonaldSeanp to potential take inspiration from. I think it'll cause failures as is though

configs/projects/puppet-agent.rb
melissa: My solution to get the msi to create the directory with spaces was to pass in `:name` if the directory is supposed to be different than the directory we're using to build. I'm sure there are going to be a ton of consequences to this approach, but I don't know what else to do.

melissa: I don't think we can pass in the shortcuts to wix

configs/projects/puppet-agent.rb
```diff
@@ -180,11 +206,19 @@
     proj.component "ruby-selinux"
   end
 
-  proj.directory proj.install_root
-  proj.directory proj.prefix
-  proj.directory proj.sysconfdir
+
+  #if platform.is_windows?
+  #  proj.directory proj.install_root, id: proj.company_id, name: proj.company_name
```
melissa: This hasn't been implemented yet.... I should probably take it out...

shermdog:
:+1:

configs/projects/puppet-agent.rb
jcoconnor: I think this is ok, but would prefer @jpogran to verify

jcoconnor:
:+1: May need to check https://github.com/puppetlabs/puppet-agent/pull/523#discussion_r51808824 but we can fix this up when get light in and do all the verification.

configs/projects/puppet-agent.rb
melissa: This shouldn't block this PR getting merged. We'll set the ID when the directory work is ready. It's just something to start thinking about. I also realized this isn't even the right directory for this comment!

PR #524, opened 2016-02-02
(maint) Ensure windows packaging is consistent

melissa:
We need to keep windows packaging consistent. This includes a few places
where we need to special case where things are installed


ferventcoder:
LGTM on first review.

PR #525, opened 2016-02-02
(PA-162) Make config shell scripts executable

MikaelSmith:
The shell scripts are meant to be invoked to setup environments with the
puppet-agent bin paths, but were installed without execute permissions.
Add execute permissions to them, which Vanagon picks up as part of
packaging.


MikaelSmith:
Produced package seems to install with correct permissions:
```
root@t9sxuk0cwori0d8:~# ls -l /etc/profile.d/
total 8
-rwxr-xr-x 1 root root 159 Feb  2 18:33 puppet-agent.sh
-rw-r--r-- 1 root root 248 May  1  2015 vim.sh
```

stahnma:
These shouldn't need to be executable, as they are sourced and not executed.

MikaelSmith:
Then PA-162 should be "Won't Fix", and I suspect Beaker is doing something wrong /cc @kevpl @LuvCurves

stahnma:
Ah. This (IMHO) is a bug in Cisco's WRL. But since they're unlikely to change that, we can change it. :cry:

shermdog:
It does need to be fixed.  We would prefer to fix this in the vanagon component definition and contain it to only cisco-wrlinux-7 though.  New PR to follow.

MikaelSmith:
Thanks for clarification.

PR #528, opened 2016-02-05
(PA-183, RE-5735) PFTW logic migration to PA

melissa:


melissa:
Requires https://github.com/puppetlabs/vanagon/pull/278

configs/components/nssm.rb
```diff
@@ -0,0 +1,12 @@
+component "nssm" do |pkg, settings, platform|
+  pkg.load_from_json("configs/components/nssm.json")
+
+  # Because we're unpacking a zip archive, we need to set the path to the executable.
+  # We don't automatically have this set on windows, unfortunately. We need to set the
+  # path to have access to 7za.
+  pkg.environment "PATH" => "/cygdrive/c/ProgramData/chocolatey/tools:$$PATH" if platform.is_windows?
```
shermdog: Is this something we should be setting at the platform level?  Seems like we may need this in other components.

melissa: Yeah, probably. Do we have the ability to set it at the platform level?

shermdog: Doesn't look like we can :(  Obviously that negate my comment.  Let's discuss offline if we should add a ticket to vanagon for that in the future.

shermdog:
:+1:

configs/components/nssm.rb
MikaelSmith: You could have provisioning cat something in the ssh profile.

MikaelSmith:
This seems ok. Is Windows going to have an implementation of `install_service`?

melissa:
@MikaelSmith see https://github.com/puppetlabs/puppet-agent/pull/519 for `install_service`

melissa:
@MikaelSmith also https://github.com/puppetlabs/vanagon/blob/ab4841e87c0d58f4762208455f8f1047a8c45d7c/lib/vanagon/component/dsl.rb#L187-L188 for `install_service` on windows

configs/components/nssm.rb
melissa: I think there's a ticket somewhere for setting up a base environment... I can't find it at the moment, but this is definitely something that I'd like to see happen at some point

stahnma: The only we  to do this globally currently that I know of is have a file with the stuff you want that many/all of the components load in and eval. Not ideal. So, IOW, LGTM right now.

PR #529, opened 2016-02-05
(PA-185) Use publicly available URIs for git clone

stahnma:
Prior to this, some components were accessing github over SSH which is
something that most of the world can't do. This commit changes to using
the public URIs. Thus removing one of many issues when trying to build
puppet-agent outside of Puppet Labs.


stahnma:
@kylo if this gets merged, does it fix the issues, or is there a jenkins job doing promotion that will change back to private URIs?

stahnma:
Got clarification from @justinstoller. This will be fine.

McdonaldSeanp:
:+1:

PR #530, opened 2016-02-05
(maint) Only check reqs on windows

melissa:
In install.rb, the check-prereqs flag will check the system we are
running on and check to see if our requirements for install are
available. This requires that facter can run on the build system. This
is generally fine except when we are cross compiling. This flag also
enables some windows-specific functionality.

Because we need the extra functionality only on windows, we only need to
set this flag for windows. Everything else, including systems we cross
compile on, shouldn't need to check prereqs


MikaelSmith:
:+1:

PR #531, opened 2016-02-08
(maint) Install pid files and log files into ProgramData

melissa:
Since we need to maintain a consistent set of files in our MSI, we need
to set the pid and log dirs to be consistent locations to what has
already been established.


jcoconnor:
:+1:

melissa:
To update https://github.com/puppetlabs/puppet-agent/pull/531

jcoconnor:
Thats very self referential

PR #532, opened 2016-02-08
(PA-160) Add platform config for huaweios-6-ppc

ScottGarman:
The HuaweiOS environment is based on a containerized instance of
Debian 8, and runs on PPC hardware. We are using the x86_64 debian 8
pooler template to run cross-compiled builds from.

### DIFF
--- a/configs/platforms/huaweios-6-ppc.rb
+++ b/configs/platforms/huaweios-6-ppc.rb
@@ -1,0 +1,12 @@
+platform "huaweios-6-ppc" do |plat|
+  plat.servicedir "/lib/systemd/system"
+  plat.defaultdir "/etc/default"
+  plat.servicetype "systemd"
+  plat.codename "jessie"
+
+  plat.add_build_repository "http://pl-build-tools.delivery.puppetlabs.net/debian/pl-build-tools-release-#{plat.get_codename}.deb"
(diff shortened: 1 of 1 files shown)


shermdog:
I know @melissa really prefers we use the full path for commands such as `apt-get`

ScottGarman:
Well, I see no other examples of using the full path to apt commands in any of the other platform configs we have...

configs/platforms/huaweios-6-ppc.rb
```diff
@@ -0,0 +1,11 @@
+platform "huaweios-6-ppc" do |plat|
+  plat.servicedir "/lib/systemd/system"
+  plat.defaultdir "/etc/default"
+  plat.servicetype "systemd"
+  plat.codename "jessie"
+
+  plat.add_build_repository "http://pl-build-tools.delivery.puppetlabs.net/debian/pl-build-tools-release-#{plat.get_codename}.deb"
+  plat.provision_with "export DEBIAN_FRONTEND=noninteractive; apt-get update -qq; apt-get install -qy --no-install-recommends build-essential devscripts make quilt pkg-config debhelper rsync fakeroot"
+  plat.install_build_dependencies_with "DEBIAN_FRONTEND=noninteractive; apt-get install -qy --no-install-recommends "
+  plat.vmpooler_template "debian-8-x86_64"
```
stahnma: Can we have a comment somewhere in here that explains the intention to cross compile? Otherwise it looks weird to see ppc and then x86_64 in the definitions.

ScottGarman:
@stahnma comment added

melissa:
What ever is easier. I like it, but that's just my opinion. I think this generally looks good

PR #534, opened 2016-02-08
(PA-159) Ruby bits for huaweios-6-ppc

ScottGarman:
All the tasty morsels required to cross-compile ruby and associated gems.


configs/components/ruby.rb
```diff
@@ -143,6 +156,7 @@
     # This tells the ruby setup that it can use the default system gcc rather
     # than our own.
     target_dir = File.join(settings[:libdir], "ruby", "2.1.0", rbconfig_info[settings[:platform_triple]][:target_double])
+    sed = "sed"
     sed = "gsed" if platform.is_solaris?
     sed = "/opt/freeware/bin/sed" if platform.is_aix?
```
melissa: this makes me sad

configs/components/ruby-augeas.rb
```diff
@@ -24,6 +24,10 @@
       pkg.environment "RUBY" => settings[:host_ruby]
     end
     ruby = "#{settings[:host_ruby]} -r#{settings[:datadir]}/doc/rbconfig.rb"
+  elsif platform.is_huaweios?
+    pkg.environment "RUBY" => settings[:host_ruby]
+    ruby = "#{settings[:host_ruby]} -r#{settings[:datadir]}/doc/rbconfig.rb"
+    pkg.environment "LDFLAGS" => settings[:ldflags]
```
mckern: We don't set LDFLAGS like this by default?

configs/components/ruby.rb
```diff
@@ -143,6 +156,7 @@
     # This tells the ruby setup that it can use the default system gcc rather
     # than our own.
     target_dir = File.join(settings[:libdir], "ruby", "2.1.0", rbconfig_info[settings[:platform_triple]][:target_double])
+    sed = "sed"
     sed = "gsed" if platform.is_solaris?
```
mckern: This is sort of meh but I know you didn't make the mess. Can we clarify intent a little with something like:

```ruby
sed = platform.is_solaris? "gsed" : "sed"
```

configs/components/ruby-augeas.rb
shrug: We *should* but I'm not sure we do, I've sen this declared like this elsewhere too.

shrug: Poking through vanagon, we do not set a default for ldflags anywhere. feature idea!

configs/components/ruby.rb
mckern: Oh, I didn't even notice the third option. Good glob, what a mess.

mckern:
Some nitpicks, but no blockers. I wonder how long it'd take to abstract out cross compiling before the next platform.

configs/components/ruby.rb
shrug: case statement?

mckern: You'd think so, but try to construct it and see what happens.

shrug: Nope, I'm dumb, that wouldn't work.

shrug:
:+1:

PR #535, opened 2016-02-09
(PA-159) Facter and leatherman support for huaweios

ScottGarman:


stahnma:
:+1:

configs/components/facter.rb
```diff
@@ -109,10 +109,13 @@
       skip_blkid = 'OFF'
     end
   end
+  if platform.is_huaweios?
+    skip_blkid = 'ON'
```
melissa: Does this not default to on? Should we append `|| !platform.is_huaweios?` to line 100?

ScottGarman: @melissa  on line 100, the platform.is_deb? would be true for huaweios and that code branch would be activated without checking any of the later tests. So that wouldn't work.

melissa: awww :( never mind then

PR #537, opened 2016-02-10
(PA-159) libxml2 and libxslt support for huaweios

ScottGarman:


McdonaldSeanp:
:+1:

configs/components/libxslt.rb
```diff
@@ -8,6 +8,13 @@
   if platform.is_aix?
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-gcc-5.2.0-1.aix#{platform.os_version}.ppc.rpm"
     pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH"
+  elsif platform.is_huaweios?
+    pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH:#{settings[:bindir]}"
+    pkg.environment "CFLAGS" => settings[:cflags]
+    pkg.environment "LDFLAGS" => settings[:ldflags]
+
+    # libxslt is picky about manually specifying the build host
+    build = "--build x86_64-unknown-linux-gnu"
```
melissa: Do we need this for any other cross compiles that are happening?

ScottGarman: @melissa I don't know why the solaris sparc cross-build doesn't need this, but I'm not inclined to change it on other platforms for the time being.

melissa: totes fair. I was just wondering. If it works, then we're probably good

PR #538, opened 2016-02-11
(RE-6464) Add custom actions for the PA install

melissa:


melissa:
Requires https://github.com/puppetlabs/vanagon/pull/285

jpogran:
:+1: lgtm

jcoconnor:
@melissa We need to keep all files to be included as fully legal wix, not include files on an ERB.

melissa:
@jcoconnor these pieces are fully legal wix though. They may not be on their own, but they should never exist on their own.

melissa:
The output looks like https://gist.github.com/melissa/23c6ca022c3aa7375c64

jcoconnor:
Agree that they are fully legal wix parts, but they should be enclosed by <wix> tags and an xml header.

melissa:
@jcoconnor the gist I posted early is an excerpt showing just the places where the variable substitution was happening to highlight what that would look like. All of that is still being parsed as a part of the main wxs file, so it already has https://github.com/puppetlabs/vanagon/blob/master/resources/windows/wix/project.wxs.erb#L1-L2. Are you telling me to double next those tags?

jcoconnor:
@melissa I more concerned with the principle of reading in and combining files using the ERB.
I think we should stick to individual files.
I'm proposing that we take the Execute sequences out of product.wxs altogether and think we agreed this when we were discussing with with @ferventcoder and @Iristyle 

Lets defer discussion until I have an alternative PR posted.

mckern:
@jcoconnor @melissa Re: partials. It's a reasonable pattern. The use of partials (in ERB or other templating systems) generally allows changes to be made in isolation and keeps large templates confined to manageable sizes. What's your concern with the use of partials?

jcoconnor:
My preference would be to keep all Wix fragments as fully formed wix fragments.
Technically, there is nothing wrong with the ERB approach, but I'm reluctant at the moment to shoehorn too much cleverness into the build - in a similar way that we have decided to revert to a simpler service files structure.
/cc @ferventcoder @Iristyle @jpogran

melissa:
Closing this, as we've decided on a different approach

ferventcoder:
@jcoconnor I disagree with the approach to move to fully formed wix fragments and think you are likely to see pain in it. That is why I suggested this approach in the first place. There is nothing clever in it, IMHO it is simply the most flexible and appropriate option available given the want to split parts to a generalized template in Vanagon and other pieces kept in the puppet-agent (project) repo IMHO.

ferventcoder:
However if you have found a more flexible option that allows for fully formed WiX fragments, then that is a different story.

PR #542, opened 2016-02-16
(maint) Nest prefix under Puppet on windows

melissa:
In how things are installed on windows, we need to ensure all our
content is safely nested under the `installdir`. This generally consists
of a format like `Program Files\company\project`. So, to both conform
with this outline and to maintain parity with the existing puppet-agent
packages, this commit assumes the installdir will be
`ProgramFiles\PuppetLabs\Puppet`, which means our main bindir will be
`ProgramFiles\PuppetLabs\Puppet\bin` and the bindir will everything in
it will be `ProgramFiles\PuppetLabs\Puppet\puppet\bin`.


McdonaldSeanp:
:+1:

PR #544, opened 2016-02-17
(PA-159) Huaweios project config for puppet-agent

ScottGarman:


configs/projects/puppet-agent.rb
```diff
@@ -30,6 +30,18 @@
...
+  # HuaweiOS is a cross-compiled platform
+  if platform.is_huaweios?
+    platform_triple = "powerpc-unknown-linux-gnu"
+    host = "--host #{platform_triple}"
+
+    # Use a standalone ruby for cross-compilation
+    proj.setting(:host_ruby, "/opt/pl-build-tools/bin/ruby")
+    proj.setting(:host_gem, "/opt/pl-build-tools/bin/gem")
+
+    proj.noarch
```
melissa: Is it really noarch? Don't we want to confine architecture? https://github.com/puppetlabs/vanagon/blob/5236eefe97963f3cfc312bbef1e3e1b388d6ffc4/resources/deb/control.erb#L10

configs/projects/puppet-agent.rb
```diff
@@ -117,7 +129,7 @@
     proj.component "shellpath"
   end
 
-  if platform.is_solaris? || platform.name =~ /^el-4/ || platform.is_aix?
+  if platform.is_solaris? || platform.name =~ /huaweios|^el-4/ || platform.is_aix?
```
underscorgan: should this be `^huaweios`

configs/projects/puppet-agent.rb
ScottGarman: I'd like to add proper support for building PPC deb packages, but there is a lot of invasive work to do in vanagon for that. Also, setting noarch on the package avoids stripping the binaries, which is what we want for initial testing. So I'd say these are reasonable compromises for getting this into the repo in a timely manner.

ScottGarman:
Updated with a minor tweak suggested by @underscorgan

configs/projects/puppet-agent.rb
melissa: Fair enough. Is there a follow up ticket for that work?

ScottGarman: There is now! RE-6520.

melissa:
:+1:

PR #546, opened 2016-02-18
(MAIN) Point to production Win-2012r2 template

jcoconnor:
The cygwin updates for vanagon are now in production, so the VM Pooler
platform needs to be updated.


mckern:
Typo in commit message. Great job @melissa!

PR #547, opened 2016-02-19
(maint) Use sysv init for huaweios

ScottGarman:
HuaweiOS is based on Debian 8 but uses sysv init instead of systemd


stahnma:
:+1:

PR #548, opened 2016-02-24
(PA-159) Update huaweios build to use correct host triplet and arch

ScottGarman:
While investigating how Debian strips cross-compiled packages, I
realized we've been using the wrong host triplet for Debian powerpc.
The correct triplet should be powerpc-linux-gnu. This also changes
the arch from ppc to powerpc, another fix to follow Debian's naming
scheme.


mckern:
This appears to be internally sane and consistent. Do we need to make that change anywhere else?

ScottGarman:
Don't merge it just yet, I need to do a final shakedown test first.

mckern:
We have labels now!

underscorgan:
This looks reasonable

underscorgan:
:+1:

PR #550, opened 2016-02-25
(maint) Update conflicting components after merge-up

MikaelSmith:
Merging master to stable left two components - facter and pxp-agent -
unable to update independently. The old SHAs use a vendored Leatherman,
but a flaw in how they added Leatherman include directories means the
installed Leatherman causes a conflict. That would be fine if it were
just one (that one would pass on update), but since two components have
that issue the one not being updated will cause a build failure.

Update them concurrently to the latest SHA to bypass the issue.

### DIFF
--- a/configs/components/facter.json
+++ b/configs/components/facter.json
@@ -1,1 +1,1 @@
-{"url": "git://github.com/puppetlabs/facter.git", "ref": "8150368f38349f1dbf44e25a238c5e3d00dcda21"}
+{"url": "git://github.com/puppetlabs/facter.git", "ref": "7c86863931e83ce9215aa35ccca7780ba33029fc"}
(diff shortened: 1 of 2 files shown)


MikaelSmith:
ping @kylog

configs/components/facter.json
```diff
@@ -1 +1 @@
-{"url": "git://github.com/puppetlabs/facter.git", "ref": "8150368f38349f1dbf44e25a238c5e3d00dcda21"}
+{"url": "git://github.com/puppetlabs/facter.git", "ref": "7c86863931e83ce9215aa35ccca7780ba33029fc"}
```
mckern: I confess I am slightly out of the loop, but should we be using a SHA here or should we be using a tag or branch name?

kylog: This is an interim measure. Hop into Puppet Core for the deets.

PR #553, opened 2016-03-02
Windows Vanagon Feature Branch

Iristyle:
This is a cleaned up / rebased version of the Windows Vanagon Feature Branch ready for merge to the master branch of puppet-agent.

@puppetlabs/windows please review

### DIFF
--- a/configs/components/nssm.rb
+++ b/configs/components/nssm.rb
@@ -1,0 +1,12 @@
+component "nssm" do |pkg, settings, platform|
+  pkg.load_from_json("configs/components/nssm.json")
+
+  # Because we're unpacking a zip archive, we need to set the path to the executable.
+  # We don't automatically have this set on windows, unfortunately. We need to set the
+  # path to have access to 7za.
+  pkg.environment "PATH" => "/cygdrive/c/ProgramData/chocolatey/tools:$$PATH" if platform.is_windows?
+
+  win_arch = platform.architecture == "x64" ? "win64" : "win32"
+
+  pkg.install_file "#{win_arch}/nssm.exe", "#{settings[:bindir]}/nssm.exe"
+end
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -1,10 +1,14 @@
 component "cpp-pcp-client" do |pkg, settings, platform|
   pkg.load_from_json('configs/components/cpp-pcp-client.json')
+
   cmake = "/opt/pl-build-tools/bin/cmake"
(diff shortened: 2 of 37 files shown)


melissa:
Testing at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/307/

Iristyle:
Looks like execution failed due to `unzip` not being found - here are the results from `x64` and `x86`

http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/PLATFORM=windows-2012r2-x64,label=vanagon/307/consoleFull
http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/PLATFORM=windows-2012r2-x86,label=vanagon/307/consoleFull

configs/components/nssm.rb
```diff
@@ -0,0 +1,12 @@
+component "nssm" do |pkg, settings, platform|
+  pkg.load_from_json("configs/components/nssm.json")
+
+  # Because we're unpacking a zip archive, we need to set the path to the executable.
+  # We don't automatically have this set on windows, unfortunately. We need to set the
+  # path to have access to 7za.
```
Iristyle: Vanagon appears to be using `unzip` here instead of `7za`

melissa: That's expected. There's a fix that will be available in 0.5.2.

melissa:
There's a fix available for the above failures that will be in vanagon 0.5.2. Because this isn't causing any failures for other platforms, I vote that not block getting this merged

Iristyle:
@joshcooper / @kylog   - this PR is currently targeted at puppet-agent#master, and I wanted to let you know of the intent to merge this "real soon now".  

This should not impact the upcoming ships of 1.3.6 / 1.4.0 that we have planned.  CI jobs will still call the existing `build-windows.rb` on Windows, otherwise they will use Vanagon for other platforms as they always have.  Therefore the Windows specific Vanagon related code in this PR remains dormant until we switch it on in the relevant jobs.

When we *do* switch it on (which is still a ways out), there might be an increase in the time it takes to run through the packaging steps.  Just wanted to make sure that's properly coordinated relative to agent releases - I presume the code paths will be activated post-Burnside release.

Iristyle:
Ah thanks @melissa - that was the `0.5.2` change you mentioned earlier - got it.

kylog:
Thanks for the heads-up @Iristyle. Anything on master shouldn't interfere with 1.4.x / burnside, so nbd on that front. However, I would email a broad heads-up when you merge, and perhaps again when you activate, since a variety of teams are affected by puppet-agent builds.

Btw, wrt the potential future slowdown, heads-up to PAGI since they are actively hunting down perf issues. /cc @shrug @geoffnichols

PR #554, opened 2016-03-02
(maint) empty aardwolf branch to avoid confusion - no longer in use.

shrug:
Updated README to point at the stable branch


stahnma:
I think an empty branch is more confusing than removing the branch.

stahnma:
We've decided to delete the branch.

PR #514, opened 2016-01-26
(WIP) (RE-6640) Service Specific Files for Vanagon PA-Side

jcoconnor:
This is the initial set of service specific files to support the Puppet
Agent Vanagon Windows build.

WIP only - supporting modifications to files under vanagon/components coming next.

PR #557, opened 2016-03-03
(RE-6640) Service Specific Files for Vanagon

jcoconnor:
This is the initial set of service specific files to support the Puppet
Agent Vanagon Windows build.

This PR replaces https://github.com/puppetlabs/puppet-agent/pull/514/

### DIFF
--- a/configs/components/marionette-collective.rb
+++ b/configs/components/marionette-collective.rb
@@ -66,7 +66,9 @@ component "marionette-collective" do |pkg, settings, platform|
   when "aix"
     pkg.install_service "resources/aix/mcollective.service", nil, "mcollective"
   when "windows"
-    puts "Service files not enabled on windows"
+    # Note - this definition indicates that the file should be filtered out from the Wix
+    # harvest. A corresponding service definition file is also required in resources/windows/wix
+    pkg.install_service "SourceDir\\#{settings[:base_dir]}\\#{settings[:company_id]}\\#{settings[:product_id]}\\puppet\\bin\\rubyw.exe"
   else
     fail "need to know where to put service files"
   end
--- a/configs/components/pxp-agent.rb
+++ b/configs/components/pxp-agent.rb
@@ -103,7 +103,9 @@ component "pxp-agent" do |pkg, settings, platform|
   when "aix"
     pkg.install_service "resources/aix/pxp-agent.service", nil, "pxp-agent"
   when "windows"
-    puts "Service files not enabled on windows"
+    # Note - this definition indicates that the file should be filtered out from the Wix
+    # harvest. A corresponding service definition file is also required in resources/windows/wix
+    pkg.install_service "SourceDir\\#{settings[:base_dir]}\\#{settings[:company_id]}\\#{settings[:product_id]}\\puppet\\bin\\nssm.exe"
   else
     fail "need to know where to put #{pkg.get_name} service files"
   end
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -6,6 +6,7 @@ project "puppet-agent" do |proj|
     proj.setting(:company_name, "Puppet Labs")
     proj.setting(:company_id, "PuppetLabs")
     proj.setting(:common_product_id, "PuppetInstaller")
+    proj.setting(:puppet_service_name, "puppet")
     proj.setting(:product_id, "Puppet")
     proj.setting(:upgrade_code, "2AD3D11C-61B3-4710-B106-B4FDEC5FA358")
     if platform.architecture == "x64"
--- a/resources/windows/wix/service.mcollective.wxs.erb
+++ b/resources/windows/wix/service.mcollective.wxs.erb
@@ -1,0 +1,40 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+    <ComponentGroup Id="Service_MARIONETTECOLLECTIVE">
+      <ComponentRef Id="Service_MARIONETTECOLLECTIVE_svc" />
+    </ComponentGroup>
+
+    <DirectoryRef Id="MARIONETTECOLLECTIVEBINDIR">
+      <Component Id='Service_MARIONETTECOLLECTIVE_svc'
+                 Guid="7601FCEA-90B3-CC69-6A69-4087FBC7292D"
+                 Win64="<%= settings[:win64] %>">
+        <File Id="RubyWExe"
+              KeyPath="yes"
+              Source="SourceDir\<%= settings[:base_dir] %>\<%= settings[:company_id] %>\<%= settings[:product_id] %>\puppet\bin\rubyw.exe" />
+        <!-- This service is installed with start set to demand because
+        it won't be correctly configured to start right away. The
+        puppet run that configures mcollective will allow it to start
+        and set it to automatic -->
+        <ServiceInstall Id="MCOServiceInstaller"
+          Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+          Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+          Description="Puppet Labs server orchestration framework"
+          DisplayName="Marionette Collective Server"
+          Interactive="no"
+          Name="mcollective"
+          Start="demand"
+          Type="ownProcess"
+          ErrorControl="normal"
+          Vital="yes"
+          Arguments="-I&quot;[INSTALLDIR]puppet\lib&quot; -rubygems &quot;[INSTALLDIR]puppet\bin\mcollectived&quot; --daemonize"
+          />
+        <ServiceControl Id="MCOStartService"
+                        Stop="both"
+                        Remove="uninstall"
+                        Name="mcollective"
+                        Wait="yes" />
+      </Component>
+    </DirectoryRef>
+  </Fragment>
+</Wix>
--- a/resources/windows/wix/service.puppet.wxs.erb
+++ b/resources/windows/wix/service.puppet.wxs.erb
@@ -1,0 +1,102 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+    <ComponentGroup Id="Service_PUPPET">
+      <ComponentRef Id="PuppetServiceAutomatic" />
+      <ComponentRef Id="PuppetServiceManual" />
+      <ComponentRef Id="PuppetServiceDisabled" />
+    </ComponentGroup>
+
+    <DirectoryRef Id="PUPPETBINDIR">
+      <!--Seriously?... http://stackoverflow.com/questions/9419411/wix-setting-service-startup-type-using-a-property-property-not-recognized -->
+      <!-- Yep, Seriously - http://sourceforge.net/mailarchive/forum.php?thread_name=CANJN1a6gDE2tUUnP4SE05m8Ojh2q0Y8LrWOxbrF89YyJJCHn1A%40mail.gmail.com&forum_name=wix-users -->
+      <Component Id='PuppetServiceAutomatic'
+                 Guid="639ECD7F-6186-43D5-9E1A-FC0278DBEE15"
+                 Win64="<%= settings[:win64] %>">
+        <Condition>
+          <![CDATA[ (PUPPET_AGENT_STARTUP_MODE ~<> "manual") AND (PUPPET_AGENT_STARTUP_MODE ~<> "disabled") ]]>
+        </Condition>
+        <File Id="RubyExeAutomatic"
+              KeyPath="yes"
+              Source="SourceDir\<%= settings[:base_dir] %>\<%= settings[:company_id] %>\<%= settings[:product_id] %>\puppet\bin\ruby.exe" />
+
+        <ServiceInstall Id="ServiceInstaller"
+                        Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+                        Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                        Description="Periodically fetches and applies configurations from a Puppet master server."
+                        DisplayName="Puppet Agent"
+                        Interactive="no"
+                        Name="<%= settings[:puppet_service_name] %>"
+                        Start="auto"
+                        Type="ownProcess"
+                        ErrorControl="normal"
+                        Vital="yes"
+                        Arguments="-rubygems &quot;[INSTALLDIR]puppet\bin\daemon.rb&quot;" />
+        <ServiceControl Id="ServiceControlOptions"
+                        Start="install"
+                        Stop="both"
+                        Remove="uninstall"
+                        Name="<%= settings[:puppet_service_name] %>"
+                        Wait="yes" />
+      </Component>
+
+      <Component Id='PuppetServiceManual'
+                 Guid="752A5A25-9619-4EBA-AA8B-12D8C8688236"
+                 Win64="<%= settings[:win64] %>">
+        <Condition>
+          <![CDATA[PUPPET_AGENT_STARTUP_MODE ~= "manual"]]>
+        </Condition>
+        <File Id="RubyExeManual"
+              KeyPath="yes"
+              Source="SourceDir\<%= settings[:base_dir] %>\<%= settings[:company_id] %>\<%= settings[:product_id] %>\puppet\bin\ruby.exe" />
+
+        <ServiceInstall Id="ServiceInstallerManual"
+                        Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+                        Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                        Description="Periodically fetches and applies configurations from a Puppet master server."
+                        DisplayName="Puppet Agent"
+                        Interactive="no"
+                        Name="<%= settings[:puppet_service_name] %>"
+                        Start="demand"
+                        Type="ownProcess"
+                        ErrorControl="normal"
+                        Vital="yes"
+                        Arguments="-rubygems &quot;[INSTALLDIR]puppet\bin\daemon.rb&quot;" />
+        <ServiceControl Id="ServiceControlOptionsManual"
+                        Stop="both"
+                        Remove="uninstall"
+                        Name="<%= settings[:puppet_service_name] %>"
+                        Wait="yes" />
+      </Component>
+
+      <Component Id='PuppetServiceDisabled'
+                 Guid="4D3A8CAF-C675-46AC-B3AD-75F00581D0DB"
+                 Win64="<%= settings[:win64] %>">
+        <Condition>
+          <![CDATA[PUPPET_AGENT_STARTUP_MODE ~= "disabled"]]>
+        </Condition>
+        <File Id="RubyExeDisabled"
+              KeyPath="yes"
+              Source="SourceDir\<%= settings[:base_dir] %>\<%= settings[:company_id] %>\<%= settings[:product_id] %>\puppet\bin\ruby.exe" />
+
+        <ServiceInstall Id="ServiceInstallerDisabled"
+                        Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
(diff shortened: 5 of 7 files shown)


jcoconnor:
This is ready for merge (subject for review of course).
The service files compile, but as yet, the light.exe command doesn't produce and MSI due to some other missing parts.
Compile log: https://gist.github.com/jcoconnor/1810099576006228bd5b

I'm requesting that we merge these now to allow the Custom Action and UI fragment files to be added and to allow @McdonaldSeanp to proceed with a directory table to support these service files.

/cc @melissa @jpogran @Iristyle @ferventcoder

configs/components/pxp-agent.rb
```diff
@@ -103,7 +103,9 @@
   when "aix"
     pkg.install_service "resources/aix/pxp-agent.service", nil, "pxp-agent"
   when "windows"
-    puts "Service files not enabled on windows"
+    # Note - this definition indicates that the file should be filtered out from the Wix
+    # harvest. A corresponding service definition file is also required in resources/windows/wix
+    pkg.install_service "SourceDir\\ProgramFiles64Folder\\PuppetLabs\\Puppet\\bin\\nssm.exe"
```
melissa: should we make this a variable so people don't have to type it twice?

resources/windows/wix/service.mcollective.wxs.erb
```diff
@@ -0,0 +1,28 @@
...
+          Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+          Description="Puppet Labs server orchestration framework"
+          DisplayName="Marionette Collective Server"
+          Interactive="no"
+          Name="mcollective"
+          Start="demand"
+          Type="ownProcess"
+          ErrorControl="normal"
+          Vital="yes"
+          Arguments="-I&quot;[INSTALLDIR]puppet\lib;[INSTALLDIR]facter\lib;[INSTALLDIR]hiera\lib;[INSTALLDIR]mcollective\lib&quot; -rubygems &quot;[INSTALLDIR]mcollective\bin\mcollectived&quot; --daemonize"
```
melissa: This arguments are no longer correct with the new libdir layout

resources/windows/wix/service.mcollective.wxs.erb
```diff
@@ -0,0 +1,28 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+    <DirectoryRef Id="PUPPETBINDIR">
```
melissa: Did you get this working? Where are we defining this directory?

resources/windows/wix/service.pxp-agent.wxs.erb
```diff
@@ -0,0 +1,45 @@
...
+  <!--
+  This is how to include wxi files.  We expect this file to define variables
+  like the version numbers and the upgrade UUID.
+  -->
+
+  <Fragment>
+    <DirectoryRef Id="PUPPETBINDIR">
+      <Component Id="service_nssm" Guid="52B1CD57-95A2-4CA4-AB8E-9DDD6DE8FC66">
+        <CreateFolder />
+        <File Id="NSSM" KeyPath="yes" Source="SourceDir\ProgramFiles64Folder\PuppetLabs\Puppet\bin\nssm.exe" />
```
melissa: This should be `SourceDir\ProgramFiles64Folder\PuppetLabs\Puppet\puppet\bin\nssm.exe` as per https://github.com/puppetlabs/puppet-specifications/pull/63

resources/windows/wix/service.pxp-agent.wxs.erb
```diff
@@ -0,0 +1,45 @@
...
+          Interactive="no"
+          Name="pxp-agent"
+          Start="demand"
+          Type="ownProcess"
+          ErrorControl="normal"
+          Vital="yes" />
+        <!-- Various registry keys documented at https://nssm.cc/usage -->
+        <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\pxp-agent\Parameters">
+          <RegistryValue Name="AppDirectory" Value="[INSTALLDIR]bin" Type="expandable" />
+          <RegistryValue Name="Application" Value="[INSTALLDIR]bin\pxp-agent.exe" Type="expandable" />
```
melissa: `Value="[INSTALLDIR]\puppet\bin\pxp-agent.exe"`

melissa:
@jcoconnor a lot of these paths are wrong

configs/components/pxp-agent.rb
McdonaldSeanp: @melissa are you talking about the directory structure above the .exe s? The precedent set by the other platforms is to leave the entire path in. unless we wanted to change all I would say no variable

resources/windows/wix/service.mcollective.wxs.erb
McdonaldSeanp: We do not have this working, However at this point I think it will be better to merge this PR with this in to facilitate other work. Eventually this will either be searched for or defined in the directorylist.wxs file I am working on. EIther way this will stay in. What john was proposing (an idea that I like) is that for now we can merge wix files that can pass the candle step. Then when we have moved past getting light to execute we can make sure all directoryrefs work.

configs/components/pxp-agent.rb
melissa: We don't have any variables that help with the other paths, since they're getting pulled from the local repos. Since we're installing `nssm.exe`, we know where it's going to go already (https://github.com/puppetlabs/puppet-agent/blob/master/configs%2Fcomponents%2Fnssm.rb#L11). In this case, we're installing nssm to bindir. Bindir currently evaluates to `C:/ProgramFiles64Folder/PuppetLabs/Puppet/puppet/bin`, so with a little path munging, if we ever change the location of nssm.exe, then we'd only have to change it in one place rather than three.

melissa: but the path munging required might make that not worth it. I dunno, I'll leave it up to you

jcoconnor:
@melissa Wasn't sure about the paths - esp the nssm ones. Why are we putting it down a level (puppet) for nssm.exe but not ruby.exe?

configs/components/pxp-agent.rb
jcoconnor: Initially I thought no to using variables, but since the service files are .erb files, no reason why we can't use a variable here. Will try it out tammara

melissa:
@jcoconnor ruby is on the same path as nssm, and is wrong in this PR

jcoconnor:
@melissa and @McdonaldSeanp 
1. I have gone thru the service files and *.rb scripts and corrected the pathnames.
2. I have also updated the test build with the latest puppet-agent....tar.gz which has a number of pathname fixes - this now appears to match the [file specification PR](https://github.com/puppetlabs/puppet-specifications/pull/63/files)

Can you check the arguments for the mco and puppet-agent service files - I think I have specified the -rubygems correctly, but would appreciate if you could verify.

resources/windows/wix/service.pxp-agent.wxs.erb
jcoconnor: Corrected - see note below

resources/windows/wix/service.pxp-agent.wxs.erb
jcoconnor: Corrected - it should be ```Value="[INSTALLDIR]puppet\bin\pxp-agent.exe"```

resources/windows/wix/service.mcollective.wxs.erb
jcoconnor: Agreed - Think they are correct now - please verify

configs/components/pxp-agent.rb
jcoconnor: I have decided not to use a variable for the moment - mainly due to the \\ and \ quoting rules.

configs/components/marionette-collective.rb
```diff
@@ -66,7 +66,9 @@
   when "aix"
     pkg.install_service "resources/aix/mcollective.service", nil, "mcollective"
   when "windows"
-    puts "Service files not enabled on windows"
+    # Note - this definition indicates that the file should be filtered out from the Wix
+    # harvest. A corresponding service definition file is also required in resources/windows/wix
+    pkg.install_service "SourceDir\\ProgramFiles64Folder\\PuppetLabs\\Puppet\\puppet\\bin\\rubyw.exe"
```
melissa: It's not always going to be `ProgramFiles64Folder`. https://github.com/puppetlabs/puppet-agent/blob/master/configs%2Fprojects%2Fpuppet-agent.rb#L19

jcoconnor: So can I just put in ```#{settings[:base_dir]}``` instead?

McdonaldSeanp: Yeah, I would do that

jcoconnor: Thats done - Have also rebased the changes to master again.

configs/projects/puppet-agent.rb
```diff
@@ -6,6 +6,7 @@
     proj.setting(:company_name, "Puppet Labs")
     proj.setting(:company_id, "PuppetLabs")
     proj.setting(:common_product_id, "PuppetInstaller")
+    proj.setting(:PuppetServiceName, "puppet")
```
melissa: Rather than using camel case here, I'd prefer we keep it in the ruby convention with underscores. Not a blocker, just a personal preference. If that opinion isn't shared, I'm not going to fight for it.

resources/windows/wix/service.mcollective.wxs.erb
```diff
@@ -0,0 +1,28 @@
...
+      <Component Id='MCOService' Guid="7601FCEA-90B3-CC69-6A69-4087FBC7292D" Win64="<%= settings[:win64] %>">
+        <File Id="RubyWExe" KeyPath="yes" Source="SourceDir\ProgramFiles64Folder\PuppetLabs\Puppet\puppet\bin\rubyw.exe" />
```
melissa: `ProgramFiles64Folder` needs to change here too

configs/projects/puppet-agent.rb
McdonaldSeanp: agreed, underscores

jcoconnor: You mean like ```puppet_service_name``` ?
Thats fine - I need to fix the program files folder anyway.

jcoconnor: Done

resources/windows/wix/service.mcollective.wxs.erb
jcoconnor: Done

resources/windows/wix/service.puppet.wxs.erb
```diff
@@ -0,0 +1,73 @@
...
+        </Condition>
+        <File Id="RubyExeDisabled" KeyPath="yes" Source="SourceDir\<%= settings[:base_dir] %>\PuppetLabs\Puppet\puppet\bin\ruby.exe" />
+
+        <ServiceInstall Id="ServiceInstallerDisabled"
+          Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+          Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+          Description="Periodically fetches and applies configurations from a Puppet master server."
+          DisplayName="Puppet Agent"
+          Interactive="no"
+          Name="<%= settings[:PuppetServiceName] %>"
```
melissa: puppet_service_name

resources/windows/wix/service.pxp-agent.wxs.erb
```diff
@@ -0,0 +1,45 @@
...
+  <!--
+  This is how to include wxi files.  We expect this file to define variables
+  like the version numbers and the upgrade UUID.
+  -->
+
+  <Fragment>
+    <DirectoryRef Id="PUPPETBINDIR">
+      <Component Id="service_nssm" Guid="52B1CD57-95A2-4CA4-AB8E-9DDD6DE8FC66">
+        <CreateFolder />
+        <File Id="NSSM" KeyPath="yes" Source="SourceDir\<%= settings[:base_dir] %>\PuppetLabs\Puppet\puppet\bin\nssm.exe" />
```
melissa: `PuppetLabs` and `Puppet` have corresponding settings, and could be parameterized

resources/windows/wix/service.puppet.wxs.erb
jcoconnor: Yeah - just spotted that in test :(  Fixed and pushed up.

resources/windows/wix/service.pxp-agent.wxs.erb
jcoconnor: Are they already used to define directories?

melissa: https://github.com/puppetlabs/puppet-agent/blob/master/configs%2Fprojects%2Fpuppet-agent.rb#L6-L9

jcoconnor: OK. Will sort Mon.

jcoconnor: Done

jcoconnor:
@jpogran Can you please review this today ?
Also - /CC @Iristyle @ferventcoder @melissa

jcoconnor:
I have just pushed some changes to this PR in connection with Vanagon PR: https://github.com/puppetlabs/vanagon/pull/308

resources/windows/wix/service.mcollective.wxs.erb
```diff
@@ -0,0 +1,40 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+    <ComponentGroup Id="Service_MARIONETTECOLLECTIVE">
```
melissa: Does this correspond to `:id => "#{service_name.gsub(/([^A-Za-z0-9])/, '').upcase}"`? Where is `:id` used?

jcoconnor: @melissa Yes - thats where :id is used.
I was going to change :id to Service_(gsubbed stuff), but after looking at Sean's code, he can also make use of the :id if we leave it without a Service_ prefix.

melissa: If that's the case, can we make this a variable? Especially since `id` has some syntax requirements

jcoconnor: OK - thats another MAINT Fix that I need to push up on the Vanagon Side - will be having BJ with @McdonaldSeanp shortly so will discuss with him as it impacts his mods too.

melissa: Okay, I'm seeing some discrepancies. In https://github.com/puppetlabs/vanagon/commit/77664321791beeb01d29ebdf3cbb941a78b669e0, we set `:id => "#{service_name.gsub(/([^A-Za-z0-9])/, '').upcase}"` but then when we reference that in the wix files, we prepend `Service_` to it. Which is it? `MARIONETTECOLLECTIVE` or `Service_MARIONETTECOLLECTIVE`? Are we ever going to need `service.id` to be `MARIONETTECOLLECTIVE` instead of `Service_MARIONETTECOLLECTIVE`?

McdonaldSeanp: So here are the things at play here: 

* I need to get :id as it is now (without the prepended service_) for the directorylist work, so when we are creating the id field in the component dsl, it should not have Service_

* The point of prepending Service_ serves as a measure of documentation. Specifically for me the prepended service makes it clear that this component element contains the service WiX for MCO and not any of the contents of MCO itself. So in this way I like that the component element has Id="Service_MARIONETTECOLLECTIVE" and not just Id="MARIONETTECOLLECTIVE"

jcoconnor:
For reference - this is now building against Sean's latest DirList changes in Vanagon: https://gist.github.com/jcoconnor/0b15bfaf6b4fd4144a37

glennsarti:
:+1: from me.  Merge it.

resources/windows/wix/service.mcollective.wxs.erb
```diff
@@ -0,0 +1,40 @@
...
+      <ComponentRef Id="MCOService" />
+    </ComponentGroup>
+
+    <DirectoryRef Id="PUPPETBINDIR">
```
melissa: Should this be `PUPPETBINDIR` or `MARIONETTECOLLECTIVEBINDIR`?

resources/windows/wix/service.pxp-agent.wxs.erb
```diff
@@ -0,0 +1,61 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+
+  <Fragment>
+    <ComponentGroup Id="Service_PXPAGENT">
+      <ComponentRef Id="Service_NSSM" />
```
melissa: Can we have this be consistent? I know it's isolated to the service itself, but we currently have three different ways we're defining this. `<ComponentRef Id="MCOService" />`, <ComponentRef Id="PuppetServiceAutomatic" /> (and friends), and `<ComponentRef Id="Service_NSSM" />`.

resources/windows/wix/service.mcollective.wxs.erb
McdonaldSeanp: Also We can't really make this a variable, since when vanagon would get to this erb it would not have context as to which service it was editing this file for, so unless we did something like take a hard line on the filename and say the filename needs to be the name of the service there is no way for us to know which service to put here.

resources/windows/wix/service.mcollective.wxs.erb
McdonaldSeanp: It should be MARIONETTECOLLECTIVEBINDIR

resources/windows/wix/service.pxp-agent.wxs.erb
melissa: Or, wait, do these need to be consistent from pftw? If that's the case, then ignore that last comment.

resources/windows/wix/service.pxp-agent.wxs.erb
```diff
@@ -0,0 +1,61 @@
...
+    </ComponentGroup>
+
+    <DirectoryRef Id="PUPPETBINDIR">
```
melissa: Should this be `PXPAGENTBINDIR`?

resources/windows/wix/service.pxp-agent.wxs.erb
jcoconnor: They don't have to be consistent with pftw - I think I was just using the original ID's.
Will modify it slightly.
You are correct about the BINDIRs - will fix these now too.

jcoconnor: Have just pushed up an additional commit for these.
Will squash the commits once I have tested a local build.

resources/windows/wix/service.mcollective.wxs.erb
```diff
@@ -0,0 +1,40 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <Fragment>
+    <ComponentGroup Id="Service_MARIONETTECOLLECTIVE">
+      <ComponentRef Id="Service_MARIONETTECOLLECTIVE_svc" />
+    </ComponentGroup>
+
+    <DirectoryRef Id="MARIONETTECOLLECTIVEBINDIR">
+      <Component Id='Service_MARIONETTECOLLECTIVE_svc'
```
melissa: I mean, I kind of preferred the `MCOService` syntax, since that means we don't have to spell out MARIONETTECOLLECTIVE everytime in capital letters.... also, if you do want to stick with this syntax, can we please please please please use the mco service ID that gets set already?

jcoconnor: I guess one way we could do this simply (lets MAINT it though) is to add a serviceref or shortref key in the ```install_service``` method, where we would use this instead of the full ```MARIONCOLLECTIVE``` name. (just thinking on the hoof). If this is not provided, then it would default to the full length name.

McdonaldSeanp: @melissa That's fine, I can open up a maint PR that overrides the service name to MCO. Which will make all these become MCO instead of MARIONETTECOLLECTIVE. And @jcoconnor overriding the name in install service is already provided. If you pass in install service like so: `install_service #{path_to_bin} nil #{name}` it will actually already do the name override for us =) no need to add anything to install_service. refer to: https://github.com/McdonaldSeanp/vanagon/blob/feature/RE-6455/dirlist/spec/lib/vanagon/platform/windows_spec.rb#L186L197

PR #561, opened 2016-03-06
(PA-242) Disable boost::locale on cisco wrlinux 

branan:
wrlinux is an embedded environment that does not have full locale support. We shouldn't be attempting to do i18n there.

This requires Leatherman 0.4.2, which we plan to tag Monday. The only change in 0.4.2 will be the necessary support for disabling using locales on Linux variants that don't properly support it.

### DIFF
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -28,6 +28,10 @@ component "cpp-pcp-client" do |pkg, settings, platform|
     pkg.build_requires "pl-gcc"
     pkg.build_requires "pl-cmake"
     pkg.build_requires "pl-boost"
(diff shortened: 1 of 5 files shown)


branan:
Updated to include the needed leatherman bump

branan:
https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/ is a run of this against a couple standard linux platforms to hopefully show no regressions with pulling in 0.4.2 before we get this in

MikaelSmith:
This should be ready to go.

stahnma:
:+1:

PR #562, opened 2016-03-07
(PA-235) Fix rubygem cross-compile issues for huaweios

ScottGarman:
It turns out there is only so much that can be done with ruby gems for
a cross-compiled platform. The nokogiri gem, due to its native extensions,
had to be archived from a native environment installation.

The newest version of nokogiri requires mini_portile2 instead of mini_portile, so I have removed the old component config.


configs/components/rubygem-nokogiri.rb
```diff
@@ -1,16 +1,31 @@
 component "rubygem-nokogiri" do |pkg, settings, platform|
-  pkg.version "1.6.6.2"
-  pkg.md5sum "fc9f91534bf93d57b84f625b55732a7c"
-  pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-#{pkg.get_version}.gem"
+  pkg.version "1.6.7.2"
+  if platform.is_huaweios?
+    pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-precompiled-huaweios-#{pkg.get_version}.tar.gz"
```
stahnma: Can we have comments about where/how this tarball was built and/or came from? Like when we get a new nokogiri, what do we do?

configs/components/rubygem-net-netconf.rb
```diff
@@ -1,17 +1,30 @@
 component "rubygem-net-netconf" do |pkg, settings, platform|
   pkg.version "0.4.3"
-  pkg.md5sum "fa173b0965766a427d8692f6b31c85a4"
-  pkg.url "http://buildsources.delivery.puppetlabs.net/net-netconf-#{pkg.get_version}.gem"
+  if platform.is_huaweios?
+    pkg.url "http://buildsources.delivery.puppetlabs.net/net-netconf-#{pkg.get_version}.tar.gz"
```
stahnma: Where do we get this?

configs/components/rubygem-nokogiri.rb
ScottGarman: How about if I include a README file in the tarball describing the build process? Or do you think it should be commented in the component config here?

configs/components/rubygem-net-netconf.rb
ScottGarman: This is the same situation as nokogiri - it's an archive from a native install. I had to do this because if you try to install this gem on an x86 server, the install process notes that no x86 extensions exist for nokogiri, and it blows up trying to build nokogiri again.

configs/components/rubygem-nokogiri.rb
mckern: Ideally, both.

configs/components/rubygem-net-netconf.rb
mckern: There are a few gems that let you basically compile native binaries into a gem, so maybe we should look at moving to that method (in the future).

mckern:
thumbsup

But we should absolutely ticket refactoring this pattern for the future.

ScottGarman:
I've now added documentation on how the nokogiri and net-netconf gem tarballs were generated. 

I also talked with Haus over lunch about a way to possibly force-install the net-netconf gem without having to to use this tarball method, but I'd like my current approach to be merged first as we are very close to the code freeze date for 1.4.0.

stahnma:
Wow, this is gross. Do we have a path to a cleaner way to do this in the near term?

mckern:
@stahnma We think so. And yes, it is. You should feel repulsed by this because @ScottGarman and I certainly do.

configs/components/rubygem-net-netconf.rb
```diff
@@ -1,17 +1,35 @@
...
+    pkg.md5sum "fa173b0965766a427d8692f6b31c85a4"
+  end
 
   pkg.build_requires "ruby"
   pkg.build_requires "rubygem-nokogiri"
   pkg.build_requires "rubygem-net-scp"
 
-  # Because we are cross-compiling on ppc, we can't use the rubygems we just built.
-  # Instead we use the host gem installation and override GEM_HOME. Yay?
-  pkg.environment "GEM_HOME" => settings[:gem_home]
```
melissa: do we no longer need `GEM_HOME`?

mckern: We might? Maybe? Ruby. Who knows?

ScottGarman: That was a valid catch, but I have a suggestion from Huas on a change I can make to the net-netconf installation process that will make that code relevant again. As soon as I get my facter PR merged, I will return to this issue and hope to eliminate the need for this custom net-netconf tarball.

PR #564, opened 2016-03-07
(RE-6149) Build with Wix310

melissa:


jcoconnor:
:+1:

jcoconnor:
@melissa Have you successfully run a pipeline build on this - if so, please reference so it can be closed.
Change is straightforward enough (assuming no typos in 3.5/3.10 change which a build will expose).

melissa:
I'm still testing. I'll ping people when testing is complete.

melissa:
Built both architectures with the `build-windows.rb` script.

I also added installing wix during the vanagon build so we know we have the right wix version when we switch packaging formats. This change will add a little extra time to the build, but it shouldn't be a huge amount.

PR #568, opened 2016-03-10
(WIP) (RE-6849) Add UI/Localisation files

jcoconnor:
Adding the UI Localisation files as these are connected with the Custom
Action file sequences and are best added here for reference

Note the LICENSE file in the sample is for reference only - a subsequent ticket will ensure that the up to date license file is included.

See Vanagon PR [#314](https://github.com/puppetlabs/vanagon/pull/314)


jpogran:
Just file adds here, no functional changes so :+1: . Maybe a note in the commit that these dialog sets are being pulled in because of a UIRef (and a note how the UIRef is used when that is decided)

McdonaldSeanp:
Yeah, I have run this code about 50+ times, definitely :+1:

jcoconnor:
Hold off merging until I rework the UI inclusion bit.

jcoconnor:
@McdonaldSeanp @jpogran @melissa Have updated this and drawn in [RE-6757](https://tickets.puppetlabs.com/browse/RE-6757) so that the generic/vanagon solution uses the minimal UI, whereas the Puppet-Agent uses the full PA UI.

PR #569, opened 2016-03-10
Merge stable to master

MikaelSmith:


MikaelSmith:
@underscorgan

configs/components/puppet-ca-bundle.json
```diff
@@ -1,4 +1,4 @@
 {
-  "url": "git://github.com/puppetlabs/puppet-ca-bundle.git",
+  "url": "git@github.com:puppetlabs/puppet-ca-bundle.git",
```
MikaelSmith: It is public. I'm not sure about ssh keys. The URLs are inconsistent; currently pxp-agent and cpp-pcp-client are the only others using `git@`.

joshcooper: Seems like we should fix those URLs in stable (in a separate PR) and merge that up to master, so we don't get different behavior across branches. /cc @melissa @underscorgan 

:+1: to merge this PR as-is

MikaelSmith:
This would be really nice to merge this, but it's unrelated to any releases.

melissa:
I believe this might need to be updated, as the security stuff was merged into stable yesterday. If that's not the case, and this already includes the cve changes, then I'm :+1:

MikaelSmith:
Ok, I'll remerge.

MikaelSmith:
Updated.

melissa:
:+1:  I'm fine with this getting merged. We should standardize the git urls, but that's for a later PR

configs/components/puppet-ca-bundle.json
melissa: https://github.com/puppetlabs/puppet-agent/commit/e0b4ba1b9f1db5c4caa52503d82d22a2b95d0a23 this was an intentional change made in puppet-agent#master. We should try to keep it around...

MikaelSmith: Should I cherry-pick that up to stable?

melissa: I don't know what the best answer is... I guess cherry-picking them to stable would be the cleanest thing to do?

MikaelSmith: I'll just avoid merging for now.

bin/build-windows.rb
```diff
@@ -119,14 +124,14 @@ def clone_and_rynsc_private_repo(fork, ref, hostname, ssh_key)
 puts "Build-Windows.rb... building cpp-pcp-client"
 Kernel.system("set -vx;#{scp_command} #{File.join(SCRIPT_ROOT, 'build-cpp-pcp-client.ps1')} Administrator@#{hostname}:/home/Administrator/")
 fail "Copying build-cpp-pcp-client.ps1 to #{hostname} failed" unless $?.success?
-clone_and_rynsc_private_repo(CPPPCPCLIENT['url'], CPPPCPCLIENT['ref'], hostname, ssh_key)
+clone_and_rynsc_private_repo(CPPPCPCLIENT['url'], CPPPCPCLIENT['ref'], hostname, ssh_key, 'cpp-pcp-client')
```
MikaelSmith: Do we want to preserve this?

MikaelSmith:
Dropped the `git@` change.

bin/build-windows.rb
melissa: Yes, this was a change that was made in the cve branch. It ultimately should be in both master and stable

melissa:
Okie dokie, I think we're good to go on merging this!

configs/components/puppet-ca-bundle.json
MikaelSmith: Cherry-pick was messy, PR at https://github.com/puppetlabs/puppet-agent/pull/574.

joshcooper:
:+1:

PR #571, opened 2016-03-11
(PA-181) Explicitly specify -march for OS X builds

mckern:
This should explcitly target `core2` as the chipset to optimize for
at compilation time. That will hopefully help move along the
"illegal instruction" problem we've been seeing.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -68,6 +68,7 @@ component 'augeas' do |pkg, settings, platform|
     end
   elsif platform.is_osx?
     pkg.environment "PATH" => "$$PATH:/usr/local/bin"
+    pkg.environment "CFLAGS" => settings[:cflags]
   end
 
   pkg.configure do
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -23,6 +23,7 @@ component "cpp-pcp-client" do |pkg, settings, platform|
     platform_flags = '-DCMAKE_SHARED_LINKER_FLAGS="-Wl,-bbigtoc"'
   elsif platform.is_osx?
     cmake = "/usr/local/bin/cmake"
+    platform_flags = "-DCMAKE_CXX_FLAGS='#{settings[:cflags]}'"
     toolchain = ""
   elsif platform.is_huaweios?
     cmake = "/opt/pl-build-tools/bin/cmake"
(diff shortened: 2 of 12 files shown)


mckern:
@stahnma We now have one confirmed report for "works on my machine," but I'd like a few more before we produce a signed build we let customers test.

geoffnichols:
@mckern If you point me to a build, I can run upstairs and test it.

stahnma:
@geoffnichols http://hdrive.websages.com/stahnma/puppet-agent-1.3.5.414.g2d2f0b1-1.osx10.10.dmg

groob:
@stahnma I've tested the above installer on a several 2010 iMacs which were previously affected by the issue. Looks good here. 
Will there be an updated version of the OS X 10.11 build once the PR is merged?

stahnma:
@groob awesome!  We need to do a bit more testing, but then we'll get this on the tree. After that, the builds should hit nightlies.puppetlabs.com (though they might miss 1.4.0, as that's being cut this week).

MikaelSmith:
Any reason we're limiting it to the Facter component?

mckern:
@MikaelSmith initially confining the change to one place specifically so that we have an easy canary for A/B testing viability. I was wondering about whether or not to make it a default for the entire platform though.

MikaelSmith:
I would expect potential issues for all compiled applications; I'm not sure what's unique about Facter here.

mckern:
@MikaelSmith CMake is what's unique. Almost everything else uses autotools.

MikaelSmith:
Ah yes; then all the CMake-based projects should use it.

arubdesu:
We've tested the above posted build on older hardware against our 2015.3 master and see basic functionality.

stahnma:
@mckern what about pxp/pcp stuff? Shouldn't that have the same cmake options?

mckern:
I'm going to apply these flags across the board for OS X builds and get an agent with that built.

shrug:
:+1:  pending a test build

mckern:
I have produced a working build from the revised `CFLAGS` (as defined at the project level, not the component level). These should be :ok_hand: to review and merge.

stahnma:
@mckern have a build available?

stahnma:
http://hdrive.websages.com/stahnma/puppet-agent-1.3.6.523.g85807c8-1.osx10.10.dmg

groob:
Installed this new build on a few macs and checked facter, ruby and the puppet agent. Everything works as expected. If you'd like to test anything specific, I'd be happy to confirm. We have a few hundred affected machines here.

shrug:
Still :+1:

PR #577, opened 2016-03-16
(maint) fix cross-compiling of virt-what

ScottGarman:
virt-what needs the --host option to specify the platform for cross
compiling, so previously this never actually worked. This commit
adds the proper --host flag and sets PATH/CFLAGS/LDFLAGS for huaweios.

I've tested this with the huaweios agent build and it now builds a binary for powerpc. I did *not* touch the el-7 powerpc64 configuration as I don't even believe it includes the necessary cross-toolchain so I can't fully test that. It looks like that platform was added by a community member.


shrug:
:+1:

PR #578, opened 2016-03-17
(RE-6520) Build the huaweios package as powerpc instead of noarch

ScottGarman:
The addition of binutils-powerpc-linux-gnu is necessary for the debian
packaging to be able to find the target arch strip command. I cannot
force debuilder to pass along a modified PATH environment to dh_strip,
so the distro version of binutils is required for this instead of our
pl-binutils.

This PR depends on [PR-322](https://github.com/puppetlabs/vanagon/pull/322) getting merged into vanagon first.


shrug:
:+1:

PR #580, opened 2016-03-17
(RE-6735) Eliminate the need for a custom net-netconf installation process

ScottGarman:
By using the --force flag, we can return to the normal process of
installing the net-netconf gem. However, in order to do this, we
need to ensure that the nokogiri gem is the *last* gem that gets
installed, since the native extesion build process is so broken
when cross-compiling.


configs/projects/puppet-agent.rb
```diff
@@ -211,8 +211,8 @@
   if platform.is_huaweios?
     proj.component "rubygem-net-scp"
     proj.component "rubygem-mini_portile2"
-    proj.component "rubygem-nokogiri"
     proj.component "rubygem-net-netconf"
+    proj.component "rubygem-nokogiri"
```
mckern: I'm not sure the load order actually matters here. I believe Vanagon will actually build a graph and map out dependencies that way, so this may be superfluous.

mckern:
I'm generally :+1: on this. This looks like a solution that's only a little bit clever and entirely livable. Great job.

configs/projects/puppet-agent.rb
ScottGarman: In some experimentation I did, the order the components are declared did matter. I'm pretty sure of this, but not 100%.

mckern: I'll default to you on this.

PR #581, opened 2016-03-18
(PA-181) Simplify/correct CFLAGS and LDFLAGS logic

mckern:
The cascading if/else for setting CFLAGS could lead to more than one situation
no LDFLAGS are set or the standard LDFLAGS are duplicated in each conditional.
We can improve this by setting broad, generally correct defaults and then only
adding to them or changing them as needed. This also eases the pressures that
likely led to a cascading if/else statement.


shrug:
yuuup :+1:

mckern:
Travis looks like it's stuck trying to do a dumb. I'll close this and reopen it to kick Travis.

PR #582, opened 2016-03-18
(RE-6507) Add USER_ACCOUNT Wix definitions

jcoconnor:
Normally the puppet services are run under LocalSystem, but the User
settings allow it to be overridden so that for example a Domain account
can be configured to run the services

### DIFF
--- a/resources/windows/wix/componentgroup.wxs.erb
+++ b/resources/windows/wix/componentgroup.wxs.erb
@@ -16,6 +16,8 @@
       <ComponentGroupRef Id="FragmentProperties" />
       <ComponentGroupRef Id="FragmentSequences" />
       <ComponentGroupRef Id="FragmentCustomActions" />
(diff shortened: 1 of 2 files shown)


jcoconnor:
This user code compiles and is added to the installer without errors.
I have performed the following installation tests:
1. Install with test user "foobar" - user does not exist on platform - fails to install as user does not exist.
2. Install with test user "foobar" - User exists on platform, but is not in Administrator group - fails to install.
3. Install with test user "foobar" - User exists in platform in Administrator group - install succeeds - service starts under "./foobar user".
4. Install with no user parameters - Install succeeds, services run under "Local System"

So this appears to behave as expected and is ready to be merged.

I checked with @rob - we don't have an Active Directory configuration available we can't test it using proper domain accounts - action for the future - liaise with IT to setup AD machine for testing.

McdonaldSeanp:
:+1:

PR #586, opened 2016-03-21
(PA-181) Simplify/correct CFLAGS and LDFLAGS logic again

mckern:
The previous version of this change had incorrect LDFLAGS defined for OS X
and not all components safely strip bad values. This has been corected, and
a comment about how LDFLAGS should be set for OS X has been added.

The cascading if/else for setting CFLAGS could lead to more than one situation
no LDFLAGS are set or the standard LDFLAGS are duplicated in each conditional.
We can improve this by setting broad, generally correct defaults and then only
adding to them or changing them as needed. This also eases the pressures that
likely led to a cascading if/else statement.

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -138,7 +138,12 @@ project "puppet-agent" do |proj|
     proj.bill_of_materials File.join(proj.datadir, "doc")
   end
 
-  # Platform specific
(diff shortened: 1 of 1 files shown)


mckern:
This built correctly. Have at it.

ScottGarman:
:+1:

PR #595, opened 2016-03-22
Re 6470/enable special case of config files in the app data folder

McdonaldSeanp:
This is direct cut over of parts of https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs as specific static files pertaining to the special case for config files in CommonAppDataFolder

### DIFF
--- a/resources/windows/wix/appdatafiles.wxs
+++ b/resources/windows/wix/appdatafiles.wxs
@@ -1,0 +1,247 @@
+<?xml version='1.0' encoding='windows-1252'?>
+<Wix xmlns='http://schemas.microsoft.com/wix/2006/wi' xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
+  <!--
+    Puppet Agent Specific Application data
+    These directories and files need to be special cased
+    in order to manage the application data for puppet
+    agent. The specific special case is that all files
+    and folders here need to be persistent across multiple
+    installs and uninstalls. We can do so by specifying
+    Permanent="yes" on each file/folder component.
+  -->
+  <Fragment>
+    <DirectoryRef Id="TARGETDIR">
+      <Directory Id="CommonAppDataFolder" Name="CommonAppData">
+        <Directory Id="APPDATADIR" Name="PuppetLabs">
+          <Directory Id="pxpAgentDataFolder" Name="pxp-agent">
+            <Directory Id="pxpAgentDataFolderVar" Name="var">
+              <Directory Id="pxpAgentDataFolderLog" Name="log" />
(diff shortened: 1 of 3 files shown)


jcoconnor:
@McdonaldSeanp This looks sound but I need to build it and test it tomorrow to confirm.

resources/windows/wix/appdatafiles.wxs
```diff
@@ -0,0 +1,160 @@
...
+          <Directory Id="EnvironmentsDir" Name="environments">
+            <Directory Id="ProductionDir" Name="production">
+              <Directory Id="ManifestsDir" Name="manifests"/>
+              <Directory Id="ModulesDir" Name="modules"/>
+            </Directory>
+          </Directory>
+        </Directory>
+      </Directory>
+    </DirectoryRef>
+    <ComponentGroup Id="AppDataComponentGroup" >
```
jcoconnor: Can you indent all the component lines as well - really anything with 3 or more items should be indented for clarity.

resources/windows/wix/appdatafiles.wxs
```diff
@@ -0,0 +1,160 @@
...
+            </Directory>
+            <Directory Id="pxpAgentDataFolderEtc" Name="etc">
+              <Directory Id="pxpAgentDataFolderModules" Name="modules"/>
+            </Directory>
+          </Directory>
+          <Directory Id="PuppetAppData" Name="puppet">
+            <Directory Id="PuppetConfDir" Name="etc"/>
+            <Directory Id="PuppetVarDir" Name="var"/>
+          </Directory>
+        </Directory>
```
jcoconnor: @McdonaldSeanp This line belongs just before line 45.
If you check the install you will see that  ```code```, ```facter``` and ```mcollective``` are being created under ```ProgramData``` and not ```PuppetLabs```

jcoconnor:
@McdonaldSeanp Have tested this and it appears to install ok, with the one exception that 3 of the directories are placed under ProgramData instead of PuppetLabs - have marked the line where I think you need to move a ```</Directory>``` statement.

The ProgramData directories and configuration files are left in place after the uninstall which is expected.

Fixing the Component indentation would also make this a lot more readable especially under githubs insanely narrow review windows.

jcoconnor:
Sean - it would be good to add some more documentation at the top of the file as to why these files are being given the special treatment.

jcoconnor:
Have just tested in build and install - the ProgramData issue has been fixed.

:+1:

ferventcoder:
LGTM :+1:

PR #596, opened 2016-03-23
(maint) update customactions to erb

McdonaldSeanp:
customactions now has some .erb in it
after a recent update from some of the
pftw work. this updates the file to .erb


McdonaldSeanp:
Merged on the trivial rule, this will allow builds of windows to finish again

PR #597, opened 2016-03-23
(maint) update the windows plat definitions to msi

McdonaldSeanp:
The MSI packaging process is in a state where it's possible
to build an MSI straight out of the box with no changes to
vanagon or any configs. We now need MSI to be the default
type of windows build

### DIFF
--- a/configs/platforms/windows-2012r2-x64.rb
+++ b/configs/platforms/windows-2012r2-x64.rb
@@ -21,5 +21,5 @@ platform "windows-2012r2-x64" do |plat|
 
   plat.platform_triple "x86_64-unknown-mingw32"
 
-  plat.package_type "nuget"
+  plat.package_type "msi"
 end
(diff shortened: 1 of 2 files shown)


McdonaldSeanp:
This has been tested and the MSI is good for x64

ferventcoder:
@Iristyle @puppetlabs/windows note this will adversely affect your https://github.com/puppetlabs/ci-job-configs/pull/1119

configs/platforms/windows-2012r2-x64.rb
```diff
@@ -21,5 +21,5 @@
 
   plat.platform_triple "x86_64-unknown-mingw32"
 
-  plat.package_type "nuget"
+  plat.package_type "msi"
```
melissa: Also, it should just default to building msis, so you should be able to just remove this line. It's not a bad thing to keep in in though, for extra clarity

McdonaldSeanp: I would say lets leave it in until we are farther in to the process, for now I think it's good to document we are using very new functionality.

jcoconnor:
:+1: 
Makes a lot of sense to do this now.

Iristyle:
:+1: - once this goes in, I will update my PR

PR #605, opened 2016-03-28
All the HuaweiOS goodies from master

ScottGarman:


configs/components/ruby-augeas.rb
```diff
@@ -46,6 +46,6 @@
 
   pkg.install do
     "chown root:root #{settings[:ruby_vendordir]}/augeas.rb"
-  end if platform.is_solaris?
+  end if platform.is_solaris? || platform.is_huaweios?
```
mckern: These are already in `master`, right?

ScottGarman: Yep, everything was cherry-picked from master. We can refactor some stuff after the fact.

mckern: We'll fix these things in `master` then and worry about it later on.

mckern:
:+1: this seems sane. I expect this would be targeted at P-A 1.4.2?

ScottGarman:
Yes, the point of this is to ensure everything is in stable for 1.4.2.

PR #609, opened 2016-03-30
(RE-6882) hard-pin vanagon to 0.5.7

ScottGarman:
This is a temporary change since the rubygems.org index cache
is still reporting that vanagon 0.5.8 is available, but now
the gem itself can no longer be downloaded as it was yanked.


shrug:
:+1:

PR #610, opened 2016-03-30
(RE-6882) hard-pin vanagon to 0.5.6

ScottGarman:
The stable branch should not follow new vanagon releases unless
explicitly updated in the Gemfile.


shrug:
:+1:

PR #613, opened 2016-03-31
(RE-6852) Add Image Resource Variable definitions

jcoconnor:
These definitions point to the Image resouces used by the installer UI.
At the moment, the main image is set to the FOSS version of the BMP


jcoconnor:
This was built and tested on Windows-2012.
The UI is now displaying the Puppet logos instead of the default Wix logos.

PR #616, opened 2016-04-01
(RE-6863) Remove pxp-agent service custom actions

jcoconnor:
These Custom actions leaked into to the Vanagon build and should be
removed. They were included for PA-65, but this change was subsequently
removed.


McdonaldSeanp:
:+1:

PR #617, opened 2016-04-03
(RE-6970) Use huaweios as the codename for the huaweios agent

ScottGarman:
Previously we were building it as 'jessie', which means the package
will get shipped to the Debian Jessie repo. Instead we want to ensure
the package is shipped to a separate apt repo for huaweios.


underscorgan:
:+1:

PR #618, opened 2016-04-04
(RE-6892) Move the .bat files location

McdonaldSeanp:
Before this we were planning on keeping the .bat files
that actually execute puppet in windows inside Puppet\puppet\bin
and creating symlinks in Puppet/bin to the .bat files. However
we have since decided that the symlinks are not the way to proceed
since they require custom actions for both install and uninstall
and the act of symlinking is moving away from being windows native.

Thus we elected to move the .bat files entirely to the Puppet\bin
directory and change the contents to point to the correct paths.

This update does the following:

* Adds each .bat file we were using in Puppet-Agent to the actual
Puppet-agent vanagon project. this means those files are now under
the resources/files/windows and we will no longer use those generated
by any of the install.rb scripts.

    * AS A SPECIAL NOTE: this PR itself does not explicitly
    deprecate those .bat files created by the install.rb
    scripts. that work (to remove those files from Puppet/puppet/bin)
    will come as a different PR.

* Updates the .bat files to point to the correct location of all the
executables in Puppet/puppet/bin

### DIFF
--- a/configs/components/hiera.rb
+++ b/configs/components/hiera.rb
@@ -10,11 +10,22 @@ component "hiera" do |pkg, settings, platform|
   pkg.replaces 'pe-hiera'
 
   pkg.install do
-    "#{settings[:host_ruby]} install.rb --ruby=#{File.join(settings[:bindir], 'ruby')} --bindir=#{settings[:bindir]} --configdir=#{settings[:puppet_codedir]} --sitelibdir=#{settings[:ruby_vendordir]} --configs --quick --man --mandir=#{settings[:mandir]}"
+    ["#{settings[:host_ruby]} install.rb \
+    --ruby=#{File.join(settings[:bindir], 'ruby')} \
+    --bindir=#{settings[:bindir]} \
+    --configdir=#{settings[:puppet_codedir]} \
+    --sitelibdir=#{settings[:ruby_vendordir]} \
+    --configs \
+    --quick \
+    --man \
+    --mandir=#{settings[:mandir]}"]
   end
 
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
-
+  if platform.is_windows?
+    pkg.add_source("file://resources/files/windows/hiera.bat", sum: "bbe0a513808af61ed9f4b57463851326")
+    pkg.install_file "../hiera.bat", "#{settings[:link_bindir]}/hiera.bat"
+  end
   pkg.configfile File.join(settings[:puppet_codedir], 'hiera.yaml')
 
   pkg.link "#{settings[:bindir]}/hiera", "#{settings[:link_bindir]}/hiera" unless platform.is_windows?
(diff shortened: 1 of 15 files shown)


McdonaldSeanp:
This will require another test, will come tomorrow.

jcoconnor:
@McdonaldSeanp Have tested this in build (but not yet installed) and it runs right through.
One question I'm trying to work through is how do the puppet_shell.bat and other files get put in place in the original build - running the current build at the moment to trace this through.

McdonaldSeanp:
@jcoconnor the files are either created or pulled in via the install.rb scripts of each component. If you look in the actual puppet repo under ext/windows you'll see the shell/interactive/run_interactive scripts and install.rb will generate the original puppet.bat script here: https://github.com/puppetlabs/puppet/blob/master/install.rb#L433L440

McdonaldSeanp:
I checked with Rob Reynolds about this and he was good /cc @ferventcoder so I am ready for this to get merged

configs/components/hiera.rb
```diff
@@ -10,11 +10,22 @@
...
+    --sitelibdir=#{settings[:ruby_vendordir]} \
+    --configs \
+    --quick \
+    --man \
+    --mandir=#{settings[:mandir]}"]
   end
 
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
 
+  pkg.add_source("file://resources/files/windows/hiera.bat", sum: "bbe0a513808af61ed9f4b57463851326")
```
melissa: this needs an `if_windows?` check

shermdog:
:+1:

resources/files/windows/run_facter_interactive.bat
```diff
@@ -0,0 +1,5 @@
+@echo Running puppet on demand ...
```
ferventcoder: @McdonaldSeanp one alibi here....

resources/files/windows/puppet_interactive.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
+echo Running Puppet agent on demand ...
```
ferventcoder: @McdonaldSeanp this may be different versus the older one.

resources/windows/wix/shortcuts.wxs.erb
```diff
@@ -131,64 +131,6 @@
...
-          Id="ShortCutRunFacterIntBat"
-          Name="run_facter_interactive.bat"
-          Target="[INSTALLDIR]puppet\bin\run_facter_interactive.bat" />
-
-        <Shortcut
-          Id="ShortCutRunPuppetIntBat"
-          Name="run_puppet_interactive.bat"
-          Target="[INSTALLDIR]puppet\bin\run_puppet_interactive.bat" />
-      </Component>
-
```
ferventcoder: :+1:

PR #620, opened 2016-04-06
(PA-279)Fix OSX service restarts

shrug:


resources/osx/preinstall-extras
```diff
@@ -0,0 +1,16 @@
+## This works around an issue with the OSX service names in
+## puppet-agent. We're using a short service name intead of
+## the fully qualified name (puppet instead of com.puppetlabs.puppet)
+
+## Putting this workaround here until we can fix the service names in
+## the next major version, at which point the more generic logic in
+## vanagon will do the right thing
+
+if [ -n "$foundpkg" ]; then
```
underscorgan: This is intentionally changing the logic from before, where we were testing ```-z "$foundpkg"```?

shrug:
Testing builds of this now, still needs a bit of work

resources/osx/preinstall-extras
shrug: Worth a discussion on which form we prefer. We're using '-n' in vanagon, and we previously used '-z' in this project. 

Summoning @mckern - he has opinions

mckern: Do we care if `the length of STRING is nonzero` (`-n`) or `the length of STRING is zero` (`-z`)?

mckern: I generally prefer the form:

```bash
if [[ ${foundpkg:-UNDEFINED} == UNDEFINED ]]; then
```

I concede that this is not portable, and not intuitive if you're not deep in the bash minutia. I like it because it's extremely explicit. But the [Bash Pitfalls wiki entry has something to offer here](http://mywiki.wooledge.org/BashPitfalls#A.5B_-n_.24foo_.5D_or_.5B_-z_.24foo_.5D). It boils down to "it doesn't matter which form you use, as long as you remember to quote the variable"

shrug:
The generated preinstall looks like https://gist.github.com/shrug/0c2069c3fbb4fdd3202c63ba54af5b38

shrug:
and the postinstall: https://gist.github.com/shrug/b9375b405661d70481a73237ac91343b

shrug:
This is admittedly terrible, and will break (or have to be updated) if we add more services to the agent, prior to fixing the overall issue in the next major, but of all the fixes I considered this was the best of the worst. And, fortunately is only a stopgap.

shermdog:
:+1:

PR #621, opened 2016-04-07
(RE-7004) update facter cmake invocation for windows

McdonaldSeanp:
For the windows platform there are cmake flags that
provide variables to facter with paths in them. The
problem is we cannot specifiy any kind of path for
a windows build until after install time finishes
in the MSI. See RE-7004 for a better description

This commit updates the cmake invocation in facter.rb
so it no longer provides those flags for the windows
platform.


melissa:
:+1:

PR #622, opened 2016-04-08
(PA-282) Fix special_flags handling to restore builds on *nix

kylog:
Commit 7b4256ab28fcc535e70d1887c991d5525c9f88cc for RE-7004 introduced
a 'special_flags += <stuff>' stanza to separate some functionality
not needed on windows. Unfortunately, special_flags is not defined
at the point the stanza is executed which results in an exception.

This change simply initializes special_flags so it is always defined.

PR #623, opened 2016-04-08
(maint) fix facter failure with special_flags

McdonaldSeanp:
special_flags is a variable we use in the cmake
invocation of facter, it was never initialized
so for many platforms facter was trying to do a
"+=" to a non existing string, this commit fixes
that


underscorgan:
:+1: pending travis

underscorgan:
Dupe of #622

PR #629, opened 2016-04-11
(HUAWEI-15) Include cross-compiled dmidecode in huaweios builds

ScottGarman:
It's unlikely that dmidecode will ever be used on CloudEngine switches, but I have confirmed that the Yocto Project builds dmidecode for powerpc 32 and 64 bit platforms, and this will quiet a warning message that puppet spews during runs when the dmidecode binary is missing.


shermdog:
:+1:

PR #630, opened 2016-04-11
(maint) ensure rubygem-deep-merge is built before nokogiri

ScottGarman:
This fixes an intermittent build failure where rubygem-deep-merge
could be installed after nokogiri. As we've found, the only way to
keep nokogiri in line is to ensure it is the *last* gem installed.


shermdog:
:+1:

PR #633, opened 2016-04-14
(RE-7021) Update the puppet-msi with fixes

McdonaldSeanp:
This replaces "settings[:product_name]" with
"settings[:product_id]" to use the correct name for
names of registry entries

I cannot fix the batch file tickets RE-7018 and RE-7019 until the major change to paths in RE-7027 is complete. so this PR is now just one ticket worth of work

### DIFF
--- a/resources/windows/wix/registryEntries.wxs.erb
+++ b/resources/windows/wix/registryEntries.wxs.erb
@@ -13,7 +13,7 @@
 
         <RegistryKey
           Root="HKLM"
-          Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>"
+          Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_id] %>"
           Action="create" >
           <!-- This is the default (aka 'unnamed') key value of this path -->
           <RegistryValue
@@ -72,7 +72,7 @@
 
         <RegistryKey
           Root="HKLM"
-          Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_name] %>"
+          Key="SOFTWARE\<%= settings[:company_name] %>\<%= settings[:product_id] %>"
           ForceCreateOnInstall="yes" >
 
           <RegistryValue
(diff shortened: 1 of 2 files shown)


resources/files/windows/run_facter_interactive.bat
```diff
@@ -1,5 +1,2 @@
-@echo Running puppet on demand ...
-@echo off
-SETLOCAL
```
jcoconnor: Why are you removing these 3 lines in the file, but leaving them in ```run_puppet_interactive.bat``` ?

jcoconnor:
@McdonaldSeanp Being picky - I know these are small changes, but I would prefer them to be broken down into individual commits with appropriate ticket numbers.

resources/files/windows/run_facter_interactive.bat
McdonaldSeanp: Only because the original file run_facter_interactive did not have them but run_puppet_interactive did. 

The only reason was to be the same as the originals

jcoconnor: Ignore - Have read the ticket again.

McdonaldSeanp: Also I am realizing that I can't even fix these files until after the change in paths is done

resources/files/windows/run_puppet_interactive.bat
```diff
@@ -2,4 +2,4 @@
 @echo off
 SETLOCAL
 call "%~dp0environment.bat" %0 %*
-@elevate.exe "%~dp0puppet_interactive.bat"
+elevate.exe "%~dp0puppet_interactive.bat"
```
ferventcoder: At this point is it okay to assume it is on the path?

ferventcoder:
also agree on splitting the changes out.

resources/files/windows/run_puppet_interactive.bat
McdonaldSeanp: Yes, it is.

PR #634, opened 2016-04-14
(PA-274) add sles-11-s390x platform

ScottGarman:
We cross-compile this platform on a sles-11-x86_64 pooler system.

It blends! The only work left to do is to include puppet-ca-bundle, which needs some work in a cross-compiled environment.


configs/components/curl.rb
```diff
@@ -4,7 +4,12 @@
   pkg.url "http://buildsources.delivery.puppetlabs.net/curl-#{pkg.get_version}.tar.gz"
 
   pkg.build_requires "openssl"
-  pkg.build_requires "puppet-ca-bundle"
+  pkg.build_requires "puppet-ca-bundle" unless platform.architecture == "s390x"
```
stahnma: Can you put a comment here about how this is currently a known issue and should be fixed?

configs/components/facter.rb
```diff
@@ -112,7 +112,7 @@
     skip_blkid = 'OFF'
   elsif platform.is_rpm?
     if (platform.is_el? && platform.os_version.to_i >= 6) || (platform.is_sles? && platform.os_version.to_i >= 11) || platform.is_fedora?
-      pkg.build_requires "libblkid-devel"
+      pkg.build_requires "libblkid-devel" unless platform.architecture == "s390x"
```
stahnma: I *think* rubocop wants || vs or

configs/components/virt-what.rb
```diff
@@ -10,7 +10,7 @@
     requires "util-linux"
   end
   if platform.name =~ /^sles-(10|11)-.*$/
-    requires "pmtools"
+    requires "pmtools" unless platform.architecture == "s390x"
```
stahnma: Why is it not included on s390?

configs/platforms/sles-11-s390x.rb
```diff
@@ -0,0 +1,12 @@
+platform "sles-11-s390x" do |plat|
```
stahnma: Maybe a comment at the top that explains this is cross compiled.

resources/files/rbconfig-s390x-linux-gnu.rb
```diff
@@ -0,0 +1,268 @@
+
+# This file was created by PuppetAgent for HuaweiOS (Debian 8 PPC).  Any
```
stahnma: This seems incorrect for SLES s390x

resources/files/rbconfig-s390x-linux-gnu.rb
```diff
@@ -0,0 +1,268 @@
...
+# changes made to this file will be lost the next time ruby is built.
```
stahnma: This file is generated by building Ruby on the platform (minus the cross compile, have we done that?)

configs/components/facter.rb
ScottGarman: rubocop isn't complaining about this. From what I've just looked up, the differences between using "or" vs. "||" only come into play when doing things like using them in assignment, and as if conditionals they are synonymous. Maybe @mckern can weigh in here?

mckern: `||` and `or` is basic precedence; how will the conditional be evaluated from left to right? But as a rule, always use `||`. It's consistent and predictable.

ScottGarman:
I've updated the PR with all feedback so far. Just to be paranoid-level safe, I'm running it through jenkins:

https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/121/

configs/components/augeas.rb
```diff
@@ -37,7 +39,11 @@
       pkg.build_requires 'runtime'
     end
 
-    pkg.build_requires 'pkgconfig'
+    if platform.architecture == "s390x"
+      pkg.environment "PATH" => "/opt/pl-build-tools/bin:$$PATH:#{settings[:bindir]}"
+      pkg.environment "CFLAGS" => settings[:cflags]
+      pkg.environment "LDFLAGS" => settings[:ldflags]
```
mckern: Probably out of scope for this PR, but why don't we set `:CFLAGS` and `:LDFLAGS` for everything everywhere all the time forever?

configs/components/curl.rb
```diff
@@ -4,7 +4,14 @@
   pkg.url "http://buildsources.delivery.puppetlabs.net/curl-#{pkg.get_version}.tar.gz"
 
   pkg.build_requires "openssl"
-  pkg.build_requires "puppet-ca-bundle"
+  # TODO: remove the exception here once puppet-ca-bundle can be
+  # installed in cross-compiled environments
```
mckern: We can fix this. Not in this PR but we can fix this. Can you please sign/date this comment?

configs/components/augeas.rb
ScottGarman: Totally agree, this is something to discuss next week when we have a work session to talk about refactoring the p-a config files, where I'll be dumping some ideas out of my head that have been rattling around for a while.

ScottGarman:
Ok, final tweaks made, and I fixed the silly md5 issue on the rbconfig file that caused the jenkins build to fail for sles-11-s390x, and rebuilt it locally to confirm it works. Ready for merge now.

configs/platforms/sles-11-s390x.rb
mckern: Seconded on cross-compile comment.

ScottGarman: I followed the example I used for huaweios and the comment is just above the line that sets vmpooler_template.

mckern: Dig it

mckern:
:+1: 
I see things we can polish, smooth over, or remove but they're deeper issues that we should in the P-A codebase. Great job.

configs/components/leatherman.rb
```diff
@@ -94,7 +94,7 @@
 
   # Make test will explode horribly in a cross-compile situation
   # Tests will be skipped on AIX until they are expected to pass
-  if platform.architecture == 'sparc' || platform.is_aix? || platform.is_huaweios?
+  if platform.architecture == 'sparc' || platform.architecture == 's390x' || platform.is_aix? || platform.is_huaweios?
```
shermdog: Did we ever add an is_cross_compile flag?

ScottGarman: No. Once this gets merged I intend to start working on that. Then the refactoring can begin.

PR #637, opened 2016-04-15
(maint) The puppet component needs to replace pe-puppetserver-common

underscorgan:
Both packages own /etc/puppetlabs/puppet/auth.conf, so this is needed
to ensure that packages get installed/uninstalled in the correct order
on upgrade.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -14,6 +14,7 @@ component "puppet" do |pkg, settings, platform|
   pkg.replaces 'pe-cloud-provisioner'
   pkg.replaces 'pe-puppet-enterprise-release', '4.0.0'
   pkg.replaces 'pe-agent'
+  pkg.replaces 'pe-puppetserver-common', '4.0.0'
 
   if platform.is_deb?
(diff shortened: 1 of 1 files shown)


configs/components/puppet.rb
```diff
@@ -14,6 +14,7 @@
   pkg.replaces 'pe-cloud-provisioner'
   pkg.replaces 'pe-puppet-enterprise-release', '4.0.0'
   pkg.replaces 'pe-agent'
+  pkg.replaces 'pe-puppetserver-common', '4.0.0'
```
stahnma: does pe-puppetserver already replace this? What happens if both do?

shrug: We talked about this a bit, I *think* that since it only replaces, doesn't provide then it's fine. But we'll find out :)

underscorgan: it does, but at least trusty didn't mind.

mckern: streams_wide-590x244

shrug:
:+1:  pending test build

underscorgan:
This works with upgrades for both trusty and el6 \o/

PR #638, opened 2016-04-15
(PA-186) components should use upstream and rewrite urls

stahnma:
Prior to this commit, for anybody not on the Puppet, Inc VPN a person
had to manually gather sources for ~20 items and place them at a
location web accessable and then modify component files to point to that
location just to get started on a build.

Vanagon has had http rewrite rule magic in it for a while. This commit
simply puts the upstream URL in place for primary url on most
components. This should remove the first hour or two of work from any
community member attempting to build out the puppet-agent.

If you're a community member, now you'll simply have to comment out the
rewrite line in configs/projects/puppet-agent.rb

This will not allow community members to build hauwei, or aix due to
other things required that our only on our build infrastructure.

Windows will not work with just this commit, as other changes are
required. That will be addressed in another PR.

Yet another note: This doesn't help with things requiring pl-gcc, pl-* and the like.

### DIFF
--- a/configs/components/nssm.json
+++ b/configs/components/nssm.json
@@ -1,5 +1,0 @@
-{
-  "version": "2.24",
-  "md5sum": "0fa251d152383ded6850097279291bb0",
-  "url": "http://buildsources.delivery.puppetlabs.net/windows/nssm/nssm-2.24.zip"
-}
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -1,7 +1,7 @@
 component 'augeas' do |pkg, settings, platform|
   pkg.version '1.4.0'
   pkg.md5sum 'a2536a9c3d744dc09d234228fe4b0c93'
-  pkg.url "http://buildsources.delivery.puppetlabs.net/#{pkg.get_name}-#{pkg.get_version}.tar.gz"
+  pkg.url "http://download.augeas.net/augeas-#{pkg.get_version}.tar.gz"
 
   pkg.replaces 'pe-augeas'
   pkg.apply_patch 'resources/patches/augeas/osx-stub-needed-readline-functions.patch'
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -1,7 +1,7 @@
 component 'curl' do |pkg, settings, platform|
   pkg.version '7.46.0'
   pkg.md5sum '230e682d59bf8ab6eca36da1d39ebd75'
-  pkg.url "http://buildsources.delivery.puppetlabs.net/curl-#{pkg.get_version}.tar.gz"
+  pkg.url "https://curl.haxx.se/download/curl-#{pkg.get_version}.tar.gz"
 
   pkg.build_requires "openssl"
   pkg.build_requires "puppet-ca-bundle"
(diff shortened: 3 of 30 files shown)


mckern:
flat 550x550 075 f u4

Gonna clone this and test it!

stahnma:
Two ways to test this.
1. Try building on pl-infrastructure and ensuring it's pulls from buildsources.
2. Try removing the rewrite rule and see that it pulls and builds (and doesn't asplode) when using upstream URLs.

configs/components/cfpropertylist.rb
```diff
@@ -1,7 +1,7 @@
 component "cfpropertylist" do |pkg, settings, platform|
   pkg.version "2.2.7"
   pkg.md5sum "be2fc619084422cc28cb68f6084bea91"
-  pkg.url "http://buildsources.delivery.puppetlabs.net/CFPropertyList-2.2.7.tar.gz"
+  pkg.url "https://rubygems.org/downloads/CFPropertyList-#{pkg.get_version}.gem"
```
mckern: We're not carrying the gem; we've got the tarball up on `buildsources`. I'm trying this with a couple different URLs but this broke the build when I tested it.

mckern: Oh this is amazing. Vanagon can't follow redirects.

```
Downloading file 'cfpropertylist-2.2.7.tar.gz' from url 'http://buildsources.delivery.puppetlabs.net/cfpropertylist-2.2.7.tar.gz'
Error: 301. Unable to get source from http://buildsources.delivery.puppetlabs.net/cfpropertylist-2.2.7.tar.gz
/Users/mckern/Documents/Repositories/puppet-agent/.bundle/gems/ruby/2.1.0/gems/vanagon-0.6.0/lib/vanagon/component/source/http.rb:62:in `block (2 levels) in download'
```

mckern:
This will benefit from https://github.com/puppetlabs/vanagon/pull/366

configs/components/cfpropertylist.rb
stahnma: That might actually be me just messing up the upstream url. I mean, I'm glad the redirect thing is now fixed, but I should probably just point to the tarball upstream, if that's possible.

mckern: The tarball URL looks like this: `https://github.com/ckruse/CFPropertyList/archive/cfpropertylist-#{pkg.get_version}.tar.gz`

Note that it's lowercase. The tarball internally is not, for one reason or another; that, coupled with GitHub's 302s on 302s on 302s, is how I found the "Vanagon doesn't handle redirects" problem.

mckern: Ah, when we snatched the source tarball it probably came from a Debian or RHEL package back in the day and we preserved the case over time because :skull::skull::skull:.

mckern:
Hey @stahnma I took a look at that CFPropertylist annoyance thing and made https://github.com/puppetlabs/puppet-agent/pull/650 for it.

stahnma:
Has a single bug around cfproperty list, will fix soon.

McdonaldSeanp:
All PRs should wait until we get a passing master again

stahnma:
Closing for now with sadness.

mckern:
Broadly 👍  on this; one step closer to setting the rewrite engine adrift on an ice floe.

configs/components/nssm.json
```diff
@@ -1,5 +0,0 @@
-{
```
MikaelSmith: You also need to change bin/build-windows.rb. Windows bulids are currently broken.

PR #641, opened 2016-04-21
(RE-7032) Fix puppet-ca-bundle installs on cross-compiled platforms

ScottGarman:
And re-enable it on sles-11-s390x.


underscorgan:
👍

PR #642, opened 2016-04-21
Revert to PTFW File Paths

McdonaldSeanp:
We are doing a major path change to the current windows Puppet-Agent work. The change itself is actually a reversion to old standards for MSIs. See https://tickets.puppetlabs.com/browse/RE-7027 for details on why we are doing this

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -28,7 +28,7 @@ component "facter" do |pkg, settings, platform|
   pkg.build_requires "openssl"
 
   if platform.is_windows?
-    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:ruby_dir]}/bin):$$(cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0"
   else
     pkg.environment "PATH" => "#{settings[:bindir]}:$$PATH"
   end
@@ -134,14 +134,14 @@ component "facter" do |pkg, settings, platform|
 
   make = platform[:make]
 
-  special_flags = ""
+  special_flags = " -DCMAKE_INSTALL_PREFIX=#{settings[:prefix]} "
 
   # cmake on OSX is provided by brew
   # a toolchain is not currently required for OSX since we're building with clang.
   if platform.is_osx?
     toolchain = ""
     cmake = "/usr/local/bin/cmake"
-    special_flags = "-DCMAKE_CXX_FLAGS='#{settings[:cflags]}'"
+    special_flags += "-DCMAKE_CXX_FLAGS='#{settings[:cflags]}'"
   elsif platform.is_cross_compiled_linux?
     ruby = "#{settings[:host_ruby]} -r#{settings[:datadir]}/doc/rbconfig.rb"
     toolchain = "-DCMAKE_TOOLCHAIN_FILE=/opt/pl-build-tools/#{settings[:platform_triple]}/pl-build-toolchain.cmake"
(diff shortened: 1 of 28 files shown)


jcoconnor:
Changed Title

McdonaldSeanp:
Many of the changes I made to add things like PATH=> to the ruby components were kind of wholesale. If there are any components that aren't even used in windows, let me know

McdonaldSeanp:
Test build @ http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/386/

configs/components/hiera.rb
```diff
@@ -9,23 +9,29 @@
...
+  end
+
   pkg.install do
     ["#{settings[:host_ruby]} install.rb \
-    --ruby=#{File.join(settings[:bindir], 'ruby')} \
-    --bindir=#{settings[:bindir]} \
-    --configdir=#{settings[:puppet_codedir]} \
-    --sitelibdir=#{settings[:ruby_vendordir]} \
     --configs \
-    --quick \
```
shrug: I assume losing '--quick' is intentional?

McdonaldSeanp: yep, it's not even an option in this install script, I'm not sure how this was working before but theres no quick option under https://github.com/puppetlabs/hiera/blob/master/install.rb#L99L122

jcoconnor:
@McdonaldSeanp Going to run this through a build and some testing and will come back to you with comments.

jcoconnor:
@McdonaldSeanp As discussed, this builds fine and installs on x64, but doesn't build on x86.
The msi installs and the services come up which indicates that you probably have most of the work done.
I will do a machine-machine comparison between this package and the original non-vanagon MSI to check for correct placement of files etc.

McdonaldSeanp:
New test build at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/

McdonaldSeanp:
http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/393/ is green, ready for merge

jcoconnor:
@McdonaldSeanp Can you rebase it please.

jcoconnor:
Otherwise 👍 for the x64 bit - still need to check the x86

jcoconnor:
@McdonaldSeanp As discussed in e-mail, X86 and X64 are building to end, but throwing an MSI error.

```
wixvariables.wxs
cd /var/tmp/tmp.hX2gOD3AJp/wix/wixobj; "$WIX/bin/light.exe" -v -cultures:en-us -ext WiXUtilExtension -ext WixUIExtension -b $(cygpath -aw /var/tmp/tmp.hX2gOD3AJp) -loc $(cygpath -aw /var/tmp/tmp.hX2gOD3AJp/wix/localization/puppet_en-us.wxl) -out $(cygpath -aw /var/tmp/tmp.84oAHgresm/output/windows/PC1/x64/puppet-agent-1.4.2.408.g5292c93.1-x64_VANAGON.msi) *.wixobj
Windows Installer XML Toolset Linker version 3.10.2.2516
Copyright (c) Outercurve Foundation. All rights reserved.

C:\cygwin64\var\tmp\tmp.hX2gOD3AJp\wix\appdatafiles.wxs(88) : error LGHT0103 : The system cannot find the file 'SourceDir\CommonAppDataFolder\Puppetlabs\puppet\etc\hiera.yaml'.
make: *** [puppet-agent-1.4.2.408.g5292c93.1-x64_VANAGON.msi] Error 103
Makefile:10: recipe for target 'puppet-agent-1.4.2.408.g5292c93.1-x64_VANAGON.msi' failed
An error was encountered evaluating block. Retrying..
Executing '(cd /var/tmp/tmp.84oAHgresm; /usr/bin/make)' on 'Administrator@tlahg55adqu2dtl.delivery.puppetlabs.net'
```

McdonaldSeanp:
Well, http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/ is all green sans solaris 10, and it didn't fail when I built it myself sooo........ green!

McdonaldSeanp:
Yeah everything is def green, http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/399/ combined with http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/398/ is a green board

McdonaldSeanp:
I am going to block this until we get a passing master of P-A

jcoconnor:
@McdonaldSeanp Clean builds on X64 and X86 - next up on Monday - some testing ;)

jcoconnor:
@McdonaldSeanp I have done a folder compare of the ```Program Files\Puppet Labs``` directory (just checking file placement only (i.e. not interested in dates/sizes as I expect these to differ).

There are some significant variances which I have attached in the report below (HTML Files enclose in Zip File).
[PTFW.zip](https://github.com/puppetlabs/puppet-agent/files/254788/PTFW.zip)

If you have any issues reading it, let me know.
I produced the report by remotely mounting the directories (using ```sshfs```) and running ```Beyond Compare```

McdonaldSeanp:
@jcoconnor None of the differences I see are unexpected. At this point the one thing I'm worried about is missing dlls in the ruby/bin directory, but that is a seperate problem I have ticketed in RE-7262 and I don't believe should be a blocker to this ticket. 

What I would like is to merge this and treat it as "passing" since the build completes and we can run puppet at least. That way we can get this work on the pipelines so it can go in to the production builds with the new path structure. Which should be okay since those builds still won't go through CI.

configs/components/facter.rb
```diff
@@ -155,19 +155,20 @@
     cmake = "/opt/pl-build-tools/i386-pc-solaris2.#{platform.os_version}/bin/cmake"
 
     # FACT-1156: If we build with -O3, solaris segfaults due to something in std::vector
-    special_flags = "-DCMAKE_CXX_FLAGS_RELEASE='-O2 -DNDEBUG'"
+    special_flags += " -DCMAKE_CXX_FLAGS_RELEASE='-O2 -DNDEBUG' "
```
shrug: Not really a fan of having to pre- and append whitespace to these.

Also notice we don't do it above in the osx conditional. Should be consistent at least. We can work on concatenating with spaces later I think/.

shrug:
Generally 👍  except for the whitespace comment

configs/components/facter.rb
McdonaldSeanp: So this is where alot of this logic gets SUPER hairy. The osx conditional is actually not concatenated on purpose, as it is supposed to overload the variable, not append to it. What I would like to do is come up with some convention for this concatenation as part of the larger puppet-agent cleanup effort as opposed to doing it here. This PR is already large enough.

ferventcoder:
Generally a 👍 . The actual details will come out better in an MSI build :D

jcoconnor:
Early burndown for next sprint.

PR #643, opened 2016-04-21
(maint) Add JDK Support for Xenial and f23

melissa:
We need to build a facter jar for use with puppetserver. This requires
that we install java in order to create that jar.


haus:
:+1: from me. should dtrt.

melissa:
Don't merge until after 1.4.2

PR #645, opened 2016-04-25
(RE-7099) Update Ruby to 2.1.9 for sles-11-s390x

ScottGarman:
Our switch to ruby 2.1.9 happened in the midst of me working on
this platform, so bumping it apart from the other platform rbconfigs.


geoffnichols:
👍

PR #646, opened 2016-04-27
(maint) merge-back for work for CVE-2016-2785

underscorgan:


configs/components/cpp-pcp-client.json
```diff
@@ -1 +1 @@
-{"url": "git://github.com/puppetlabs/cpp-pcp-client.git", "ref": "2a0cbb5dbab8c967272b9cd04a0a923b6b42b73f"}
+{"url": "git://github.com/puppetlabs/cpp-pcp-client.git", "ref": "refs/tags/1.1.2"}
```
shrug: Do we want the jsons? Or should we exclude those from merges?

underscorgan: That sha lines up with the tag. Left SHAs for things that have promoted since the release branch (puppet/facter)

underscorgan: which was handled in https://github.com/puppetlabs/puppet-agent/pull/646/commits/4116dfaa22a95abd246822e8e87fe3beecf81040

shrug:
👍

PR #647, opened 2016-04-27
(RE-7029) update the paths to windows shortcuts

McdonaldSeanp:
The batch files that the P-A shortcuts point to live under INSTALLDIR/bin, not
INSTALLDIR/puppet/bin. this needs to be updated


shermdog:
:+1:

PR #648, opened 2016-04-27
(PA-308) update timeout for windows

McdonaldSeanp:
windows builds of puppet agent are longer than the standard 1 hour set by the
new vanagon retry logic. we need to extend the timeout.

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -253,4 +253,6 @@ project "puppet-agent" do |proj|
   proj.directory proj.logdir unless platform.is_windows?
   proj.directory proj.piddir unless platform.is_windows?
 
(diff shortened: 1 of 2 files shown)


configs/projects/puppet-agent.rb
```diff
@@ -253,4 +253,6 @@
   proj.directory proj.logdir unless platform.is_windows?
   proj.directory proj.piddir unless platform.is_windows?
 
+  proj.timeout 7200 if platform.is_windows?
```
mckern: You won't monkey patch Vanagon from inside a component  to make the timeout exponentially increase on every failure. You're scared.

shrug:
👍

mckern:
sadface_okay_meme_face-s331x283-344201

PR #650, opened 2016-04-28
(MAINT) Use a SHA reference for CFPropertylist

mckern:
The tarball we use to package this is of unknown provenance and the mixed case has been sort of an annoyance to use with Vanagon's rewrite engine. By switching to a git SHA (specifically for the version we're already using), we should hopefully avoid both of these problems.

### DIFF
--- a/configs/components/cfpropertylist.json
+++ b/configs/components/cfpropertylist.json
@@ -1,0 +1,1 @@
+{"url": "git@github.com:ckruse/CFPropertyList.git", "ref": "2.2.7"}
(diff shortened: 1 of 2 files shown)


underscorgan:
👍

configs/components/cfpropertylist.json
```diff
@@ -0,0 +1 @@
+{"url": "git@github.com:ckruse/CFPropertyList.git", "ref": "2.2.7"}
```
shrug: Can we add this to our git mirror? I had adding a dep to github

mckern: had or hate?

Also, I'm not opposed but does the rewrite engine handle git sources?

shrug: hate, but if we're not enforcing that for other components (and we're not) then 👍

stahnma: The rewrite engine does not support git yet, however, we should still add this to the mirror.

haus:
I think you probably wanted 'cfpropertylist-2.2.7' as the tag.

mckern:
Urgh. You're right.

PR #658, opened 2016-05-04
(maint) Add xenial to build targets

melissa:
It seems that we need to specify deb_targets in order to have packages
promote successfully into PE. The promotion job takes this list of
platforms and will only promote the packages that correspond to this
list somehow. I could very well be wrong in this assertion, but that is
certainly how things seem to work at the moment.


underscorgan:
👍

shrug:
👍

PR #662, opened 2016-05-06
(PA-285) Branding name change

jpogran:
Update references from puppetlabs to puppet

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -22,11 +22,11 @@ project "puppet-agent" do |proj|
       proj.setting(:RememberedInstallDirRegKey, "RememberedInstallDir")
     end
     proj.setting(:links, {
-        :HelpLink => "http://links.puppetlabs.com/customer-support-foss",
-        :CommunityLink => "http://links.puppetlabs.com/windows-installer-starting-out",
-        :ForgeLink => "http://links.puppetlabs.com/forge-windows",
-        :NextStepLink => "http://links.puppetlabs.com/windows-installer-next-steps-foss",
-        :ManualLink => "http://links.puppetlabs.com/windows-manual-foss",
+        :HelpLink => "http://puppet.com/services/customer-support",
+        :CommunityLink => "http://docs.puppet.com",
+        :ForgeLink => "http://forge.puppet.com",
+        :NextStepLink => "http://docs.puppet.com",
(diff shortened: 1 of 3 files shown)


shrug:
Adding do not merge labels until we get a clean ci run

stahnma:
👍

configs/projects/puppet-agent.rb
```diff
@@ -22,11 +22,11 @@
       proj.setting(:RememberedInstallDirRegKey, "RememberedInstallDir")
     end
     proj.setting(:links, {
-        :HelpLink => "http://links.puppetlabs.com/customer-support-foss",
-        :CommunityLink => "http://links.puppetlabs.com/windows-installer-starting-out",
-        :ForgeLink => "http://links.puppetlabs.com/forge-windows",
-        :NextStepLink => "http://links.puppetlabs.com/windows-installer-next-steps-foss",
-        :ManualLink => "http://links.puppetlabs.com/windows-manual-foss",
+        :HelpLink => "http://puppet.com/services/customer-support",
+        :CommunityLink => "http://docs.puppet.com",
```
Iristyle: I believe `CommunityLink` should be https://puppet.com/community

configs/projects/puppet-agent.rb
```diff
@@ -22,11 +22,11 @@
...
+        :ForgeLink => "http://forge.puppet.com",
+        :NextStepLink => "http://docs.puppet.com",
```
Iristyle: So previously `windows-installer-next-steps-pe` would be going to https://docs.puppet.com/pe/latest/quick_start_install_agents_windows.html 

Since there is no longer a distinction between PE and FOSS, I think that it should go there as well.

Other reasonable options:

* https://docs.puppet.com/puppet/latest/reference/install_windows.html
* https://docs.puppet.com/puppet/latest/reference/services_agent_windows.html
* https://docs.puppet.com/puppet/latest/reference/services_commands_windows.html

configs/projects/puppet-agent.rb
```diff
@@ -22,11 +22,11 @@
...
+        :ManualLink => "http://docs.puppet.com",
```
Iristyle: We probably want this to go to https://docs.puppet.com/puppet/latest/reference/

Iristyle:
Follow up PR at #674

PR #665, opened 2016-05-09
Revert "(RE-7221) Update to OpenSSL 1.0.2h"

shrug:
Reverts puppetlabs/puppet-agent#659

The openssl update is causing a test failure, reverting until we fully understand the root cause.


underscorgan:
👍

PR #666, opened 2016-05-10
Re-add openssl 1.0.2h and add 'no-ssl2-method' configure flag

shrug:
Between openssl 1.0.2g and 1.0.2h, the behavior of the no-ssl2 configure
flag was changed. Previously, no-ssl2 disabled all ssl2 methods. In 1.0.1h,
the methods are no longer disabled, they just return NULL, unless the new
configure flag 'no-ssl2-method' is added. This behavior change was
breaking our tests that ensure ssl2 is disabled.

This adds the no-ssl2-method flag to the openssl build, restoring
the same behavior seen in 1.0.2g: specifically
```
ruby -ropenssl -e "puts OpenSSL::SSL::SSLContext::METHODS.include?(:SSLv2)"
```
will return 'false'


underscorgan:
👍

PR #670, opened 2016-05-11
(maint) Bump to pre-release cpp-pcp-client

MikaelSmith:
Bump to cpp-pcp-client#stable, the pre-release code for cpp-pcp-client.


geoffnichols:
👍

PR #674, opened 2016-05-11
(PA-285) Branding name change

jpogran:
Update references from puppetlabs to puppet


configs/projects/puppet-agent.rb
```diff
@@ -23,10 +23,10 @@
     end
     proj.setting(:links, {
         :HelpLink => "http://puppet.com/services/customer-support",
-        :CommunityLink => "http://docs.puppet.com",
+        :CommunityLink => "http://docs.puppet.com/community",
```
Iristyle: puppet.com/community not docs.puppet.com/community

PR #678, opened 2016-05-13
(maint) Enable tests, use check on all CMake projects

MikaelSmith:
Enable tests on cpp-pcp-client and pxp-agent (on platforms where they're
expected to pass). Update facter and leatherman to use the
Component#check method for tests.

### DIFF
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -69,6 +69,14 @@ component "cpp-pcp-client" do |pkg, settings, platform|
     ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
   end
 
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Testing with http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/412/.

stahnma:
Needs a rebase plz.

MikaelSmith:
Done.

stahnma:
Do we have any idea how long this will extend the build times?

stahnma:
👍 but have a few concerns on duration of builds.

MikaelSmith:
The unit tests take ~5-10 seconds for each project. So maybe +20 seconds?

PR #680, opened 2016-05-18
(PA-347) Put back nssm.json

stahnma:
nssm.json was removed because the contents of that file were moved into
nssm.rb which vanagon uses. Currently, however, the Windows builds still
use the bin/build-windows command which needs to read information about
nssm from the json file.

Once we move fully to a pure-vanagon build of the Windows MSI, nssm.json
should be removed.


MikaelSmith:
👍

PR #683, opened 2016-05-18
(maint) Bump leatherman to 0.7.0

branan:
Facter is about to require this version


branan:
Not sure why I opened two of these

PR #685, opened 2016-05-20
(RE-7020) Remove extraneous batch files from bin dirs

McdonaldSeanp:
Now that puppet-agent carries batch files inside it's own project, we need to
specify to the component installers that they do not need to install them.


geoffnichols:
👍

underscorgan:
does it blend?

PR #686, opened 2016-05-23
(RE-7262) Update runtime to pull in dlls + update components to use ruby_bindir

McdonaldSeanp:
Ruby links against several DLLs during its own compilation. Some of these DLLs
    are also runtime deps, and need to be pulled in after compilation so they will
    be included in the ruby we vendor

### DIFF
--- a/configs/components/ruby.rb
+++ b/configs/components/ruby.rb
@@ -147,12 +147,15 @@ component "ruby" do |pkg, settings, platform|
     pkg.build_requires "pl-libffi-#{platform.architecture}"
     pkg.build_requires "pl-pdcurses-#{platform.architecture}"
 
-    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:bindir]}):$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:tools_root]}/include):$$(cygpath -u #{settings[:bindir]}):$$(cygpath -u #{settings[:ruby_bindir]}):$$(cygpath -u #{settings[:includedir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
+
     # So we need to pass -static-libgcc to the compiler, but we cannot pass -static-libgcc as
     # a regular flag, for information as to why: http://www.mingw.org/wiki/HOWTO_Sneak_GCC_Switches_Past_Libtool
     # in order to actually get gcc to honor the flag you need to set CC specifically with the flag in it
     pkg.environment "CC" => "gcc -static-libgcc"
+    pkg.environment "optflags" => settings[:cflags]
+    pkg.environment "LDFLAGS" => settings[:ldflags]
 
     special_flags = " CPPFLAGS='-DFD_SETSIZE=2048' debugflags=-g --prefix=#{settings[:ruby_dir]} --with-opt-dir=#{settings[:prefix]} "
   end
--- a/configs/components/runtime.rb
+++ b/configs/components/runtime.rb
@@ -37,7 +37,15 @@ component "runtime" do |pkg, settings, platform|
 
     # Curl is dynamically linking against zlib, so we need to include this file until we
     # update curl to statically link against zlib
-    pkg.install_file "#{settings[:tools_root]}/bin/zlib1.dll", "#{settings[:bindir]}/zlib1.dll"
+    pkg.install_file "#{settings[:tools_root]}/bin/zlib1.dll", "#{settings[:ruby_bindir]}/zlib1.dll"
+
+    # gdbm, yaml-cpp and iconv are all runtime dependancies of ruby, and their libraries need
+    # To exist inside our vendored ruby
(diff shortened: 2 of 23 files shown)


mckern:
Oh ha. Sorry Sean.

McdonaldSeanp:
np

configs/components/ruby.rb
```diff
@@ -147,12 +147,16 @@
     pkg.build_requires "pl-libffi-#{platform.architecture}"
     pkg.build_requires "pl-pdcurses-#{platform.architecture}"
 
-    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:bindir]}):$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:tools_root]}/include):$$(cygpath -u #{settings[:bindir]}):$$(cygpath -u #{settings[:includedir]}):$$PATH"
     pkg.environment "CYGWIN" => settings[:cygwin]
+    pkg.environment "LDFLAGS" => "-Wl,-rpath=/opt/puppetlabs/puppet/lib"
```
mckern: Should this path actually be hardcoded like this?

McdonaldSeanp: no

configs/components/ruby.rb
```diff
@@ -147,12 +147,15 @@
     pkg.build_requires "pl-libffi-#{platform.architecture}"
     pkg.build_requires "pl-pdcurses-#{platform.architecture}"
 
-    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:bindir]}):$$PATH"
+    pkg.environment "PATH" => "$$(cygpath -u #{settings[:gcc_bindir]}):$$(cygpath -u #{settings[:tools_root]}/bin):$$(cygpath -u #{settings[:tools_root]}/include):$$(cygpath -u #{settings[:bindir]}):$$(cygpath -u #{settings[:includedir]}):$$PATH"
```
jcoconnor: Curious as to why you aren't using ```settings[:ruby_bindir]``` here.

jcoconnor:
@McdonaldSeanp Looks ok except for one comment in the code.

configs/components/ruby.rb
McdonaldSeanp: There's actually no reason, and I'm now realizing that some of the ruby deps are put in ruby_bindir before ruby gets installed there so It likely does need to be there.

configs/components/runtime.rb
```diff
@@ -37,7 +37,17 @@
 
     # Curl is dynamically linking against zlib, so we need to include this file until we
     # update curl to statically link against zlib
-    pkg.install_file "#{settings[:tools_root]}/bin/zlib1.dll", "#{settings[:bindir]}/zlib1.dll"
+    pkg.install_file "#{settings[:tools_root]}/bin/zlib1.dll", "#{settings[:ruby_bindir]}/zlib1.dll"
+
+    # gdbm, yaml-cpp and iconv are all runtime dependancies of ruby, and their libraries need
+    # To exist inside our vendored ruby
+    # SPECIAL NOTE::: libffi is uneccessary in our vendored ruby here, because we install ffi
```
Iristyle: ^ Be careful about how the above is worded.  Our FFI gem does include it's own DLL (which as we discovered is an older version of `libFFI` than what Ruby bundles), but that doesn't mean that the Ruby vendored one is unnecessary.  It is used by Ruby's [`Fiddle module`](http://ruby-doc.org/stdlib-2.0.0/libdoc/fiddle/rdoc/Fiddle.html), which we've always shipped.

In other words, while we don't use `Fiddle` in our code, 3rd parties / end users could be using it in custom type and provider code, etc.  The chances are slim that that's the case I believe, but if we choose to later remove `libffi-6.dll` from our custom version, we should also remove `Fiddle` from Ruby, and it would need to be on a breaking version boundary.  I'm not sure it's worth the effort to do that.

McdonaldSeanp: Sorry, the comment is out of date

Iristyle:
Note - still deciding on whether we want to include `libyaml-0-2.dll` as we have in the past.  We haven't built an embedded libyaml on Windows before (based on the vendored code in https://github.com/ruby/ruby/tree/ruby_2_1/ext/psych/yaml), but that's what we're doing on other platforms.

The safer (but less consistent with other platform) approach is to not use the `bundled-libyaml` switch when building Ruby on Windows, and to still include / use the version that we have in the past at http://dl.bintray.com/oneclick/OpenKnapsack/x64/:libyaml-0.1.6-x64-windows.tar.lzma

We assume that the code used for the libyaml is at the `0.1.6` tag at  https://bitbucket.org/xi/libyaml/src/bce8b60f0b9a?at=0.1.6


Based on that assumption @McdonaldSeanp has produced the diff https://gist.github.com/McdonaldSeanp/3c55ee99e39d72ea4e699459ef4030d1 which we are examining for differences.

Iristyle:
It appears that the difference in Ruby to libyaml was introduced in https://github.com/ruby/ruby/commit/2d9d5c67f25ad85a82e5947ddfef2c67d6614449

Based on the diff, it looks like we can go forward with this change.

Iristyle:
New builds will be noted in https://tickets.puppetlabs.com/browse/RE-7216

McdonaldSeanp:
Yep, builds available at http://builds.puppetlabs.lan/vanagon-PA-windows/windows/PC1/

PR #687, opened 2016-05-25
(PA-241) AIO_AGENT_VERSION excludes git sha

MikaelSmith:
Make the `aio_agent_version` fact in Facter consistent with that in PE
by excluding the git sha from git describe.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -171,6 +171,10 @@ component "facter" do |pkg, settings, platform|
     end
   end
 
+  # In PE, aio_agent_version is distinguished from aio_agent_build by not including the git sha.
+  # Strip it off; this should have no impact on final releases, as git sha would not be included.
+  aio_agent_version = settings[:package_version].match(/^\d+\.\d+\.\d+(\.\d+){0,2}/).to_s
+
   unless platform.is_windows?
     special_flags += " -DFACTER_PATH=#{settings[:bindir]} \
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@jpartlow does this match your expectation?

configs/components/facter.rb
```diff
@@ -171,6 +171,10 @@
     end
   end
 
+  # In PE, aio_agent_version is distinguished from aio_agent_build by not including the git sha.
+  # Strip it off; this should have no impact on final releases, as git sha would not be included.
+  aio_agent_version = settings[:package_version].match(/^\d+\.\d+\.\d+(\.\d+){0,2}/).to_s
```
MikaelSmith: Chosen to match https://github.com/puppetlabs/puppetlabs-puppet_enterprise/blob/2016.2.x/lib/facter/aio_agent_version.rb#L7.

jpartlow:
:+1: Shall I merge?

configs/components/facter.rb
jpartlow: Oh, actually, the to_s of a MatchData, does that return just the matched portion?

jpartlow: Yes apparently so:

irb(main):002:0> settings = { :package_version => '1.5.0.123-gabcde45' }
=> {:package_version=>"1.5.0.123-gabcde45"}
irb(main):003:0> aio_agent_version = settings[:package_version].match(/^\d+\.\d+\.\d+(\.\d+){0,2}/).to_s
=> "1.5.0.123"

MikaelSmith: I tested by adding `puts "#{aio_agent_version}"; exit`, then run a build (which would exit before starting to build things).

PR #693, opened 2016-06-02
(PA-315) Add el-6-s390x platform

ScottGarman:
This platform is for RHEL running on IBM z-series

### DIFF
--- a/configs/components/openssl.rb
+++ b/configs/components/openssl.rb
@@ -11,8 +11,9 @@ component "openssl" do |pkg, settings, platform|
     pkg.build_requires "pl-gcc-#{platform.architecture}"
     pkg.build_requires 'runtime' if platform.is_huaweios?
     # needed for the makedepend command
+    pkg.build_requires 'imake' if platform.name =~ /^el/
+    pkg.build_requires 'xorg-x11-util-devel' if platform.name =~ /^sles/
     pkg.build_requires 'xutils-dev' if platform.is_huaweios?
-    pkg.build_requires 'xorg-x11-util-devel' if platform.architecture == "s390x"
   elsif platform.is_linux?
     pkg.build_requires 'pl-binutils'
     pkg.build_requires 'pl-gcc'
--- a/configs/components/ruby-selinux.rb
+++ b/configs/components/ruby-selinux.rb
@@ -11,7 +11,6 @@ component "ruby-selinux" do |pkg, settings, platform|
     pkg.url "https://raw.githubusercontent.com/wiki/SELinuxProject/selinux/files/releases/20100525/devel/libselinux-#{pkg.get_version}.tar.gz"
   end
 
-
   pkg.replaces 'pe-ruby-selinux'
 
   pkg.build_requires "ruby"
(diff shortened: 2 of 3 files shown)


ScottGarman:
Running this through a jenkins ad-hoc build to ensure no breakage for the non cross compiled EL platforms: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/144/

mckern:
> This platform is for RHEL running on **IBM z-series**.

configs/components/ruby-selinux.rb
```diff
@@ -20,10 +19,16 @@
...
+  if platform.is_cross_compiled_linux?
+    pkg.environment "RUBY" => settings[:host_ruby]
+    ruby = "#{settings[:host_ruby]} -r#{settings[:datadir]}/doc/rbconfig.rb"
+  end
+
   pkg.build do
-    ["export RUBYHDRDIR=$(shell #{settings[:bindir]}/ruby -rrbconfig -e 'puts RbConfig::CONFIG[\"rubyhdrdir\"]')",
-     "export VENDORARCHDIR=$(shell #{settings[:bindir]}/ruby -rrbconfig -e 'puts RbConfig::CONFIG[\"vendorarchdir\"]')",
-     "export ARCHDIR=$${RUBYHDRDIR}/$(shell #{settings[:bindir]}/ruby -rrbconfig -e 'puts RbConfig::CONFIG[\"arch\"]')",
+    ["export RUBYHDRDIR=$(shell #{ruby} -e 'puts RbConfig::CONFIG[\"rubyhdrdir\"]')",
```
mckern: This is more of a personal nitpick so it's nothing I'd block over. But when you've got multiple kinds of nested quotations going on like this, consider the humble `%()`:

```ruby
%(export RUBYHDRDIR=$(shell #{settings[:bindir]}/ruby -rrbconfig -e 'puts RbConfig::CONFIG["rubyhdrdir"]'))
```

A little less escaping, a slightly cleaner string (I think).

configs/components/openssl.rb
```diff
@@ -11,8 +11,9 @@
     pkg.build_requires "pl-gcc-#{platform.architecture}"
     pkg.build_requires 'runtime' if platform.is_huaweios?
     # needed for the makedepend command
+    pkg.build_requires 'imake' if platform.name =~ /^el/
```
mckern: Have we never needed `imake` before?

ScottGarman: That's a good question. I think something about customizing the target arch for cross-compiles invokes makedepend in a way that native builds do not. But I haven't dug into that. 

Either way, without this the build clearly fails with an error about needing the makedepend command, and succeeds when it has it. Distros are all over the map in how they package it.

geoffnichols: Not sure if this is useful, but [this thread](https://bugzilla.redhat.com/show_bug.cgi?id=890712) might be tangentially related to the `imake` question.

geoffnichols:
👍

PR #697, opened 2016-06-03
Revert "(maint) Update Leatherman to 0.7.2"

MikaelSmith:
Needs https://github.com/puppetlabs/leatherman/pull/186 to work on some platforms, and seems to have broken something on Windows as well. Needs more testing before bumping.


MikaelSmith:
Nevermind, changes in cpp-pcp-client depend on this. I'll look into just fixing it instead.

PR #699, opened 2016-06-03
(PA-375) Disable locales on SLES and EL-4

MikaelSmith:
Our images of these platforms come with an old version of gettext
installed. When the build attempts to use it, we run into errors because
expected gettext features are not available. Disable using gettext and
locales on these platforms.

This should have no other notable effects, as we use the default locale
on non-Windows platforms.


MikaelSmith:
Testing this out on EL-4.

MikaelSmith:
Leatherman built successfully with this on EL-4.

configs/components/cpp-pcp-client.rb
```diff
@@ -46,7 +46,10 @@
     pkg.build_requires "pl-cmake"
     pkg.build_requires "pl-boost"
 
-    if platform.is_cisco_wrlinux?
+    # SLES and EL-4 have old versions of gettext on them; until we build our own,
```
haus: Is it all sles with old gettext or just certain versions of sles?

haus: Also might be worth noting what version of gettext is required here.

Sles 10 has gettext 0.14.5-13.2
Sles 11 SP3 has gettext 0.17-61.40
Sles 12 has gettext 0.19.2-1.103

MikaelSmith: Looks like just SLES 10, so 0.17 appears to be new enough.

MikaelSmith: Updated the comments; I could also restrict to just SLES-10 and EL-4, or even just gettext versions prior to `0.17`. That'd take more effort, and I don't see a reason to get that specific yet.

haus: As long as it's written down, that seems sane to me.

johnduarte:
:+1:

PR #691, opened 2016-05-31
Revert "(maint) Enable tests, use check on all CMake projects"

McdonaldSeanp:
Reverts puppetlabs/puppet-agent#678

PR #700, opened 2016-06-03
(maint) Enable tests, use check on all CMake projects

MikaelSmith:
Enable tests on cpp-pcp-client and pxp-agent (on platforms where they're
expected to pass). Update facter and leatherman to use the
Component#check method for tests.

Re-implements https://github.com/puppetlabs/puppet-agent/pull/678 with a fix for the failure mentioned in https://github.com/puppetlabs/puppet-agent/pull/691.

### DIFF
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -69,6 +69,14 @@ component "cpp-pcp-client" do |pkg, settings, platform|
     ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
   end
 
+  # Can't run binaries when cross-compiling, and Solaris has an issue with
+  # accessing the certs while testing. AIX tests are always iffy.
+  if !platform.is_cross_compiled? && !platform.is_solaris? && !platform.is_aix?
+    pkg.check do
+      ["#{make} test ARGS=-V"]
+    end
+  end
+
   pkg.install do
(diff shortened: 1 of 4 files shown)


configs/components/cpp-pcp-client.rb
```diff
@@ -69,6 +69,14 @@
     ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1)"]
   end
 
+  # Can't run binaries when cross-compiling, and Solaris has an issue with
+  # accessing the certs while testing. AIX tests are always iffy.
+  if !platform.is_cross_compiled? && !platform.is_solaris? && !platform.is_aix?
```
stahnma: Why are the AIX tests iffy? It's not cross-compiled...

shrug:
Should probably hold until p-a master is green again, think fixes for the existing issues are working their way through now

configs/components/cpp-pcp-client.rb
MikaelSmith: I'm actually not sure, I just recall them being trouble on Facter. I can take a look at whether they should work for these.

MikaelSmith: Every unit test suite has a SIGABRT somewhere on AIX. It'll require more digging to figure out why.

MikaelSmith:
The cpp-pcp-client unit tests are now approaching 30 seconds as well. Maybe we should hold off on this until we have a proper caching mechanism.

stahnma:
Closing until we have a caching mechanism.

PR #704, opened 2016-06-07
(PA-349) Remove facter/curl/puppet-ca-bundle curcular deps

melissa:
Prior to this commit, facter depended on curl, which depended on
puppet-ca-bundle, which depended on facter. This is bad, since we don't
want circular dependencies in our build systems.

This commit removed that circular dependency by altering
puppet-ca-bundle's requirement on facter. The only reason facter was
listed as a build dependency was to get a list of platforms that had
java installed on them. Though the Makefile for puppet-ca-bundle does
not require java explicitly, this was a means to determine which
platforms would be made available as master platforms. These required
the additional keystore to be available to the puppetserver package. The
original logic behind this decision was to avoid duplication of logic in
determining which platforms to make the keystore available on. Rather
than perpetuating the circular logic in favor of reduced maintenance,
this commit simply generates the keystore for all agents.

### DIFF
--- a/configs/components/puppet-ca-bundle.rb
+++ b/configs/components/puppet-ca-bundle.rb
@@ -1,9 +1,7 @@
 component "puppet-ca-bundle" do |pkg, settings, platform|
   pkg.load_from_json("configs/components/puppet-ca-bundle.json")
+
   pkg.build_requires "openssl"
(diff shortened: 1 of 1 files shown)


melissa:
Alternate solution: https://github.com/puppetlabs/puppet-agent/pull/702

PR #706, opened 2016-06-07
(maint) move project.wxs from vanagon to puppet-agent

McdonaldSeanp:
After realizing that project.wxs does not actually constitute a default that
belongs in vanagon, this file should be moved under puppet-agent. Vanagon is
already built to handle this change so no vanagon changes will be required.


McdonaldSeanp:
This was tested and will compile cleanly

PR #708, opened 2016-06-08
(PA-392) Ensure linking against PL libs

stahnma:
Prior to this commit, there wasn't any verification that we were linking
against our  libstdc++ vs the system-wide library. We'd like to make
sure that we've either statically linked libstdc++ and libgcc_s into our
builds or that we're using the runtime-provided components that would
end up in /opt/puppetlabs/puppet/lib.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -219,6 +219,14 @@ component "facter" do |pkg, settings, platform|
     ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
   end
 
+  pkg.check do
+    [
+      # Ensure we're linking against our libstdc++ and libgcc_s or statically linking
+      "ldd lib/libfactero.so | grep libstdc++ && ldd lib/libfacter.so | grep libstdc++ | grep puppet || true",
+      "ldd lib/libfacter.so | grep libgcc_s && ldd lib/libfacter.so | grep puppet || true"
+    ]
+  end
+
   pkg.install_file ".gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}.gemspec"
   if platform.is_windows?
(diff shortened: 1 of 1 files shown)


configs/components/facter.rb
```diff
@@ -219,6 +219,14 @@
     ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
   end
 
+  pkg.check do
+    [
+      # Ensure we're linking against our libstdc++ and libgcc_s or statically linking
+      "ldd lib/libfactero.so | grep libstdc++ && ldd lib/libfacter.so | grep libstdc++ | grep puppet || true",
+      "ldd lib/libfacter.so | grep libgcc_s && ldd lib/libfacter.so | grep puppet || true"
```
mckern: `ldd` sometimes chokes on cross-compiled libraries. What happens here?

```
vagrant@vagrant-deb8-builder:~/sources/ruby-shadow-2.5.0$ ldd shadow.so
	not a dynamic executable
vagrant@vagrant-deb8-builder:~/sources/ruby-shadow-2.5.0$ file shadow.so
shadow.so: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, BuildID[sha1]=b7399b36e9256aff581ec5e36ee5ecc43a4bc58b, not stripped
```

stahnma: Good question. I thought I tested that. Let me re-verify.

stahnma: Tested against: sles-12-s390x, solaris-10-i386, solaris-11-sparc, el-5-i386, el-7-x86_64 and ubuntu-16.04-amd64 without issues.

ScottGarman:
:+1:

PR #707, opened 2016-06-08
Merge stable to master

shrug:
Merge stable to master to pick up some fixes

PR #709, opened 2016-06-08
(maint) Update to correct Leatherman version after merge

MikaelSmith:
Merge from stable didn't bring over a related Leatherman ref bump to
support new configuration. Update it now.


MikaelSmith:
Missed this in https://github.com/puppetlabs/puppet-agent/pull/707; builds will probably fail without it.

PR #710, opened 2016-06-08
(RE-6958) Update build-data to include sles-11-s390x

McdonaldSeanp:
Update the pe_platforms list to include sles-11-s390x to include it in nightlies


ScottGarman:
:+1:

PR #712, opened 2016-06-08
(RE-7178) Update build-data to include sles-12-s390x

McdonaldSeanp:
Update the pe_platforms list to include sles-12-s390x to include it in
nightlies


mckern:
👍

PR #714, opened 2016-06-09
Revert "(PA-392) Ensure linking against PL libs"

stahnma:
Reverts puppetlabs/puppet-agent#708

This test was not catching most cases where problems exist. More work required.


underscorgan:
👍

PR #715, opened 2016-06-13
(maint) Skip runtime.sh on OSX

shrug:
The recent addition of 'runtime.sh' fails with an
error on OSX. As it's not needed or useful on OSX,
this skips the invocation on those platforms.


underscorgan:
👍

PR #717, opened 2016-06-13
(PA-379) Prepare PA 1.5.2 for release

melissa:


configs/components/windows_puppet.json
```diff
@@ -1,4 +1,4 @@
 {
   "url": "git://github.delivery.puppetlabs.net/puppetlabs-puppet_for_the_win.git",
-  "ref": "origin/master"
+  "ref": "refs/tags/4.5.1"
```
shrug: Should we tag this at 4.5.2 also?

melissa: cc/ @Iristyle there were no changes to the master branch since that 4.5.1 tag was put in. I'm happy to add a 4.5.2 tag to the same sha if that helps

shrug: I think 4.5.1 was tagged at the same sha as 4.5.0 also

melissa: kk, I'll make that happen

shrug:
👍

PR #720, opened 2016-06-15
(PA-318) Add el-7-s390x platform

ScottGarman:
This is for RHEL 7 running on IBM z-series.

### DIFF
--- a/configs/components/libxslt.rb
+++ b/configs/components/libxslt.rb
@@ -15,6 +15,8 @@ component "libxslt" do |pkg, settings, platform|
 
     # libxslt is picky about manually specifying the build host
     build = "--build x86_64-linux-gnu"
+    # don't depend on libgcrypto
+    disable_crypto = "--without-crypto"
   elsif platform.is_solaris?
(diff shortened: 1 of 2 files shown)


ScottGarman:
Running an ad-hoc jenkins build for additional testing: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/157/

configs/components/libxslt.rb
```diff
@@ -15,6 +15,8 @@
 
     # libxslt is picky about manually specifying the build host
     build = "--build x86_64-linux-gnu"
+    # don't depend on libgcrypto
+    disable_crypto = "--without-crypto"
```
geoffnichols: Why do we need to add this flag?

ScottGarman: Without this flag, the build fails because the configure script finds the libgcrypto header file from the x86_64 environment and tries to build and link to it.

geoffnichols:
👍

PR #722, opened 2016-06-16
(RE-7398) Update libxslt to 1.1.29

sarameisburger:

PR #726, opened 2016-06-17
Revert "(RE-7398) Update libxslt to 1.1.29"

McdonaldSeanp:
Reverts puppetlabs/puppet-agent#722, This update was causing build failures with solaris and the issue is non-trivial to fix, so we are reverting for now to keep the pipelines clear


ScottGarman:
As an alternate to this and then having to revert the nokogiri update, could we instead just disable the libxslt component for all non-huawei builds? Or even just for solaris?

McdonaldSeanp:
We are going to only revert this for solaris

PR #728, opened 2016-06-17
(RE-7540) revert libxslt bump back to 1.1.28 on solaris

McdonaldSeanp:
Due to failing builds on solaris OSes we are reverting the libxslt version bump
on only solaris platforms. This should be a temporary fix.


shrug:
👍

PR #729, opened 2016-06-20
(RE-7407) move facter rb

McdonaldSeanp:
due to the legacy build process, puppet-agent and the test suite expect
facter.rb to exist under facter/lib, not the vendor_ruby directory. This commit
uses the ruby_lib_install option of facter's cmake makefile to move that file.


McdonaldSeanp:
This PR depends on https://github.com/puppetlabs/facter/pull/1355

mckern:
👍

PR #736, opened 2016-06-27
(PA-390) On AIX, restart services on upgrade

geoffnichols:
Prior to this, services on AIX are not restarted on upgrade, so
the old puppet and mcollective services may be left running.

This updates puppet-agent's puppet and mcollective component
configurations to save and apply the pre-upgrade service state
for puppet and mcollective (as is done for other RPM platforms 
not using systemd)

Also, on AIX, the system default `install` utility does not create
directories. This commit updates the puppet and mcollective preinstall
actions to use a more portable way of creating the `rpm_statedir`
directories.


geoffnichols:
Depends on https://github.com/puppetlabs/vanagon/pull/387.

geoffnichols:
Closing to retarget against master.

PR #738, opened 2016-06-28
(PA-390) On AIX, restart services on upgrade

geoffnichols:
Prior to this, services on AIX are not restarted on upgrade, so
the old puppet and mcollective services may be left running.

This updates puppet-agent's puppet and mcollective component
configurations to save and apply the pre-upgrade service state
for puppet and mcollective (as is done for other RPM platforms 
not using systemd)

Also, on AIX, the system default `install` utility does not create
directories. This commit updates the puppet and mcollective preinstall
actions to use a more portable way of creating the `rpm_statedir`
directories.

### DIFF
--- a/configs/components/marionette-collective.rb
+++ b/configs/components/marionette-collective.rb
@@ -55,7 +55,7 @@ component "marionette-collective" do |pkg, settings, platform|
     service_statefile = "#{rpm_statedir}/service.pp"
     pkg.add_preinstall_action ["upgrade"],
       [<<-HERE.undent
-        install --owner root --mode 0700 --directory #{rpm_statedir} || :
+        mkdir -p  #{rpm_statedir} && chown root #{rpm_statedir} && chmod 0700 #{rpm_statedir} || :
         if [ -x #{puppet_bin} ] ; then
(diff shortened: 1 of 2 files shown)


geoffnichols:
Replaces https://github.com/puppetlabs/puppet-agent/pull/736.

geoffnichols:
Fix is incomplete without https://github.com/puppetlabs/vanagon/pull/390.

configs/components/marionette-collective.rb
```diff
@@ -73,6 +49,29 @@
     fail "need to know where to put service files"
   end
 
+  if (platform.servicetype == "sysv" && platform.is_rpm?) || platform.is_aix?
```
shrug: only sysv?

geoffnichols: If I'm following the flow correctly, I think that matches what we're doing now? https://github.com/puppetlabs/puppet-agent/pull/738/files#diff-9ae79323449fab62c617785704463a77L29.

shrug: you're right, I can't read

shrug:
👍

PR #742, opened 2016-06-29
(PA-392) Ensure linking against PL libs

geoffnichols:
This commit reinstates https://github.com/puppetlabs/puppet-agent/pull/708. From the original PR:

```
Prior to this commit, there wasn't any verification that we were linking
against our libstdc++ vs the system-wide library. We'd like to make
sure that we've either statically linked libstdc++ and libgcc_s into our
builds or that we're using the runtime-provided components that would
end up in /opt/puppetlabs/puppet/lib.
```

In addition, this changeset adds 'runtime' as a build dependency for all platforms and
updates the libfacter linking check to catch more cases where problems
exist.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -17,6 +17,7 @@ component "facter" do |pkg, settings, platform|
   pkg.build_requires "ruby"
   pkg.build_requires 'openssl'
   pkg.build_requires 'leatherman'
+  pkg.build_requires 'runtime'
 
   if platform.is_linux? && !platform.is_huaweios?
     # Running facter (as part of testing) expects virt-what is available
@@ -49,19 +50,16 @@ component "facter" do |pkg, settings, platform|
     pkg.build_requires "pl-boost-#{platform.architecture}"
     pkg.build_requires "pl-yaml-cpp-#{platform.architecture}"
     pkg.build_requires "pl-cmake"
-    pkg.build_requires "runtime" if platform.is_cross_compiled_linux?
   elsif platform.is_aix?
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-gcc-5.2.0-1.aix#{platform.os_version}.ppc.rpm"
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-cmake-3.2.3-2.aix#{platform.os_version}.ppc.rpm"
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-boost-1.58.0-1.aix#{platform.os_version}.ppc.rpm"
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-yaml-cpp-0.5.1-1.aix#{platform.os_version}.ppc.rpm"
-    pkg.build_requires "runtime"
   elsif platform.is_windows?
     pkg.build_requires "cmake"
     pkg.build_requires "pl-toolchain-#{platform.architecture}"
     pkg.build_requires "pl-boost-#{platform.architecture}"
     pkg.build_requires "pl-yaml-cpp-#{platform.architecture}"
-    pkg.build_requires "runtime"
   else
     pkg.build_requires "pl-gcc"
     pkg.build_requires "pl-cmake"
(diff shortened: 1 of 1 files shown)


geoffnichols:
Doing a test build with this now.

underscorgan:
👍

geoffnichols:
Update commit to correct copypasta error.

configs/components/facter.rb
```diff
@@ -226,6 +224,23 @@
...
+    ldd = "ldd"
+  end
+
+  unless platform.is_windows? || platform.is_cross_compiled_linux? || platform.architecture == 'sparc'
+    pkg.check do
+      [
+        # Ensure we're linking against our libstdc++ and libgcc_s or statically linking
+        "#{ldd} lib/libfacter.so",
+        "[ ! $$(#{ldd} lib/libfacter.so | grep -c libstdc++) -ge 1 ] || #{ldd} lib/libfacter.so | grep libstdc++ | grep puppet || #{ldd} lib/libfacter.so | grep libstdc++ | grep pl-build-tools",
+        "[ ! $$(#{ldd} lib/libfacter.so | grep -c libgcc_s) -ge 1 ] || #{ldd} lib/libfacter.so | grep libgcc_s | grep puppet || #{ldd} lib/libfacter.so | grep libgcc_s | grep pl-build-tools"
```
mckern: These initial tests (`[]`) are kind of janky. Why are we inverting the test conditional in there? Wouldn't this be the same:

```bash
mckern@weth puppet $ [ ! $(ldd lib/libfacter.so | grep -c libstdc++) -ge 1 ] && echo "true" || echo "false"
true
mckern@weth puppet $ [ $(ldd lib/libfacter.so | grep -c libstdc++) -eq 0 ] && echo "true" || echo "false"
true
mckern@weth puppet $
```

mckern: I think we could simplify the command pipelines here with a little bit of globbing

```bash
mckern@weth puppet $ ( [ $(ldd lib/libfacter.so | grep -c libstdc++) -ge 0 ] ||
>   ldd lib/libfacter.so | grep "puppet*libstdc++" ||
>   ldd lib/libfacter.so | grep "pl-build-tools*libstdc++" ) &&
> echo "works for giant man"
works for giant man
mckern@weth puppet $
mckern@weth puppet $ ( [ $(ldd lib/libfacter.so | grep -c libgcc_s) -ge 0 ] ||
>   ldd lib/libfacter.so | grep "puppet*libgcc_s" ||
>   ldd lib/libfacter.so | grep "pl-build-tools*libgcc_s" ) &&
> echo "still works for giant man"
still works for giant man
mckern@weth puppet $
```

geoffnichols:
Updated.

configs/components/facter.rb
```diff
@@ -226,6 +224,23 @@
...
+    ldd = "ldd"
+  end
+
+  unless platform.is_windows? || platform.is_cross_compiled_linux? || platform.architecture == 'sparc'
+    pkg.check do
+      [
+        # Ensure we're linking against our libstdc++ and libgcc_s or statically linking
+        "#{ldd} lib/libfacter.so",
+        "[ $$(#{ldd} lib/libfacter.so | grep -c libstdc++)  -lt 1 ] || #{ldd} lib/libfacter.so | grep libstdc++ | grep -v ' /lib'",
+        "[ $$(#{ldd} lib/libfacter.so | grep -c libgcc_s) -lt 1 ] || #{ldd} lib/libfacter.so | grep libgcc_s | grep -v ' /lib'"
```
mckern: Isn't `0` the only integer that's less than `1`? Why not `-eq 0`?

mckern: Is the path to `libstdc++` static? Could we possibly grep for the entire path?

mckern:
Generally 👍 on this. I could nitpick it into oblivion but we can always iterate.

configs/components/facter.rb
geoffnichols: Testing at build time, it looks like the path to `libstdc++` on most platforms (where we dynamically link) is `/opt/puppetlabs/puppet/lib`. However, there are some outliers (EL4, Ubuntu 15.10), where the path at build time is either `/opt/pl-build-tools/lib` or `/opt/pl-build-tools/lib64`.

geoffnichols:
Running another test build at https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/188/.

configs/components/facter.rb
```diff
@@ -226,6 +224,23 @@
...
+    ldd = "ldd"
+  end
+
+  unless platform.is_windows? || platform.is_cross_compiled_linux? || platform.architecture == 'sparc'
+    pkg.check do
+      [
+        # Check that we're not linking against system libstdc++ and libgcc_s
+        "#{ldd} lib/libfacter.so",
+        "[ $$(#{ldd} lib/libfacter.so | grep -c libstdc++) -eq 0 ] || #{ldd} lib/libfacter.so | grep libstdc++ | grep -v ' /lib'",
+        "[ $$(#{ldd} lib/libfacter.so | grep -c libgcc_s) -eq 0 ] || #{ldd} lib/libfacter.so | grep libgcc_s | grep -v ' /lib'"
```
mckern: I'm not blocking on it, but whitespace specific grep's seem like they'll be fragile. Let's run with this for a while and see how it goes.

geoffnichols:
Test [build](https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/188/) completed successfully.

McdonaldSeanp:
👍

PR #744, opened 2016-06-29
(RE-7338) Add Fedora 23 and 24 to 'foss_platforms'

geoffnichols:
This commit adds Fedora 23 and 24 to the 'foss_platforms' list, so we
can ship them to nightlies.


shermdog:
:+1:

PR #746, opened 2016-06-30
(maint) Use upstream URLs for libxml2, pkg-config, and mini_portile2.

geoffnichols:
We recently started using upstream component sources in puppet-agent
component source configs.

This commit updates the component configs for `libxml2`,
`mini_portile2`, and `pkg-config` to use the upstream source URLs.


shrug:
👍

PR #748, opened 2016-06-30
(RE-7570) Add rsync and fakeroot to Ubuntu 12.04 and 14.04

shermdog:
We require fakeroot to build Puppet Agent.  The templates for Ubuntu 12.04 and 14.04 were previously handcrafted and included this package.  By default it is not installed and needs to be explicitly called out to support moving to packer based templates.


mckern:
Good idea. Can you add `rsync` too, for the SAME reason?

geoffnichols:
👍

shermdog:
@mckern good call - added.

mckern:
👍 AF

PR #749, opened 2016-06-30
(maint) Use upstream URLs for semantic_puppet and libxslt

geoffnichols:
We recently started using upstream component sources in puppet-agent
component source configs.

This commit updates the component configs for `semantic_puppet` and
`libxslt` to use the upstream source URLs.


shermdog:
:+1:  and validated the URL's

PR #753, opened 2016-07-05
(RE-7587) (RE-7593) Windows fixes to mco and pxp-agent

McdonaldSeanp:
~~This PR requires https://github.com/puppetlabs/pxp-agent/pull/448~~

### DIFF
--- a/configs/components/marionette-collective.rb
+++ b/configs/components/marionette-collective.rb
@@ -81,14 +81,15 @@ component "marionette-collective" do |pkg, settings, platform|
   end
 
   flags = " --bindir=#{settings[:bindir]} \
+            --sbindir=#{settings[:bindir]} \
             --sitelibdir=#{settings[:ruby_vendordir]} \
             --ruby=#{File.join(settings[:bindir], 'ruby')} "
 
   if platform.is_windows?
-    pkg.install_file "ext/windows/daemon.bat", "#{settings[:bindir]}/mco_daemon.bat"
-    pkg.add_source("file://resources/files/windows/mco.bat", sum: "2d29af9c926dcf8b50ae9ac1bdb18e1f")
+    pkg.add_source("file://resources/files/windows/mco.bat")
     pkg.install_file "../mco.bat", "#{settings[:link_bindir]}/mco.bat"
     flags = " --bindir=#{settings[:mco_bindir]} \
+              --sbindir=#{settings[:mco_bindir]} \
               --sitelibdir=#{settings[:mco_libdir]} \
               --no-service-files \
               --ruby=#{File.join(settings[:ruby_bindir], 'ruby')} "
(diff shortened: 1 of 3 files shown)


McdonaldSeanp:
pxp-agent PR was merged, this is ready

McdonaldSeanp:
matrix build at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/539/

configs/components/marionette-collective.rb
```diff
@@ -81,14 +81,15 @@
...
 
   flags = " --bindir=#{settings[:bindir]} \
+            --sbindir=#{settings[:bindir]} \
             --sitelibdir=#{settings[:ruby_vendordir]} \
             --ruby=#{File.join(settings[:bindir], 'ruby')} "
 
   if platform.is_windows?
-    pkg.install_file "ext/windows/daemon.bat", "#{settings[:bindir]}/mco_daemon.bat"
-    pkg.add_source("file://resources/files/windows/mco.bat", sum: "2d29af9c926dcf8b50ae9ac1bdb18e1f")
+    pkg.add_source("file://resources/files/windows/mco.bat")
```
melissa: `add_source` no longer needs a checksum?

McdonaldSeanp: Yep, it no longer requires checksums for local files

configs/components/ruby.rb
```diff
@@ -189,6 +189,21 @@
...
+    # (puppet.bat ,mco.bat etcc) the location of openssl away from where ruby is
+    # installed will cause a failure. Specifically this is meant to help services like
+    # mco that require openssl but don't have access to environment.bat. Refer to
+    # https://tickets.puppetlabs.com/browse/RE-7593 for details on why this causes
+    # failures and why these copies fix that.
+    #                   -Sean P. McDonald 07/01/2016
+    pkg.install do
+      [
+        "cp #{settings[:prefix]}/bin/ssleay32.dll #{settings[:ruby_bindir]}",
+        "cp #{settings[:prefix]}/bin/libeay32.dll #{settings[:ruby_bindir]}",
```
melissa: Hmm, I'm wondering if this change makes more sense to do in the openssl.rb file? Or is it that we need to make these files available specifically in the ruby bindir? I'm not sure what makes more sense...

melissa: So, I guess I'm wondering is: should we default to install openssl where ruby is installed?

configs/components/marionette-collective.rb
melissa: solid!

melissa:
👍  aside from question on the default openssl install location. But I think this is a very reasonable change

configs/components/ruby.rb
McdonaldSeanp: I believe that will break the puppet service in the same way. From what I know, I believe they need to exists in both places ... I will check it out.

melissa: If that breaks things, then don't worry about it

PR #755, opened 2016-07-09
(PA-320) Install semantic_puppet to vendor_ruby

kylog:


kylog:
This moves semantic_puppet to vendor_ruby instead of doing a gem install, as per a proposal in email a few weeks ago. The idea of doing this is so that it is available to puppetserver at runtime, since puppetserver intentionally puts vendor_ruby in it's jruby loadpath but *not* puppet-agent's own gems (to avoid pulling in native gems that wouldn't work with jruby).

I verified this seems to work on one platform (el7). I *think* this logic will work for other platforms, e.g. windows (once we turn on vanagon-based msi builds) and the cross-platform targets, but I'm out of my depth on both counts so I'd appreciate review with that in mind (of course, along with general suitability of this approach).

configs/components/rubygem-semantic_puppet.rb
```diff
@@ -1,25 +1,8 @@
...
-  # set RUBYLIB to include puppet and hiera, so that their gemspecs can resolve
-  # hiera/version and puppet/version requires. Without this the gem install
-  # will fail by blowing out the stack.
-  pkg.environment "RUBYLIB" => "#{settings[:ruby_vendordir]}:$$RUBYLIB"
+  pkg.url "git@github.com:puppetlabs/semantic_puppet.git"
+  pkg.ref "v0.1.2"
 
   pkg.install do
-    ["#{settings[:gem_install]} semantic_puppet-#{pkg.get_version}.gem"]
+    ["cp -pr lib/* #{settings[:ruby_vendordir]}"]
```
stahnma: Do we not want this gem to show up as a gem?

kylog: That's a great question. This makes it available to puppetserver. But if we want *both* those things (puppetserver availability and visibility to rubygems) then we need a new mechanism somehow.

Meanwhile, it seems like more of the discussion is going on in RE-7610, so maybe let's follow along there and leave this PR as a POC for what one approach looks like.

stahnma:
@kylog should this be closed due to other attempts at fixing the issues this raised?

kylog:
Sure thing. Closing.

PR #756, opened 2016-07-14
(Maint) Fix up pkg-config dependency for cross compiles

stahnma:
Previously, cross compiled Debians required pl-pkg-config to build
libxml2 but was only referenced in augeas. This commit places the build
requirement in the correct spot (libxml2) and ensures that anything
cross compiled as a Debian system will require pl-pkg-config.

If you just use system pkg-config, the cross compile will fail on some
platforms as the flags returned are incorrect.


shrug:
👍

PR #757, opened 2016-07-15
(maint) Use git protocol instead of ssh

joshcooper:
Previously, we were cloning cfpropertylist using ssh, which requires
ssh authentication and an entry in ~/.ssh/known_hosts. Since the
cfpropertylist repo is public, switch to the git protocol instead.


joshcooper:
ping @shrug

shrug:
👍

PR #758, opened 2016-07-18
(RE-7419) Ship sles packages to foss repos

melissa:


shrug:
lgtm 👍  but ping @stahnma to make sure that's the right set of stuff

PR #759, opened 2016-07-19
(maint) Update hiera config to not use --no-batch-files

McdonaldSeanp:
This option was added in master hiera, but will not be available until after
the 1.5.3 release


shrug:
👍

PR #760, opened 2016-07-19
(PA-427) Add cpp-hocon library to the build

branan:


configs/components/cpp-hocon.json
```diff
@@ -0,0 +1 @@
+{"url": "git://github.com/puppetlabs/cpp-hocon.git", "ref": "master"}
```
Magisus: Do we want this to point to the tagged version now?

branan: maybe

PR #762, opened 2016-07-22
Revert HOCON

branan:
We've encountered some build issues with HOCON on some platforms. To get things green we're going to revert for now, and re-apply when we have it (more) sorted


Magisus:
Sounds good, I'll try to get these resolved asap. Merging.

PR #764, opened 2016-07-26
Ruby 2.3 bump

McdonaldSeanp:


stahnma:
Per @shrug let's get a good build of PA through on master before merging this. 💃

geoffnichols:
👍

PR #767, opened 2016-07-28
(RE-7685) (RE-7684) Update ruby to revert el-6-s390, windows

McdonaldSeanp:
Both the el-6-s390x and windows platforms are not ready for the 2.3 bump, this
commit should revert some of the ruby work for those platforms to bring them
back in line with 2.1.9.


stahnma:
👍

melissa:
So, the failures this PR addresses is that the patches fail to apply because they're meant for a newer version of ruby? What's the problem this change is fixing?

McdonaldSeanp:
@melissa yes, this will revert to building 2.1.9 for windows packages to remove CI failures while we message ruby on windows to get 2.3.1 to play nice with puppet

PR #773, opened 2016-07-29
(PUP-6459) Test updated win32-service 0.8.8

Iristyle:
- Though future packaging should change to Vanagon based, current
   pipelines are still testing with packages produced with the older
   process that rely on the puppet-win32-ruby repository.

   To verify that the updated win32-service 0.8.8 gem passes through
   acceptance properly, update to the latest SHAs at the head of the
   2.1.x-x{86|64} branches of puppet-win32-ruby that include
   win32-service 0.8.8

   Once validated, these SHAs can be turned into tags, and this
   component configuration can be updated.


MosesMendoza:
👍

PR #774, opened 2016-07-29
(PUP-6459) Remove win32-service 0.8.6 patch

Iristyle:
- 1476ac8117fe27edb4416594b6fca742a510a02d introduced a patch to the
   win32-service gem to re-enable Windows 2003 support (as part of
   PUP-4390).  This was since the proposed patch to the upstream gem was
   outright rejected: https://github.com/djberg96/win32-service/pull/26

   At this time, not only does this patch not cleanly apply to the
   recently updated win32-service 0.8.8, but since 2003 is no longer a
   supported platform, it can be removed outright.


McdonaldSeanp:
This will require removal of this line: https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/rubygem-win32-service.rb#L22

Iristyle:
Thanks @McdonaldSeanp!

I didn't realize patching was an opt-in like that.  I thought it was an automatic convention based system that automatically found patches based on directory structure.

McdonaldSeanp:
👍

melissa:
Aww, I remember that patch

PR #776, opened 2016-08-01
(maint) Update Leatherman to 0.8.1

MikaelSmith:
Changes needed PCP-494.


branan:
woah, I didn't realize we were running with a SHA O_O When did that happen?

PR #780, opened 2016-08-01
(maint) Update ruby congifuration for windows

McdonaldSeanp:
Windows builds of the vendored ruby will incorrectly create batch files using
old legacy rbinstall code. This commit patches out the old legacy code and
keeps creation of ruby wrappers in line with what rubygems will do.


shrug:
👍

PR #782, opened 2016-08-02
(RE-7667) OSX: Update 'include/memory' header for libc++ bug

geoffnichols:
Prior to this commit, cpp-hocon builds fail on OS X due to a known issue
(https://llvm.org/bugs/show_bug.cgi?id=18843) in an earlier version of
the libc++ headers shipped with LLVM (in the Xcode Command Line Tools).

This issue is fixed in Xcode Command Line Tools 7.3.

This changeset updates the OS X 10.11 platform config to use Xcode Command
Line Tools 7.3.1, the current version for OS X 10.11.

Unfortunately, installation of Xcode CLT 7.3 is blocked on OS X 10.9 and 10.10. 
As a workaround, this changeset updates puppet-agent's OS X 10.9 and 10.10 
platform configurations to patch libc++'s 'include/memory' header as a
provisioning step before the build.


geoffnichols:
Ping @Magisus @branan

Magisus:
Thanks for getting this figured out! 👍

PR #790, opened 2016-08-03
(maint) Update ruby to use dynamic linking

McdonaldSeanp:
Due to issues trying to statically link libgcc on windows x86, we are going to
dynamically link that lib and copy it over to the ruby bindir.


ScottGarman:
Need to fix the commit msg.

ScottGarman:
:+1:

PR #794, opened 2016-08-03
Bump Puppet and Facter to planned release SHAs

branan:


branan:
There may be another Puppet bump from the windows team, but we're currently pretty sure this is what we want to ship

kylog:
👍

PR #795, opened 2016-08-03
(maint) Update readme to not reference puppet_for_the_win

McdonaldSeanp:


mckern:
Excise all the things.
👍

mckern:
Oh, do you want to alter/remove the `bin/build-windows.rb` stub?

McdonaldSeanp:
No, that needs to stay as long as we support 3.8

PR #800, opened 2016-08-04
bumping hiera to 1a407b121dc839494577f953aa6d5ca34ae5b5ab

kylog:


kylog:
This is the hiera sha from hiera#master including the fix for HI-528.

PR #802, opened 2016-08-04
(maint) Reset puppet to a pre-corrective-change-reporting SHA

branan:
This resets the Puppet SHA to a version before corrective change
landed. This will let us get a clean build to the PE integration team
while we sort out what we're going to do in corrective change land.

This can't be our final shipping SHA since it's missing some other
fixes, but it should be good enough to unblock the integration team.

### DIFF
--- a/configs/components/puppet.json
+++ b/configs/components/puppet.json
@@ -1,1 +1,1 @@
-{"url": "git://github.com/puppetlabs/puppet.git", "ref": "2ebbe8f2998ddb27f6bee9ab45513889cda485b0"}
+{"url": "git://github.com/puppetlabs/puppet.git", "ref": "6933ab99bd40122b765b8603bd601786286a7010"}


whopper:
I should note that the integration team isn't actually blocked right now. The bad promotions were reverted and we're just sitting on the last good one from whenever that was.

underscorgan:
👍

branan:
@whopper shush, I'm trying to help you :P

kylog:
👍

whopper:
@branan :D

kylog:
@whopper we've never run some pretty big changes in puppet through PE builds. This enables earlier exposure to that testing.

whopper:
@kylog oh yeah that's why I mentioned this to @branan in the first place - not because we're blocked, but because PA is getting stale over there and could probably use some time in integration :)

PR #803, opened 2016-08-05
bumping hiera to 98a0d54a367146e7dca4e170b548a675dab90570

kylog:


kylog:
This is the bump to the SHA for what should be hiera 3.2.1; note that comes off hiera#stable since this is a hiera Z.

underscorgan:
👍

PR #809, opened 2016-08-09
(PA-466) Correct env_windows_installdir

MosesMendoza:
In commit 24f8e0d we introduced a different path for
FACTER_env_windows_installdir - it used to "C:\\Program Files\\Puppet
Labs\\Puppet" but now goes to "C:\\Program Files\\Puppet Labs\\Puppet\\facter".
This commit fixes this. The impact is that in PE, the aio_agent_build fact uses
this fact to create the path to the VERSION file - without it, no
aio_agent_build fact is reported, causing CI pipelines to fail.

### DIFF
--- a/resources/files/windows/environment.bat
+++ b/resources/files/windows/environment.bat
@@ -5,7 +5,7 @@ REM Avoid the nasty \..\ littering the paths.
 SET PL_BASEDIR=%PL_BASEDIR:\bin\..=%
 
 REM Set a fact so we can easily source the environment.bat file in the future.
-SET FACTER_env_windows_installdir=%PL_BASEDIR%\facter
+SET FACTER_env_windows_installdir=%PL_BASEDIR%\puppet
 
 SET PUPPET_DIR=%PL_BASEDIR%\puppet
 REM Facter will load FACTER_ env vars as facts, so don't use FACTER_DIR


MosesMendoza:
Tested on Windows Server 2008 R2

McdonaldSeanp:
👍

Iristyle:
FYI @shrug we're still validating that this is the correct location.  Previously we had 3 places where this was used:

* `%env_windows_installdir%\VERSION`
* `%env_windows_installdir%\bin\environment.bat`
* `%env_windows_installdir%\bin\puppet.bat`

In the previous installer, these mapped to:

* `C:\Program Files\Puppet Labs\Puppet\VERSION`
* `C:\Program Files\Puppet Labs\Puppet\bin\environment.bat`
* `C:\Program Files\Puppet Labs\Puppet\bin\puppet.bat`


So it looks like this PR might have an extraneous `puppet` in there.  Verifying again with @McdonaldSeanp and @MosesMendoza

Iristyle:
So it turns out the new MSI wasn't maintaining the same relative pathing structure.  The @MosesMendoza fix was correct for the (accidentally incorrect) location of `VERSION`.  Unfortunately it won't work for the MCO requirements.

So in addition to amending this with a new PR to remove the `puppet`, @McdonaldSeanp will also put up a new PR to move / copy `VERSION` to the previous location.

Iristyle:
Superceded by #810

PR #813, opened 2016-08-10
(PA-466) correct env_windows_installdir

McdonaldSeanp:
A previous commit incorrectly set the env_windows_installdir fact to be
suffixed with 'puppet'. This was tested with the aio_agent_build fact available
in PE, but it turns out that is actually incorrect and the combination of the
two returned a false positive. This commit updates the path to the correct
(pre-vanagon) path.

This is only a cherry pick


Iristyle:
👍

PR #814, opened 2016-08-10
(maint) Update cp to use full path

McdonaldSeanp:
For some reason, the environment is preventing cygwin from seeing the cp
command while building puppet, update the copy command for the windows build to
use the full path


MikaelSmith:
The "for some reason" is because CMake won't work correctly if it sees `sh.exe` in the PATH.

McdonaldSeanp:
Ah, ty

PR #820, opened 2016-08-17
(maint) Add commit checks

underscorgan:
We want to make sure we aren't using private projects in our commit
messages. This will allow maint, doc, docs, packaging, pa, and cpr

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -36,3 +36,41 @@ if File.exist?(build_defs_file)
     end
   end
 end
+
+desc "verify that commit messages match CONTRIBUTING.md requirements"
+task(:commits) do
+  commits = ENV['TRAVIS_COMMIT_RANGE']
+  if commits.nil?
+    puts "TRAVIS_COMMIT_RANGE is undefined, I don't know what to check."
+    exit
+  end
+
+  %x{git log --no-merges --pretty=%s #{commits}}.each_line do |commit_summary|
+    error_message=<<-HEREDOC
+\n\n\n\tThis commit summary didn't match CONTRIBUTING.md guidelines:\n \
+\n\t\t#{commit_summary}\n \
+\tThe commit summary (i.e. the first line of the commit message) should start with one of:\n  \
+\t\t(docs)\n \
+\t\t(maint)\n \
+\t\t(packaging)\n \
+\t\t(<ANY PUBLIC JIRA TICKET>)\n \
+\n\tThis test for the commit summary is case-insensitive.\n\n\n
+    HEREDOC
+
+    if /^\((maint|doc|docs|packaging)\)|revert|bumping/i.match(commit_summary).nil?
+      ticket = commit_summary.match(/^\(([[:alpha:]]+-[[:digit:]]+)\).*/)
+      if ticket.nil?
+        raise error_message
+      else
+        require 'net/http'
+        require 'uri'
+        uri = URI.parse("https://tickets.puppetlabs.com/browse/#{ticket[1]}")
+        response = Net::HTTP.get_response(uri)
+        if response.code != "200"
+          raise error_message
+        end
+      end
+    end
+  end
+end
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,5 @@ matrix:
   include:
     - rvm: 2.1.7
       env: "CHECK='rubocop -D'"
+    - rvm: 2.1.7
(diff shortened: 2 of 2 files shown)


Rakefile
```diff
@@ -36,3 +36,26 @@ if File.exist?(build_defs_file)
...
+  # know how to get that information. Absent that, comparing with master should work in most cases.
+  %x{git log --no-merges --pretty=%s master..$HEAD}.each_line do |commit_summary|
+    # This regex tests for the currently supported commit summary tokens: maint, doc, packaging, or pup-<number>.
+    # The exception tries to explain it in more full.
+    if /^\((maint|doc|docs|packaging|pa-\d+|cpr-\d+)\)|revert|bumping/i.match(commit_summary).nil?
+      raise "\n\n\n\tThis commit summary didn't match CONTRIBUTING.md guidelines:\n" \
+        "\n\t\t#{commit_summary}\n" \
+        "\tThe commit summary (i.e. the first line of the commit message) should start with one of:\n"  \
+        "\t\t(pa-<digits>) # this is most common and should be a ticket at tickets.puppetlabs.com\n" \
+        "\t\t(cpr-<digits>) # this is most common and should be a ticket at tickets.puppetlabs.com\n" \
```
mckern: I understand the sentiment but they can't both be **most** common, can they?

Rakefile
```diff
@@ -36,3 +36,26 @@ if File.exist?(build_defs_file)
...
+        "\t\t(docs)\n" \
+        "\t\t(maint)\n" \
+        "\t\t(packaging)\n" \
+        "\n\tThis test for the commit summary is case-insensitive.\n\n\n"
```
mckern: Is this something we could reasonably drop into a HEREDOC?

Rakefile
```diff
@@ -36,3 +36,35 @@ if File.exist?(build_defs_file)
     end
   end
 end
+
+desc "verify that commit messages match CONTRIBUTING.md requirements"
+task(:commits) do
+  %x{git log --no-merges --pretty=%s stable..$HEAD}.each_line do |commit_summary|
```
underscorgan: I don't know if there's a better way to have this use the right branch, rather than hardcoding it.

mckern: We want it to be the branch we're targeting, right?

underscorgan: Yeah, the easy/hacky answer is "have it different on the different branches". I don't know if there's something cleaner though.

mckern: It involves querying the github API, unless it's exposed in ENV vars I don't know about. I'll research this afternoon?

mckern: @underscorgan and I talked about this, but we're gonna look at using some of the [Envrionment Variables provided by Travis](https://docs.travis-ci.com/user/environment-variables/)

mckern:
The only thing I'd change would be throwing some anonymous methods into the Rakefile so that the actual task just called a handful of methods instead of contained a blob of code but that's a style nitpick. This will get the job done. 👍

underscorgan:
this isn't quite working :(

Rakefile
```diff
@@ -36,3 +36,36 @@ if File.exist?(build_defs_file)
     end
   end
 end
+
+desc "verify that commit messages match CONTRIBUTING.md requirements"
+task(:commits) do
+  commits = ENV['TRAVIS_COMMIT_RANGE'] || 'master..HEAD'
```
underscorgan: So, rather than doing hacky things to try to get this working with shallow cloning, let's just use the fact that travis knows what commits are in the PR when doing to check.

Default to master..HEAD when running locally, which probably isn't quite what you want to do, but it should work.

mckern:
I am still 👍 on this even if I hate rake and I hate using more rake.

Rakefile
stahnma: doing a check on master from stable seems odd to me.

mckern: It shouldn't ever happen when running from Jenkins. That's basically a fallback default; I guess I could see changing it to `stable..HEAD` but also, what about not running it at all if `TRAVIS_COMMIT_RANGE` is undefined?

underscorgan: I was worried this was going to end up mangled with the amount of merging we do between branches.

mckern: I am comfortable merging this as-is and fixing it up after some careful experimentation w/ Travis env. vars to try to minimize mangling.

PR #822, opened 2016-08-24
(PA-495) Set library paths for CMake projects explicitly

MikaelSmith:
Previously facter and pxp-agent set CMAKE_INSTALL_RPATH to ensure rpath
is set when installing to non-default locations. Packaging relies on
this to get the correct rpath configuration. Leatherman was missing this
configuration, which was exposed when moving to dynamically linking
against `libstdc++`.

Fix link paths by explicitly setting RPATH for all CMake projects
(in conjunction with settings in the toolchain fails, in accordance with
https://cmake.org/Wiki/CMake_RPATH_handling's "Always full RPATH"
scenario). Also explicitly set LD_LIBRARY_PATH/LIBPATH for test
executables, as they won't have the correct rpath for `libstdc++`.

### DIFF
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -62,6 +62,7 @@ component "cpp-pcp-client" do |pkg, settings, platform|
           -DCMAKE_VERBOSE_MAKEFILE=ON \
           -DCMAKE_PREFIX_PATH=#{settings[:prefix]} \
           -DCMAKE_INSTALL_PREFIX=#{settings[:prefix]} \
+          -DCMAKE_INSTALL_RPATH=#{settings[:libdir]} \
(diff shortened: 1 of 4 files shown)


MikaelSmith:
This completed a successful build against el-7-x86_64 using pl-build-tools-staging. Other systems probably need some testing.

MikaelSmith:
@branan

shrug:
ping @McdonaldSeanp as well

MikaelSmith:
Sorry, did too many things in one commit. I could probably break this into 3 commits.

MikaelSmith:
Done. Code hasn't changed.

shrug:
ping @McdonaldSeanp

stahnma:
👍

shrug:
Also 👍

PR #832, opened 2016-09-02
Mergeup master to stable

McdonaldSeanp:


configs/components/leatherman.json
```diff
@@ -1 +1 @@
-{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "0.8.1"}
+{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "0.9.0"}
```
McdonaldSeanp: Ping @MikaelSmith on this, do we need leatherman 0.9.0 for ruby 2.3 to work? if not I'd like to pin back down to the 0.8.1 where stable was before

McdonaldSeanp:
This still needs one more commit, closing

PR #834, opened 2016-09-02
(maint) Re-add the -j flag to make during ruby build

McdonaldSeanp:
The removal of the parallel make was made during dev for ruby 2.3, and should
have been added back.


mckern:
👍

PR #836, opened 2016-09-02
Mergeup from master to stable

McdonaldSeanp:


configs/components/leatherman.json
```diff
@@ -1 +1 @@
-{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "0.8.1"}
```
McdonaldSeanp: ping @MikaelSmith on this change, do we need leatherman 0.9.0 for ruby 2.3? If not I would like to keep leatherman 0.8.1 for the next release.

MikaelSmith: 0.9.0 is necessary if we're building a newer version of Facter. If we're sticking to facter#stable, we can probably stay at 0.8.1.

McdonaldSeanp: Thanks

geoffnichols:
👍

PR #840, opened 2016-09-08
(PA-540) retag all components not updating in 1.7.0

McdonaldSeanp:
This will also update the SHA of the puppet component to the latest of https://github.com/puppetlabs/puppet/tree/4.7.x


McdonaldSeanp:
What specifically should have a ticket, the pup fix?

MosesMendoza:
I was thinking a ticket to track this effort - some explanatory info on why we're changing these refs like we are.

McdonaldSeanp:
Done, updated to reference a ticket

MosesMendoza:
👍

PR #841, opened 2016-09-10
(PA-542) Add conflicts with PE < 2016.3 components

geoffnichols:
Gem path changes between Ruby 2.1 and 2.3 cause issues for a few
components shipped in PE < 2016.3.

To prevent unintended installation of puppet-agent with Ruby 2.3.1 on
earlier PE versions, this commit adds versioned conflicts for
`pe-puppet-enterprise-release`, `pe-installer`, `pe-r10k`, and
`pe-client-tools`.

### DIFF
--- a/configs/components/ruby.rb
+++ b/configs/components/ruby.rb
@@ -17,6 +17,16 @@ component "ruby" do |pkg, settings, platform|
   pkg.replaces 'pe-ruby-ldap'
   pkg.replaces 'pe-rubygem-gem2rpm'
 
+  # Gem path changes between Ruby 2.1 and 2.3 mean that we don't
+  # want to co-install puppet-agent with Ruby 2.3 on PE < 2016.3
+  # infrastructure nodes, especially ones with components with
+  # vendored gems with Ruby 2.1 gem paths (i.e. pe-r10k).
+  # Conflicts with less than the specified version.
+  pkg.conflicts 'pe-puppet-enterprise-release', '2016.3.0.0'
+  pkg.conflicts 'pe-installer', '2016.1.0.70'
+  pkg.conflicts 'pe-r10k', '2.4.3.1'
+  pkg.conflicts 'pe-client-tools', '1.0.5.3'
+
   base = 'resources/patches/ruby'
 
   # These are a pretty smelly hack, and they run the risk of letting tests


Magisus:
Out of curiosity, what happens as a result of these lines? Does it issue a warning and quit if trying to install with a conflicting component?

geoffnichols:
Results using test build to attempt puppet-agent upgrade on a monolithic CentOS 7 PE 2016.2.1 master:

```
# rpm -Uvh puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64.rpm
error: Failed dependencies:
       	pe-puppet-enterprise-release < 2016.3.0.0 conflicts with puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64
       	pe-installer < 2016.1.0.70 conflicts with puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64
       	pe-r10k < 2.4.3.1 conflicts with puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64
       	pe-client-tools < 1.0.5.3 conflicts with puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64
```

configs/components/ruby.rb
```diff
@@ -17,6 +17,11 @@
   pkg.replaces 'pe-ruby-ldap'
   pkg.replaces 'pe-rubygem-gem2rpm'
 
+  pkg.conflicts 'pe-puppet-enterprise-release', '2016.3.0.0'
```
MosesMendoza: I'm not an expert at the syntax here, but it looks like the subject line says the conflicts should be on PE less than 2016.3 and this implies to me conflicts PE less than _or equal_ to 2016.3. Is that right?

Also, would you mind adding a code comment (maybe with ticket number) explaining why these conflicts are here?

stahnma: @MosesMendoza I'm reading it the way you are.

configs/components/ruby.rb
```diff
@@ -17,6 +17,11 @@
...
+  pkg.conflicts 'pe-installer', '2016.1.0.70'
```
stahnma: Why not version all of these? Is that a debian limitation?

configs/components/ruby.rb
geoffnichols: I'll double-check. My understanding was the conflict is less than the specified version.

https://github.com/puppetlabs/vanagon/blob/master/resources/rpm/project.spec.erb#L59-L61

https://github.com/puppetlabs/vanagon/blob/master/resources/deb/control.erb#L18

configs/components/ruby.rb
geoffnichols: This one is interesting. PE 2016.2.1 shipped with `pe-installer` build `2016.1.0.47`.  Current pre-release builds of PE have `pe-installer` build `2016.1.0.70`.

geoffnichols:
@moses @McdonaldSeanp Updated commit to add code comment.

configs/components/ruby.rb
MosesMendoza: ah I see. thanks

MosesMendoza:
When a user upgrades PE to >= 2016.3.0, will the upgrade process attempt to upgrade puppet agent before it upgrades any of these components?

geoffnichols:
It looks like the upgrade process upgrades puppet-agent and the conflicting components in the same transaction. Install log snippets from test upgrade of mono install of PE 2016.2.1 to a custom 2016.3 pre-release build:

- [el7](https://gist.github.com/geoffnichols/999cfdf3eb98b64f4cd44639e64d51e1)
- [ubuntu 14.04](https://gist.github.com/geoffnichols/f1ae330219d68f85a3491fa4fc3bea0d)

MosesMendoza:
@geoffnichols per conversation this is quite nice of our package managers to do for us. So while previously we wouldn't upgrade pe-bundler, pe-r10k, etc. until later via the puppet_enterprise module, now those packages are getting upgraded in the initial transaction along with the puppet agent (example from your output):

```
--> Running transaction check
---> Package pe-bundler.noarch 0:1.8.2-3.pe.el7 will be updated
---> Package pe-bundler.noarch 0:1.8.2-5.pe.el7 will be an update
---> Package pe-installer.x86_64 0:2016.1.0.47-1.el7 will be updated
---> Package pe-installer.x86_64 0:2016.1.0.70-1.el7 will be an update
---> Package puppet-agent.x86_64 0:1.5.3-1.el7 will be updated
---> Package puppet-agent.x86_64 0:1.6.2.136.g458ee11-1.el7 will be an update
--> Processing Conflict: puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64 conflicts pe-r10k < 2.4.3.1
--> Restarting Dependency Resolution with new changes.
--> Running transaction check
---> Package pe-r10k.x86_64 0:2.3.0.1-1.el7 will be updated
---> Package pe-r10k.x86_64 0:2.4.3.1-1.el7 will be an update
--> Processing Conflict: puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64 conflicts pe-puppet-enterprise-release < 2016.3.0.0
--> Restarting Dependency Resolution with new changes.
--> Running transaction check
---> Package pe-puppet-enterprise-release.noarch 0:2016.2.1.0-1.pe.el7 will be updated
---> Package pe-puppet-enterprise-release.noarch 0:2016.3.0.0-1.pe.el7 will be an update
--> Processing Conflict: puppet-agent-1.6.2.136.g458ee11-1.el7.x86_64 conflicts pe-client-tools < 1.0.5.3
--> Restarting Dependency Resolution with new changes.
--> Running transaction check
---> Package pe-client-tools.x86_64 0:1.0.4.4.g9f88c4a-1.el7 will be updated
---> Package pe-client-tools.x86_64 0:1.0.5.3.gc3b43ad-1.el7 will be an update
--> Finished Dependency Resolution
```

I _think_ this is fine, unless there's some point later in the upgrade that we expect these packages to be their earlier versions for some reason, i.e., for some reason this behavior breaks something in the modules that relies on those packages getting upgraded until later, but this seems like a long shot.

However, as we discussed, this can't actually be fully locally validated - the current system requires us to merge this and promote into PE to actually do full test runs, so it seems like next step is merge.

Ping @jpartlow @ericwilliamson this PR is going into an agent version inbound to PE 2016.3.x, and we'll be watching CI (especially upgrades) for any impact. Can you think of any reason why this change would cause regressions?

PR #842, opened 2016-09-12
(PA-463) Add OS X 10.12 (Sierra) platform

ScottGarman:
Many thanks to Heath Seals for getting this working. The agent builds with this and can apply a few test manifests.

### DIFF
--- a/configs/platforms/osx-10.12-x86_64.rb
+++ b/configs/platforms/osx-10.12-x86_64.rb
@@ -1,0 +1,14 @@
+platform "osx-10.12-x86_64" do |plat|
+  plat.servicetype 'launchd'
+  plat.servicedir '/Library/LaunchDaemons'
+  plat.codename "sierra"
+
+  plat.provision_with 'curl http://pl-build-tools.delivery.puppetlabs.net/osx/homebrew_sierra.tar.gz | tar -x --strip 1 -C /usr/local -f -'
+  plat.provision_with 'ssh-keyscan github.delivery.puppetlabs.net >> ~/.ssh/known_hosts; /usr/local/bin/brew tap puppetlabs/brew-build-tools gitmirror@github.delivery.puppetlabs.net:puppetlabs-homebrew-build-tools'
+  plat.provision_with '/usr/local/bin/brew tap-pin puppetlabs/brew-build-tools'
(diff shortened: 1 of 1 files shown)


McdonaldSeanp:
:+1:

configs/platforms/osx-10.12-x86_64.rb
```diff
@@ -0,0 +1,14 @@
+platform "osx-10.12-x86_64" do |plat|
+  plat.servicetype 'launchd'
+  plat.servicedir '/Library/LaunchDaemons'
+  plat.codename "sierra"
+
+  plat.provision_with 'curl http://pl-build-tools.delivery.puppetlabs.net/osx/homebrew_sierra.tar.gz | tar -x --strip 1 -C /usr/local -f -'
+  plat.provision_with 'ssh-keyscan github.delivery.puppetlabs.net >> ~/.ssh/known_hosts; /usr/local/bin/brew tap puppetlabs/brew-build-tools gitmirror@github.delivery.puppetlabs.net:puppetlabs-homebrew-build-tools'
```
MosesMendoza: Why are the keyscan and brew command on the same line?

MosesMendoza: (just out of curiosity)

mckern: They were likely copied in from a previous definition (10.11 or 10.10) and I believe the original reason was for brevity. They **are** conceptually related (since git pulls in the Homebrew tap over SSH) but this isn't very clear.

configs/platforms/osx-10.12-x86_64.rb
```diff
@@ -0,0 +1,14 @@
+platform "osx-10.12-x86_64" do |plat|
+  plat.servicetype 'launchd'
+  plat.servicedir '/Library/LaunchDaemons'
+  plat.codename "sierra"
+
+  plat.provision_with 'curl http://pl-build-tools.delivery.puppetlabs.net/osx/homebrew_sierra.tar.gz | tar -x --strip 1 -C /usr/local -f -'
```
heathseals: just a heads up that the homebrew_sierra tarball was just a quick way to get us building.  we were using a much older homebrew tarball that did not have definitions for sierra.  this more recent tarball is a lot more verbose ( Homebrew/brew@27b0ca4 ) about running builds as root ( as in don't do it ).

mckern: I'd like to refactor the way we use Homebrew a bit soon if possible, so should we roll running Brew based builds as someone other than root?

PR #844, opened 2016-09-13
(maint) move the puppet component SHA to stable

McdonaldSeanp:
The old SHA was from the 4.7.x branch, which contained the correct commits,
however during the mergup after the release will dirty up the commit history.
Stable has been cleaned up and is ready to go in to the RC


MosesMendoza:
@McdonaldSeanp and this new ref includes the fix for PUP-6636?

McdonaldSeanp:
@MosesMendoza Yep

ScottGarman:
:+1:

PR #846, opened 2016-09-14
(PA-544) Fix solaris sparc cross-builds of ruby socket library

ScottGarman:
When cross-compiling ruby 2.3.1 on solaris sparc, the socket library
will not be built unless you specify a close-fds-by-recvmsg-with-peek
configure option.

### DIFF
--- a/configs/components/ruby.rb
+++ b/configs/components/ruby.rb
@@ -162,7 +162,8 @@ component "ruby" do |pkg, settings, platform|
         pkg.build_requires 'pl-ruby'
       end
 
-      special_flags += " --with-baseruby=#{settings[:host_ruby]} "
+      # The rcvmsg flag is needed to cross-compile the socket library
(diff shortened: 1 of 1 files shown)


stahnma:
Assuming this fixes the issue, LGTM. 👍

MosesMendoza:
@ScottGarman this one's already been merged but it would be great to include a link in the code comment to the compiler doc that further explains this flag and why this is necessary.

Also good on a PR like this would be a link to successful local tests run against this change or if that's impossible just a CLI or command line output of before/after the change, e.g.,

```
# before this PR
require 'socket'
< broken all over >
```

```
# after this PR
require 'socket'
< everything is awesome >
```

Also, I was wondering - given our sparc CI is currently manually triggered, is there any form of `make test` that we could run as part of compilation, maybe against just a subset of tests or even one that could act as a compilation smoke test? Maybe worth a quick time boxed  investigation?

PR #848, opened 2016-09-14
(RE-7842) Add osx 10.12 Sierra to nightly foss platforms

ScottGarman:


ScottGarman:
Don't merge this just yet - I need to first confirm that package signing still works the same on Sierra.

ScottGarman:
I've verified package signing on osx sierra works with our packaging rake task without modification, so this is now ready for merge.

McdonaldSeanp:
:+1:

PR #849, opened 2016-09-14
(PA-546) Use prebuilt nokogiri based on ruby 2.3

ScottGarman:
We had forgotten to create a new prebuilt nokogiri tarball during the ruby 2.3.1 bump, so this addresses that. Tested an agent built with these changes by verifying nokogiri and net-netconf were installed via 'gem list', and ran ldd on the nokogiri library file to confirm it is linking against our vendored libxml2 and libxslt libs. 

Note: this workaround is only done for huaweios. We need to vendor nokogiri as it is a dependency of net-netconf and HuaweiOS doesn't come with compiler tools by default.

### DIFF
--- a/configs/components/rubygem-nokogiri.rb
+++ b/configs/components/rubygem-nokogiri.rb
@@ -5,8 +5,8 @@ component "rubygem-nokogiri" do |pkg, settings, platform|
     # including its dependencies (mini_portile2), installing it on the target, and
     # installing the nokogiri gem using our gem command. The added files from
     # :gem_home were then tar'ed up and packed into this container tarball.
-    pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-precompiled-huaweios-#{pkg.get_version}.tar.gz"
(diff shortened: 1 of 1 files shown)


mckern:
Is this another one-off Nokogiri build especially for this particular package, that would be difficult to reproduce in an isolated sandbox?

ScottGarman:
@mckern I'm not quite sure what you mean. In theory we could look into trying to automate the generation of this tarball, but it would require spinning up a Debian Jessie PowerPC instance running on QEMU or something, and in the past we've found trying to compile things in a cross-arch QEMU environment is abysmally slow. So what I've been doing instead is actually running them on a Huawei switch, and that requires a bunch of manual setup and teardown that isn't especially script-friendly.

mckern:
I'm sad we're building another iteration of a previous special snowflake instead of normalizing how the way that components are built for this platform. It's another item in the "fix our cross compile process" laundry list.

McdonaldSeanp:
:+1:

branan:
I basically agree entirely with Ryan on this, but given our current plans and schedule... ¯\_(ツ)_/¯

PR #850, opened 2016-09-15
(PA-545) Update all platforms to use scripts to run executables

McdonaldSeanp:
This will update the files under /opt/puppetlabs/bin to be bash scripts
instead of symlinks. This allows us to set the environment in a specific way
to get around issues with LD_PRELOAD

### DIFF
--- a/resources/files/facter.sh
+++ b/resources/files/facter.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+unset LD_PRELOAD
+unset LD_LIBRARY_PATH
+
+exec /opt/puppetlabs/puppet/bin/facter $@
+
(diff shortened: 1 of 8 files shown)


MosesMendoza:
This feels like a pretty big change to me... Why is this change necessary? 

Are there other implications or side-effects of changing our executables to bash scripts instead of the original ruby scripts?

resources/files/facter.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+unset LD_PRELOAD
+unset LD_LIBRARY_PATH
```
MosesMendoza: Why are we modifying `LD_LIBRARY_PATH` prior to executing any of Puppet's executables? This feels like it could have side effects in a user's environment and be very surprising to them, especially if they have their own LD_LIBRARY_PATH settings in place?

MosesMendoza: Also if the whole point of changing the symlinks to bash scripts is to modify these variables, make it makes more sense to modify these variables in the actual project executables rather than create this extra layer of indirection?

geoffnichols:
Could the patch to `configure.in` at https://github.com/ruby/ruby/commit/68c062b1f06f384df91a198e69495f185b12e334 possibly account for the difference in behavior between Ruby 2.1 and 2.3?

PR #851, opened 2016-09-15
(PA-545) Update cisco to use scripts to run executables

McdonaldSeanp:
This will update the files under /opt/puppetlabs/bin to be bash scripts
instead of symlinks. This allows us to set the environment in a specific way
to get around issues with LD_PRELOAD


ScottGarman:
:+1:  assuming you've built a package and verified the fix works.

PR #858, opened 2016-09-19
Mergup from 1.7.0-release

McdonaldSeanp:


shrug:
What is the intended effect of doing the merge-up before the tag?

McdonaldSeanp:
We would like to start doing mergeups as changes occur, as opposed to all at once after a tag.

shrug:
Understood, I just don't see any changes...

As originally envisioned, changes would all land on stable and be cherry-picked to the release branch (except for the json file updates to pin to tags), so merge-ups would only be required to pick up tags. If we're going to be making changes in the release branch, then 👍 to merge-ups.

McdonaldSeanp:
Yeah, there was a bunch that happened in the release branch so this is in order to keep history the same between release and stable. in the ideal world your thoughts are correct, the only mergedown will be the tag

PR #864, opened 2016-09-26
Add internal_list key to MAINTAINERS

theshanx:
This change adds a reference to the Google group the maintainers are associated with.


MosesMendoza:
@theshanx I opened https://github.com/puppetlabs/puppet-agent/pull/868 that has your commit but with an amended prefix of "(maint)" which we like to have (or a ticket number) in our commit messages. That PR will supersede this one

PR #866, opened 2016-09-29
(PUP-572) Use correct RPATH for runtime on SPARC

branan:
When we updated the compiler runtime to have an RPATH, we failed to
bring the right version of GCC to SPARC. This commit bumps the GCC patch
version so that we can ship it correctly.


ScottGarman:
:+1:

PR #868, opened 2016-09-30
(maint) Add internal_list key to MAINTAINERS

MosesMendoza:
This change adds a reference to the Google group the maintainers are associated with.


Magisus:
👍

PR #870, opened 2016-10-03
(PA-575) Disable curl builds for cisco-wrlinux platforms

ScottGarman:
cisco-wrlinux platforms are not compute clusters, so we can safely
disable curl builds for them.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -125,7 +125,7 @@ component "facter" do |pkg, settings, platform|
 
   # curl is only used for compute clusters (GCE, EC2); so rpm, deb, and Windows
   skip_curl = 'ON'
-  if (platform.is_linux? && !platform.is_huaweios?) || platform.is_windows?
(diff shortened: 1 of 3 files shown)


ScottGarman:
Will post a link to a passing jenkins ad-hoc build when complete.

ScottGarman:
https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/369/ is looking green. The SLES failures are a known issue being fixed by RelEng, and the travis falure is just due to the fact that this ticket number isn't public.

branan:
This looks reasonable to me.

@ScottGarman before we merge this, would you mind filing a ticket for moving the "is this a compute platform" logic someplace centralized? This is a great example of the duplication I'd like to see us trying to clean up (especially since this conditional is now fairly complex).

McdonaldSeanp:
@branan, we should have a conversation about why I don't think we should do that, if you convince me otherwise I'll open that ticket myself. Please merge this PR

PR #871, opened 2016-10-06
(PA-520) Add armel support for builds

stahnma:
This adds initial support for building debian-8-armel via a cross
compile.

Caveats:
  - [x] We need a merge of https://github.com/puppetlabs/cpp-pcp-client/pull/203


stahnma:
This should be g2g, assuming cpp-pcp-client promotes into master at some point soon.

stahnma:
@jhg03a yeah, that's the same for all the other platforms. (not a new problem). It's being worked on, but slowly 😞

configs/projects/puppet-agent.rb
```diff
@@ -94,6 +94,7 @@
   platform_triple = "powerpc64le-unknown-linux-gnu" if platform.architecture == "ppc64le"
   platform_triple = "s390x-linux-gnu" if platform.architecture == "s390x"
   platform_triple = "arm-linux-gnueabihf" if platform.name == 'debian-8-armhf'
+  platform_triple = "arm-linux-gnueabi" if platform.name == 'debian-8-armel'
```
stahnma: Why shouldn't this be here?

PR #872, opened 2016-10-10
(PA-544) Fix solaris sparc cross-builds of ruby socket library

ScottGarman:
When cross-compiling ruby 2.3.1 on solaris sparc, the socket library
will not be built unless you specify a close-fds-by-recvmsg-with-peek
configure option.

Note: This commit got reverted back when we decided to release 1.7.0 with ruby 2.2 instead of ruby 2.3.1, so this PR is to restore it with clearer commenting.


MosesMendoza:
👍

PR #874, opened 2016-10-18
(PA-620) Supply ssl cert for rubygems.org

Magisus:
Rubygems recently updated its ssl cert. This patches the new cert into the Rubygems cert bundle in our vendored Ruby.

I build a Windows MSI from this branch and installed it. The cert was dropped in the correct directory, and I was able to successfully add `https://rubygems.org` as a gem source.

### DIFF
--- a/configs/components/ruby.rb
+++ b/configs/components/ruby.rb
@@ -233,6 +233,12 @@ component "ruby" do |pkg, settings, platform|
     end
     pkg.directory settings[:ruby_dir]
   end
+
+  # This became necessary when rubygems.org updated its SSL cert. This patches the new cert into the Rubygems
+  # cert bundle in our vendored Ruby. See PA-620.
+  pkg.add_source "http://buildsources.delivery.puppetlabs.net/GlobalSignRootCA.pem", sum: "7872bab9a5e1a71e3d7f77836a841a04"
+  pkg.install_file "../GlobalSignRootCA.pem", "#{settings[:libdir]}/ruby/2.3.0/rubygems/ssl_certs"
+
   if platform.is_cross_compiled_linux? || platform.is_solaris? || platform.is_aix?
     # Here we replace the rbconfig from our ruby compiled with our toolchain
     # with an rbconfig from a ruby of the same version compiled with the system


McdonaldSeanp:
👍

McdonaldSeanp:
@Magisus Can you add a comment above the code changes with the commit info, so our future selves know what's going on?

Magisus:
@McdonaldSeanp udpated

configs/components/ruby.rb
```diff
@@ -233,6 +233,12 @@
     end
     pkg.directory settings[:ruby_dir]
   end
+
+  # This became necessary when rubygems.org updated its SSL cert. This patches the new cert into the Rubygems
+  # cert bundle in our vendored Ruby. See PA-620.
+  pkg.add_source "http://buildsources.delivery.puppetlabs.net/GlobalSignRootCA.pem", sum: "7872bab9a5e1a71e3d7f77836a841a04"
```
MosesMendoza: Should this only be on Windows, or did we decide to add it for all platforms?

Magisus: We decided to do it on all platforms.

Iristyle:
Upon further discussion, I'm not sure this is actually needed.

Our stance in the past has been that if you want to install gems on Windows with the Puppet distribution of Ruby, you start with launching a `Start Command Prompt with Puppet` as described in https://github.com/puppetlabs/puppetlabs-azure#installing-the-azure-module .  One of the reasons we have that command prompt is to call `environment.bat` and init all relevant ENV vars.

Unfortunately it sounds like that information is not widely disseminated (it's not in the AWS or vSphere module docs for instance) - so I think we could make some improvements to those `README` files at the very least.

Iristyle:
There is a manifest based workaround we can promote for users of puppet-agent < 1.6.0 at https://gist.github.com/Iristyle/4434a36fe3d168212f2318eda0ade51c

A blog post is underway to notify end users of the issue. It includes the above referenced code so that Puppet may fix Puppet on older agent releases.

joshcooper:
The new cert is going to be added to rubygems: https://github.com/ruby/ruby/blob/ruby_2_3/lib/rubygems/ssl_certs/GlobalSignRootCA.pem as soon as we upgrade to 2.3. We might want to understand what the AIX issues were so that we don't run into the same issue when we try to upgrade ruby. I don't know if ruby backported the change to earlier versions...

Magisus:
I don't think the AIX issue had anything to do with the cert itself, but more likely with the way we were inserting it. The build choked when examining the directory.

PR #876, opened 2016-10-19
(PA-554) Create switchable ruby versions

McdonaldSeanp:
Add functionality to allow developers to switch vendored ruby version from
2.1.9 to 2.3.1 freely with minimal changes. This is achieved through the
following:

* All resources required by ruby (i.e. the rbconfigs, specific patches, batch
files for windows) have been moved in to version specific directories, with
the moniker ruby_[version]

* the ruby component has been updated to point to the new source directories
for resources

* Resources and the ruby config from ruby 2.1.9 have been added to the
project. the ruby component configs for both ruby 2.3.1 and 2.1.9 have been
renamed to ruby-2.3.1 and ruby-2.1.9 respectively

* A ruby_version setting was added to the project

* All components that had a build_requires "ruby" have been updated to use:

pkg.build_requires "ruby-#{settings[:ruby_version]}"

to reference the newly renamed ruby component configs.

* the precompiled nokogiri binary for hauwei has been renamed to:

nokogiri-precompiled-huaweios-#{pkg.get_version}-for-ruby-#{settings[:ruby_version]}.tar.gz

to allow us to specify ruby version in the binary

To change the ruby version, all that needs to be done now is to update
ruby_version and gem_home to the appropriate values for either 2.3.1 or 2.1.9


Magisus:
Seems legit, and fairly straightforward to add new versions when we need to. Nice work! 👍

MosesMendoza:
@McdonaldSeanp would it be possible for you to post diffs (if any) of the content a package built prior to this change vs with it?

configs/components/rubygem-nokogiri.rb
```diff
@@ -5,14 +5,18 @@
     # including its dependencies (mini_portile2), installing it on the target, and
     # installing the nokogiri gem using our gem command. The added files from
     # :gem_home were then tar'ed up and packed into this container tarball.
-    pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-precompiled-huaweios-#{pkg.get_version}.tar.gz"
-    pkg.md5sum "8a783bdda33cf5f10afe84cf6fd8f28c"
+    pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-precompiled-huaweios-#{pkg.get_version}-for-ruby-#{settings[:ruby_version]}.tar.gz"
+    if settings[:ruby_version] == "2.1.9"
```
MosesMendoza: It kind of seems to me like a cleaner way to do this would to have the huaweios precompiled nokogiri be it's own separate source with a separate component file, and then select it based on platform/ruby version.. i don't know if that's possible though..

I.e., 
`configs/components/rubygem-nokogiri-huaweio-2.3.1.rb`
`configs/components/rubygem-nokogiri-huaweio-2.1.9.rb`
`configs/components/rubygem-nokogiri.rb`

and then select the right one based on what platform/ruby version we're building?

MosesMendoza:
reviewing

configs/components/rubygem-nokogiri.rb
McdonaldSeanp: @MosesMendoza while this component does seem a little dirty, I disagree that the three components would be a cleaner solution. The way that I updated this functionality is consistent with all the other components of puppet-agent when checking for platform when platform specific changes are required.

The argument against using different component files would be that if we needed to update something across the board (i.e. something about the install command needed to change) we would need to update it in 3 different places and not one.

I would however like to clean up this component to fit with the https://github.com/puppetlabs/vanagon/blob/master/docs/STYLEGUIDE.md set forth in vanagon. But I think that's out of scope for this PR.

McdonaldSeanp:
@MosesMendoza the content should be entirely unchanged. There will be no diff.

MosesMendoza:
@McdonaldSeanp cool, I also would expect no diff. I think explicitly confirming that would be great though..

McdonaldSeanp:
@MosesMendoza:
```
rpm -qlp puppet-agent-1.7.0.138.g86b2768-1.el7.x86_64.rpm > old.log
rpm -qlp puppet-agent-1.6.2.294.gf518851-1.el7.x86_64.rpm > new.log
```
And the diff:
```
diff new.log old.log
8432,8433c8432,8433
< /usr/share/doc/puppet-agent-1.6.2.294.gf518851
< /usr/share/doc/puppet-agent-1.6.2.294.gf518851/bill-of-materials
---
> /usr/share/doc/puppet-agent-1.7.0.138.g86b2768
> /usr/share/doc/puppet-agent-1.7.0.138.g86b2768/bill-of-materials

```

configs/components/rubygem-nokogiri.rb
McdonaldSeanp: Yeah, we had many discussions about how to clean up components in RE, and I think the conclusions we came to were good. We wanted to revisit the cleanliness of components and better ways to remove some conditionals, but not until we updated all the configs to use the styleguide. That way when all the components were using the same styleguide we could identify patterns that we could eliminate.

McdonaldSeanp:
Also, on a testing note: This PR has gone through the ad-hoc pipeline with both ruby versions 2.3.1 and 2.1.9. The completed jobs at:

2.3.1: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/96/

2.1.9: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/97/

MosesMendoza:
@McdonaldSeanp nice!! I love that diff :)

Any chance you would be able to update the README to include instructions on how someone building this themselves (say for an ad-hoc run) would switch the ruby version?

McdonaldSeanp:
Sure =)

McdonaldSeanp:
@MosesMendoza Updated

MosesMendoza:
👍

PR #880, opened 2016-10-20
(PA-629) --CRITICAL-- Update el-4-x86 to use redhat builder

McdonaldSeanp:
Centos-4-x86_64 vmpooler vms are non-responsive and will not work as builders.
We need to update the build job to use redhat instead of centos. The issue
with that is redhat 4 does not use yum, so we needed to update
plat.provision_with to use RPM to install the centos yum packages available
on our osmirror for centos. Then update the yum install to correctly pull from
centos repos,

country style


McdonaldSeanp:
Currently doing a manual build, this will be blocked on that.

MosesMendoza:
@McdonaldSeanp could you update with ticket?

McdonaldSeanp:
@MosesMendoza updated

McdonaldSeanp:
closing, will get this in a mergeup of stable -> master

PR #881, opened 2016-10-20
(PA-629) --CRITICAL-- Update el-4-x86 to use redhat builder

McdonaldSeanp:
Centos-4-x86_64 vmpooler vms are non-responsive and will not work as builders.
We need to update the build job to use redhat instead of centos. The issue
with that is redhat 4 does not use yum, so we needed to update
plat.provision_with to use RPM to install the centos yum packages available
on our osmirror for centos. Then update the yum install to correctly pull from
centos repos,

country style

When the centos hosts are available again, we should revert this work and use
centos-4-x86_64 again.


MosesMendoza:
👍  on link of successful build, and assuming eventual revert as documented in PA-629

Magisus:
I agree with @MosesMendoza 👍

PR #887, opened 2016-10-27
(PA-575) Disable curl builds for cisco-wrlinux platforms

Magisus:
cisco-wrlinux platforms are not compute clusters, so we can safely
disable curl builds for them.

Note: this reapplies these changes, as some of them were lost in the
merge from master to stable.


Magisus:
The ticket this commit refers to is still internal, which is causing Travis to fail.

PR #891, opened 2016-10-29
(maint) Update Facter and Puppet refs to include acceptance test changes

Magisus:
Some of the acceptance tests for Puppet and Facter do not work correctly
on the eccentric platforms. This updates the SHAs for those components
to commits that include acceptance test changes which will allow
validation on the manual pipelines.


geoffnichols:
👍

PR #897, opened 2016-11-08
(PA-679) Allow optional attributes on DOCTYPE ENTITY

abottchen:
Prior to this patch, the Xml augeas lens would fail to parse
a file containing a DOCTYPE with an ENTITY using the "SYSTEM" or
"PUBLIC" optional parameters.  With this patch in place, those
files are parsed successfully, with the optional attributes stored
as the label.

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -6,6 +6,7 @@ component 'augeas' do |pkg, settings, platform|
   pkg.replaces 'pe-augeas'
   pkg.apply_patch 'resources/patches/augeas/osx-stub-needed-readline-functions.patch'
   pkg.apply_patch 'resources/patches/augeas/sudoers-negated-command-alias.patch'
+  pkg.apply_patch 'resources/patches/augeas/entitys-allow-system-public-option.patch'
   if platform.is_sles? && platform.os_version == '10'
     pkg.apply_patch 'resources/patches/augeas/augeas-1.2.0-fix-services-sles10.patch'
   end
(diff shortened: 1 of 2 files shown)


joshcooper:
Can we get this merged upstream so we don't have to carry a patch? I hear @lutter knows a thing or two about augeas?

lutter:
More than happy to - would be great if there were some tests, too. Or, at a minimum, a couple of examples of the syntax that can now be parsed.

lutter:
Looking at the definitions [here](https://www.w3.org/TR/xml11/#sec-external-ent), I think the lens is not quite right. It's ok for `SYSTEM` entities, but `PUBLIC` entities have two strings after them: a public ID and a system id.

Before addressing this, I'd definitely want to know whether that is just an oversight or whether this deviation from the standard was intentional.

abottchen:
That was an oversight, I've attached a new version that covers both the SYSTEM option with one entity and PUBLIC with two.  With this patch in place, I can parse DOCTYPE ENTITY with SYSTEM and PUBLIC specified and they show up correctly in the tree.  However, I cannot write them out to a file.  I do not understand augeas lenses well enough to say why I can the get is working but the put is failing.  Perhaps @lutter  could offer some advice?

lutter:
Posted this PR upstream that takes care of this issue: https://github.com/hercules-team/augeas/pull/426

MosesMendoza:
Closing as superseded by https://github.com/hercules-team/augeas/pull/426

PR #898, opened 2016-11-08
(PA-676) Fix cross-compilation of ruby-selinux

ScottGarman:
Prevent the cross-compiled builds of ruby-selinux from finding the selinux include headers in /usr/include, and specify the sysroot includedir instead.

Did some runtime testing of this package by applying puppet test manifest, and ensuring I could require 'selinux' successfully from our irb command.


ScottGarman:
Also, I did some multi-platform build testing with this on the ad-hoc jenkins builder, which is fully green:

https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/412/

Since the only platforms this change applies to are the cross-compiled platforms that use selinux (which are only redhat platforms), this should be sufficient to ensure this change isn't going to break other platforms.

ScottGarman:
Closing this PR in favor of https://github.com/puppetlabs/puppet-agent/pull/899 which targets the LTS branch instead of master.

PR #900, opened 2016-11-09
(MAINT) Use Homebrew env. vars to prevent 💩

mckern:
Homebrew 1.1.0 removes SHA1 sums (fine, whatever), but it also removes
the ability to run `brew` as root. Yes, this is a good idea but we
have to refactor the entire macOS build process to fix that from our
end.

But "we don't upgrade Homebrew! We use a snapshot!" you say. Well,
true. But `brew` will automatically, and silently, upgrade itself in
the background when you tap a new Homebrew keg... unless you tell it
not to.

While I was in there, I also turned off emoji and turned up verbosity.
Might as well, right?

Yes, this is specific to macOS Sierra right now. It's a snowflake.

### DIFF
--- a/configs/platforms/osx-10.12-x86_64.rb
+++ b/configs/platforms/osx-10.12-x86_64.rb
@@ -3,6 +3,9 @@ platform "osx-10.12-x86_64" do |plat|
   plat.servicedir '/Library/LaunchDaemons'
   plat.codename "sierra"
 
+  plat.provision_with 'export HOMEBREW_NO_AUTO_UPDATE=true'
+  plat.provision_with 'export HOMEBREW_NO_EMOJI=true'
+  plat.provision_with 'export HOMEBREW_VERBOSE=true'
   plat.provision_with 'curl http://pl-build-tools.delivery.puppetlabs.net/osx/homebrew_sierra.tar.gz | tar -x --strip 1 -C /usr/local -f -'
   plat.provision_with 'ssh-keyscan github.delivery.puppetlabs.net >> ~/.ssh/known_hosts; /usr/local/bin/brew tap puppetlabs/brew-build-tools gitmirror@github.delivery.puppetlabs.net:puppetlabs-homebrew-build-tools'
   plat.provision_with '/usr/local/bin/brew tap-pin puppetlabs/brew-build-tools'


mckern:
I really channeled my inner @rick on this one.

rick:
> Oh, once I knew what that wisdom was for / September Girls think those pearls just wash up on the shore

configs/platforms/osx-10.12-x86_64.rb
```diff
@@ -3,6 +3,9 @@
   plat.servicedir '/Library/LaunchDaemons'
   plat.codename "sierra"
 
+  plat.provision_with 'export HOMEBREW_NO_AUTO_UPDATE=true'
+  plat.provision_with 'export HOMEBREW_NO_EMOJI=true'
+  plat.provision_with 'export HOMEBREW_VERBOSE=true'
```
rick: :lol:

rick:
Or, more appropriately: http://www.azlyrics.com/lyrics/destroyer/thebadarts.html

rick:
:bow:

Magisus:
I have run a build and verified this solves the issue. Would you mind updating the commit message to include the associated ticket number, since there IS a ticket: https://tickets.puppetlabs.com/browse/RE-8331.

Magisus:
Update: the build got quite a bit further, but failed trying to find Boost when building leatherman. Could this be part of the Homebrew issues? Build --> https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/BUILD_TARGET=osx-10.12-x86_64,SLAVE_LABEL=beaker/415/console

Magisus:
Also, this issue is also affecting the LTS and stable branches. Can you retarget this PR to the LTS branch, and then we'll merge up from there?

mckern:
@Magisus I actually pulled the ticket number out because the ticket is not a public ticket. I originally had the ticket in there and the Travis tests failed due to it not being in the PA space.
I am unsure about the Boost issue, but that seems like the next thing to dig into.
RE: retargetting: sure.

mckern:
Retargeted. Can dig into the Boost problems next if desired, or can leave it to @branan

underscorgan:
👍

mckern:
For reasons I do not understand, Boost is not being installed. I need to dig into the OS X provisioner script we use, and that's not tracked as part of this repo.

mckern:
@Magisus I have added these env. vars to the `osx-deps` script and confirmed that this lets us get back to the business of building what we need.

ScottGarman:
:+1:

PR #901, opened 2016-11-09
(PA-678) Cause puppet-agent to conflict with r10k versions less than 2.5.0

Magisus:
After version 2.5.0, pe-r10k begins vendoring the gettext gems, which
puppet-agent also vendors as of 1.8.0, causing these packages to
conflict. This adds a conflicts statement, which prevents them from
being installed together and ensures upgrades are handled correctly.

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -1,6 +1,11 @@
 project "puppet-agent" do |proj|
   platform = proj.get_platform
 
+  # (PA-678) pe-r10k versions later than 2.5.0 ship gettext gems.
+  # Since we also ship those gems as part of puppet-agent
+  # at present, we need to conflict with pe-r10k > 2.5.0
+  proj.conflicts "pe-r10k", "2.5.0"
+
   # Project level settings our components will care about
   if platform.is_windows?
     proj.setting(:company_name, "Puppet Inc")


Magisus:
It still needs to be validated that the conflict actually works. Any help with how to do that would be appreciated.

configs/projects/puppet-agent.rb
```diff
@@ -1,6 +1,8 @@
 project "puppet-agent" do |proj|
   platform = proj.get_platform
 
+  proj.conflicts "pe-r10k" "2.5.0"
```
McdonaldSeanp: This is missing a comma, should be:

`proj.conflicts "pe-r10k", "2.5.0"`

McdonaldSeanp:
@Magisus there was an error, note the line comment

Magisus:
Updated.

McdonaldSeanp:
I am currently building, and will test an el-7 package

configs/projects/puppet-agent.rb
```diff
@@ -1,6 +1,8 @@
 project "puppet-agent" do |proj|
   platform = proj.get_platform
 
+  proj.conflicts "pe-r10k", "2.5.0"
```
MosesMendoza: Would you mind adding a code comment for our future selves explaining why this conflicts is here? Also, maybe include the version of PE that this conflicts with in said comment?

MosesMendoza: as well as the ticket # ?

McdonaldSeanp:
I have validated the updated code works and will conflict.

McdonaldSeanp:
👍

Magisus:
Thanks @McdonaldSeanp!

mckern:
I am 👍 on this and willing to merge it if previous 👍 's stand.

PR #904, opened 2016-11-10
(PA-633) Add acceptance testing setup for Puppet Agent

MosesMendoza:
This PR adds the ability to add and run acceptance tests to Puppet Agent.

Prior to this PR, Puppet Agent could only be tested by running acceptance tests from its various components in 'package testing' (aio) mode, which indirectly tested that the packages themselves were sane. Unfortunately, this led to a number of problems, including:
- Much of the package creation and manipulation logic is untested, which means it is currently difficult to know if changes will introduce regressions
- What package testing does exist has been spread out across repos including puppet and beaker, making it difficult to find, augment, and maintain
- We don't have any automated testing to validate that our shipped packages are viable/usable by users

This PR starts with the basic framework in use by the Puppet Agent components, that is:
- a Rakefile that defines tasks and methods which drive beaker
- a set of pre-suite setup tasks for installation and configuration of Puppet Agent
- a tests directory containing (in this PR) a single test which is a good candidate for one to move from Puppet to Puppet Agent

This PR deviates from the component acceptance tests:
- No 'gem' or 'git' acceptance testing is added (these are component-driven delivery mechanisms)
- Some obsolete workaround logic has been removed
The majority of differences exist in `acceptance/Rakefile` and `acceptance/setup/aio/pre-suite/010_Install.rb`.
- Perhaps most notably this PR doesn't actually (yet) test the source of puppet agent or support testing a local cache of built package artifacts. Rather, it assumes that packages of the version specified to test via ENV['SHA'] exist on one of the supported package delivery mechanisms - `builds.delivery.puppetlabs.net` (puppet internal), `nightlies.puppetlabs.com` (public), or our production download repositories/site (public). If these tests are added to the end of the existing Puppet Agent test pipeline, this should always be true as earlier steps in the pipeline generate PA artifacts to test from. This is a substantial TODO that needs ticketed.

Usage
As documented in the `acceptance/README.md` in the PR, usage is drive by rake:
`rake acceptance:development`
- Run acceptance tests against Puppet Agent packages on unreleased repos (ie builds.delivery.puppetlabs.net for internal and nightlies.puppet.com for public)
`rake acceptance:released`
- Run acceptance tests against Puppet Agent packages from released packages existing on our production Puppet repos, ie yum.puppetlabs.com. This will allow us to validate that our shipped production packages are sane.

For more info on the specific configuration/environment variables supported see the README.md.

Out of scope from this PR is describing how frequently and when P-A acceptance tests will be run, adding this to CI, or additional scouring of repos to pull out tests that really belong in this one. That needs to be ticketed separately.

A temporary commit (temp commit to fix ssl) is included in the PR to work around PA-620 in case folks want to pull this down to try out themselves, but will be removed before merge.


acceptance/config/nodes/aix-53-power.yaml
```diff
@@ -0,0 +1,19 @@
+---
+HOSTS:
+  master:
+    roles:
+      - master
+    platform: el-7-x86_64
+    hypervisor: vcloud
```
McdonaldSeanp: This should be vmpooler

McdonaldSeanp: The same goes for all the subsequent files with a hypervisor entry under el-7

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -0,0 +1,152 @@
...
+          # Passenger, in particular, must be shutdown for the cert setup steps to work,
+          # but any running puppet master will interfere with webrick starting up and
+          # potentially ignore the puppet.conf changes.
+          on(master, puppet('resource', 'service', master['puppetservice'], "ensure=stopped"))
+        end
+
+        step "Clear SSL on all hosts"
+        hosts.each do |host|
+          ssldir = on(host, puppet('agent --configprint ssldir')).stdout.chomp
+          on(host, "rm -rf '#{ssldir}'")
```
McdonaldSeanp: Doesn't this remove the actual #{ssldir} directory? There may be something I don't understand but If i'm not mistaken we just want to remove all the certs under that directory, not the directory itself right?

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -0,0 +1,152 @@
...
+        end
+
+        step "Master: Start Puppet Master" do
+          master_opts = {
+            :main => {
+              :dns_alt_names => "puppet,#{hostname},#{fqdn}",
+            },
+            :__service_args__ => {
+              # apache2 service scripts can't restart if we've removed the ssl dir
```
McdonaldSeanp: And here it looks like we are working around the above comment about how we removed the entire ssldir and not just the contents

McdonaldSeanp:
I feel as though since this is only testing changes, the PR should target the LTS branch

MosesMendoza:
@McdonaldSeanp 👍  on retargeting against LTS. Given the git histories this isn't just a simple rebase so I'm going to leave this PR open to gather feedback against, but will do a retarget/merge manually once it's ready for merge.

acceptance/lib/puppet/acceptance/common_utils.rb
MosesMendoza: I hadn't actually looked to closely at this - just copied from puppet. I think we actually tell folks to rm the ssl dir - that's what we say here anyway: https://docs.puppet.com/pe/latest/agent_cert_regen.html. I'm not sure it makes a difference.

McdonaldSeanp:
@MosesMendoza I approve of your explanations of the ssl_dir stuff. I'll give this a 👍 if you update the `hypervisor: vcloud` entries to `hypervisor: vmpooler`

MosesMendoza:
Thanks for the reminder @McdonaldSeanp - I had forgotten about that bit. Pushed a commit that updates.

MosesMendoza:
@McdonaldSeanp I'm going to close this now and open a new one w/that last commit squashed against the LTS-1.7 branch

PR #905, opened 2016-11-11
(PA-682) Bump curl to 7.51.0

Magisus:
This updates the version of curl used in our builds to 7.51.0, in
response to a recent security update.


Magisus:
Validated manually on el4, verified curl could be used. Result of a successful build on some weird platforms is here: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/416/
Note: the failed platform was a pooler issue.

McdonaldSeanp:
👍

PR #907, opened 2016-11-15
(maint) Fix XCC of curl for arm platforms

stahnma:
Prior to this commit debian-8-arm* platforms were getting a recv error
during configure on curl. This was due to using system pkg-config and
not the one found in /opt/pl-build-tools. This commit simply adjusts
PATH to prefer /opt/pl-build-tools/bin for cross_compiled_linux
platforms in the curl component.

This was a new problem introduced with the recent upgrade of curl.


ScottGarman:
:+1:

PR #908, opened 2016-11-18
(PA-696) Update components to tags for 1.8.1 release

Magisus:
This sets all the components to the tags slated for the 1.8.1 release.
All components except puppet are pinned to the versions released in
1.8.0, while puppet is updated to 4.8.1.

### DIFF
--- a/configs/components/facter.json
+++ b/configs/components/facter.json
@@ -1,1 +1,1 @@
-{"url": "git://github.com/puppetlabs/facter.git", "ref": "e8a11caac43c9798912ee8b8f0ca8a0c412d27ce"}
+{"url": "git://github.com/puppetlabs/facter.git", "ref": "refs/tags/3.5.0"}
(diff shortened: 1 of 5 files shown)


Magisus:
Note: once this PR is merged, the release pipeline will most likely try to run again, and it WILL FAIL, because setting some of the components back to previously released tags removes some necessary acceptance test fixes.

MosesMendoza:
👍

geoffnichols:
👍

joshcooper:
I noticed https://github.com/puppetlabs/puppet-agent/blob/1.8.1-release/configs/components/windows_puppet.json isn't tagged, but I don't think that component is used now that MSIs are built by vanagon?

Magisus:
@joshcooper you're correct, it's not currently used, and hasn't been tagged for several releases now (at least).

PR #910, opened 2016-11-30
(PUP-6474) Specify install location for Puppet localedir

Magisus:
This passes a new flag to the puppet install script, which specifies the
correct location for translation files.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -137,6 +137,7 @@ component "puppet" do |pkg, settings, platform|
         --vardir=#{vardir} \
         --rundir=#{piddir} \
         --logdir=#{logdir} \
+        --localedir=#{settings[:datadir]}/locale \
         --configs \
         --quick \
         --no-batch-files \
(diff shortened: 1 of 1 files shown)


configs/components/puppet.rb
```diff
@@ -133,6 +133,7 @@
         --vardir=#{vardir} \
         --rundir=#{piddir} \
         --logdir=#{logdir} \
+        --localedir=#{settings[:datadir]}/locale \
```
geoffnichols: 👍

Magisus:
I have tested this on both a Linux and on Windows, and it appears to behave according to the file paths specified in https://github.com/puppetlabs/puppet-specifications/pull/91.

MosesMendoza:
Added the `don't merge` label on the assumption that this is blocked on https://github.com/puppetlabs/puppet/pull/5366, as that PR adds the support for `--locale-dir` to `install.rb`

MosesMendoza:
However once that is merged I am 👍

Magisus:
This will fix puppet test failures on Windows: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/119/SLAVE_LABEL=beaker,TEST_TARGET=windows2016-64a/

PR #911, opened 2016-12-01
(PUP-6474) Update gettext-setup gem to 0.8

Magisus:
This bumps the gettext-setup gem version to 0.8, which allows the gem
user to specify which file format to expect for translation files.

This will need to go in before https://github.com/puppetlabs/puppet/pull/5366.


branan:
:+1:

PR #912, opened 2016-12-02
(PA-729) Bump leatherman in LTS-1.7 to 0.10.1

mruzicka:
Bump leatherman in LTS-1.7 to the 0.8.2 release which brings in the following fix:
- Handle null characters in JSON strings (LTH-116)

### DIFF
--- a/configs/components/leatherman.json
+++ b/configs/components/leatherman.json
@@ -1,1 +1,1 @@
-{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "0.8.1"}
+{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "refs/tags/0.10.1"}


Magisus:
👍 unless we want to skip this in favor of bumping all branches to 10.1.

MikaelSmith:
Lets update this to use Leatherman 0.10.1.

MikaelSmith:
Updated.

Magisus:
Commit message is updated, but code still says `0.8.2`, at least according to Github...

MikaelSmith:
Oops, fixed. @mruzicka is working on the other changes associated with this as well.

Magisus:
Do those other changes block merging this?

MikaelSmith:
No, this should be the first one to go in. It'll break puppet-agent#stable compilation when we merge LTS-1.7 to stable, which we'll follow up with a fix to pxp-agent#stable.

Magisus:
Sounds good, merging! I'll be ready to merge the others when you have them. Would love to see these pipelines go green over the weekend if we can :)

PR #913, opened 2016-12-02
(PA-729) Bump leatherman in stable to 0.9.4

mruzicka:
Bump leatherman in stable to the 0.9.4 release which brings in the following fixes:
- Handle null characters in JSON strings (LTH-116)
- Explicitly pass release flag to pod2man


Magisus:
👍 unless we want to skip this in favor of bumping all branches to 10.1.

MikaelSmith:
Going to merge up 0.10.1 from LTS-1.7 instead.

PR #915, opened 2016-12-02
(PA-247) Don't halt pxp-agent child processes

MikaelSmith:
Disables halting pxp-agent child processes when restarting or stopping
the service. This makes behavior consistent with all other platforms for
pxp-agent.

Pickup pxp-agent fix from PCP-276 required for this to work.


kurtwall:
This will be great! 👍

Iristyle:
@MikaelSmith could you just add a note to the commit message that as a result of this change, `pxp-agent` now has the possibility of deadlocking and ultimately becoming unresponsive to the Windows SCM?

MikaelSmith:
Updated.

Iristyle:
👍

PR #926, opened 2016-12-14
(INTL-8) Update gettext-setup gem to 0.10

Magisus:
This update adds functionality which allows files to be excluded during
POT generation.


MosesMendoza:
👍

PR #931, opened 2016-12-19
(maint) Make puppet component depend on rubygem-win32-dir

Magisus:
A recent refactor changed the way that Puppet's install script required
the `win32/dir` gem. However, this require was broken during a Vanagon
build, because the script required the `win32/dir` gem before Vanagon
installed that gem. This makes the new dependency explicit.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -4,6 +4,10 @@ component "puppet" do |pkg, settings, platform|
   pkg.build_requires "ruby-#{settings[:ruby_version]}"
   pkg.build_requires "facter"
   pkg.build_requires "hiera"
+  # Used to specify default directories when installing puppet
+  if platform.is_windows?
+    pkg.build_requires "rubygem-win32-dir"
+  end
 
   pkg.replaces 'puppet', '4.0.0'
   pkg.provides 'puppet', '4.0.0'


Iristyle:
Did you by chance have some before / after run output that shows the change?

configs/components/puppet.rb
```diff
@@ -4,6 +4,7 @@
   pkg.build_requires "ruby-#{settings[:ruby_version]}"
   pkg.build_requires "facter"
   pkg.build_requires "hiera"
+  pkg.build_requires "rubygem-win32-dir"
```
Iristyle: On Windows only, right?

McdonaldSeanp: This will need to be confined to windows installs only, As the rubygem-win32-dir gem is only installed for those platforms.

Iristyle: Another thought here is that it feels a little weird to me to require this gem, just so the Puppet installation can continue.

Of course, if we were to inline the equivalent code into `install.rb` to find the 2 paths that we care about we'd probably also need `FFI` (or could potentially use `Fiddle` which is in the box for Ruby, as long as we build / distribute it still).

Maybe we do this now, and file a ticket to clean up the tech debt? (removing `Dir::PROGRAM_FILES`, etc from `install.rb`)

Magisus: In the interest of getting the component pipeline unblocked, I vote for that last suggestion.

Magisus:
I did the build locally, so I don't think it saved a log anywhere. I can do an ad hoc run with this branch though if you'd like.

McdonaldSeanp:
👍

configs/components/puppet.rb
Iristyle: 👍

PR #934, opened 2016-12-22
(PA-786) Update md5sum for rubygem-nokogiri

branan:
Because we do not build nokogiri as part of puppet-agent (only
precompiled on one platform), we missed updating the source md5sum when
we revved to 1.6.8.


branan:
I've targeted this at LTS-1.7 since it's the oldest one with the issue, but I'm happy to point it to stable if we'd rather not bother fixing the LTS (since we don't use this codepath internally)

MosesMendoza:
```
☂ : openssl dgst -md5 nokogiri-1.6.8.gem
MD5(nokogiri-1.6.8.gem)= 51402a536f389bfcef0ff1600b8acff5
```
👍

branan:
I've also confirmed that the nokogiri we have on buildsources is correct, and this was purely a human error in updating the sum

PR #935, opened 2016-12-28
(PA-790) fix arm builds for Debian

mckern:
Puppet Agent 1.8.x cross compiles fail while attempting to build Curl. The root cause is that Curl is picking up the wrong compiler and trying to use the upstream Debian cross-compiler. This winds up including a local x86_64 library path, which means that curl crosslinks against the wrong zlib during configuration and fails. I'm working up a PR to address this, and as a side-effect I may have sussed out some other cross-compilation wonkiness related to the environment variables we're setting.

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -21,7 +21,7 @@ component 'curl' do |pkg, settings, platform|
   end
 
   pkg.configure do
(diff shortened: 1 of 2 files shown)


ScottGarman:
:+1:  lgtm

PR #936, opened 2017-01-04
(maint) Fix targets for puppet-agent

underscorgan:
These targets should match the PE master platforms.


underscorgan:
this *should* only affect promotions into pe, but if this breaks things feel free to revert. I did a build of stable with this change yesterday and it looked normal, but testing out the promotions workflow and things that could go wrong gets really wonky.

McdonaldSeanp:
👍 LGTM

PR #939, opened 2017-01-05
Cherry pick of Fedora 25 platform addition to p-a stable

ScottGarman:


ScottGarman:
Note: Travis failure is just due to the RE ticket number not being public.

MosesMendoza:
@ScottGarman is this needed because these commits originally landed on master but are desired for stable? If so, given our merge strategy, I think the ideal path is to cherry-pick these back to stable and then merge up again stable => master (even though the commits are no-ops). Ideally we aren't merging master to stable.

ScottGarman:
@MosesMendoza yeah, sorry, I was using the wrong terminology. These are cherry picks from master.

MosesMendoza:
Ah, brilliant. thanks @ScottGarman

MosesMendoza:
Do we have a Puppet Agent ticket for Fedora 25 addition? Seems like we might as well, and it would address the travis issue.

ScottGarman:
@MosesMendoza it's not that we're missing a ticket number for the fedora 25 platform addition, it's that the RE ticket referenced in the second commit which adds the platform to build_data.yaml is internal.

MosesMendoza:
👍

ScottGarman:
It turns out we don't want Fedora 25 in the stable CI pipelines, so I'm closing this PR.

PR #941, opened 2017-01-06
(BUILD-17) Add support for Debian 9 i386 and amd64

ScottGarman:
Debian 9.0 is currently in alpha, but these changes can be used to
generate puppet agents for both of the x86 architectures. Additionally,
building this agent no longer requires packages from pl-build-tools,
and uses the distro provided gcc, boost, and libyaml-cpp packages.

### DIFF
--- a/configs/components/cpp-hocon.rb
+++ b/configs/components/cpp-hocon.rb
@@ -5,9 +5,16 @@ component "cpp-hocon" do |pkg, settings, platform|
 
   make = platform[:make]
 
-  # cmake on OSX is provided by brew
-  # a toolchain is not currently required for OSX since we're building with clang.
-  if platform.is_osx?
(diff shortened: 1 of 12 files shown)


ScottGarman:
It blends, and I've smoke tested the agents with simple puppet manifests that exercise the file resource and verified puppet --version works.

ScottGarman:
Here's an ad-hoc agent build showing that this isn't breaking anything as well: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/460/

Magisus:
👍 from me, so long as we're okay with the private ticket number.

stahnma:
LGTM

PR #942, opened 2017-01-09
(RE-8432) Add debian 9 platforms to foss_platforms whitelist

ScottGarman:


ScottGarman:
The travis failure is just due to the ticket being non-public, since this work has to do with Puppet's internal release infrastructure.

Magisus:
👍

PR #945, opened 2017-01-12
(PA-811) Update MCO and Hiera to tags for 1.7.2 release

McdonaldSeanp:
These two components have already been released. So they can be pinned to tags
right now.


glennsarti:
Verified tags look correct.  Good for merge.

PR #948, opened 2017-01-19
(PA-825) Merge 1.7.2-release -> LTS-1.7

McdonaldSeanp:


configs/components/puppet.json
```diff
@@ -1 +1 @@
-{"url": "git://github.com/puppetlabs/puppet.git", "ref": "ddc0c9766564b6cf08fe583a2f7e64a567bdbbb7"}
```
Magisus: I think we want to keep this SHA over the tag, the SHA is newer.

Magisus:
👍

PR #951, opened 2017-01-20
Merge 1.8.3-release to stable

Magisus:
For the components that we pinned back to previous tags (hiera, mcollective), I had to go in and edit the merge to prefer the SHAs in stable over the tags in 1.8.3-release. For Facter, I preferred the updated SHA which didn't make it into the release. Puppet and pxp-agent were tagged at their most recent stable SHA.


MosesMendoza:
:thu

PR #952, opened 2017-01-20
(PA-890) Revert conflicts on (pe-)puppetserver

MosesMendoza:
(PA-890)
We have decided to delay/rethink these package conflicts to avoid unanticipated
breakage in a minor version and consequences thereof. In the interim, we will
vendor libs in question, per PA-890 and PUP-7114.


McdonaldSeanp:
👍

PR #958, opened 2017-01-26
Bump component SHAs for hiera and puppet

ScottGarman:
These are needed to ensure we get the hiera version bump (kinda important!) and the puppet fix for PUP-6925.


geoffnichols:
Verified the puppet and hiera SHAs match head of puppet#master and hiera#master and were the last ones successfully promoted into puppet-agent#master. 👍

PR #971, opened 2017-01-31
(PA-855) Use component tags for 1.9.0 release

ScottGarman:


Magisus:
Looks good, those tags are present in all component repos, and they match the stated versions on Confluence for 1.9.0.

McdonaldSeanp:
👍

PR #972, opened 2017-02-01
(PA-625) Remove puppet_for_the_win configs, scripts, and docs

Magisus:
Now that puppet-agent is fully vanagonized, we no longer need special
tooling for Windows. This commit deletes all references to the unused
puppet_for_the_win build scripts and component configs.

The only other references seem to be in the [packaging repo README](https://github.com/puppetlabs/packaging/blob/master/README.md#setting-up-projects-for-the-packaging-repo), in a section that looks very out-of-date in other respects.


configs/components/windows_puppet.json
```diff
@@ -1,4 +0,0 @@
-{
-  "url": "git://github.delivery.puppetlabs.net/puppetlabs-puppet_for_the_win.git",
-  "ref": "origin/master"
-}
```
joshcooper: Can we delete [`windows_ruby.json`](https://github.com/puppetlabs/puppet-agent/blob/2.0.x/configs/components/windows_ruby.json) also and references to it?

Iristyle: Yes, 👍  - that should also be removed.

But not only that.  I think everything from `bin` can be removed as well, except for `install-chocoately.ps1` which is being used in the config definitions.

So remove:

* build-cpp-pcp-client.ps1
* build-facter.ps1
* build-helpers.ps1
* build-leatherman.ps1
* build-pxp-agent.ps1
* build-windows.rb
* windows-env.ps1
* windows-toolset.ps1

Magisus: Updated

PR #984, opened 2017-02-10
(maint) Remove foss_platforms from build_defaults

mwaggett:
This commit removes the list of foss platforms from build_defaults.yaml, since this gets overwritten by the list in the build-data repo.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -2,42 +2,6 @@
 project: 'puppet-agent'
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
-# foss_platforms will be shipped to the nightly repos
-foss_platforms:
-  - cumulus-2.2-amd64
-  - debian-7-amd64
-  - debian-7-i386
-  - debian-8-amd64
-  - debian-8-i386
-  - el-5-i386
-  - el-5-x86_64
-  - el-6-i386
-  - el-6-x86_64
-  - el-7-x86_64
-  - eos-4-i386
-  - fedora-f23-i386
-  - fedora-f23-x86_64
-  - fedora-f24-i386
-  - fedora-f24-x86_64
-  - fedora-f25-i386
-  - fedora-f25-x86_64
-  - huaweios-6-powerpc
-  - osx-10.10-x86_64
-  - osx-10.11-x86_64
-  - osx-10.12-x86_64
-  - sles-11-i386
-  - sles-11-x86_64
-  - sles-12-x86_64
-  - ubuntu-12.04-amd64
-  - ubuntu-12.04-i386
-  - ubuntu-14.04-amd64
-  - ubuntu-14.04-i386
-  - ubuntu-16.04-amd64
-  - ubuntu-16.04-i386
-  - ubuntu-16.10-amd64
-  - ubuntu-16.10-i386
-  - windows-2012-x86
-  - windows-2012-x64
 pe_platforms:
   - aix-5.3-power
   - aix-6.1-power


melissa:
@underscorgan especially, we need to remove this duplication. We have another list of foss_platforms that lives in the build-data repo (https://github.com/puppetlabs/build-data/blob/release/builder_data.yaml#L63-L99) which overwrites this one. As a result, when 16.10 was added here but not to the build-data repo, 16.10 wasn't shipped out during the initial PA 1.9.1 ship. We had to go back and ship it after we realized it was missing.

HOWEVER, I don't know if there might be other reasons this data needs to persist in this repo. So, that is why @mwaggett opened this PR, so we could get some discussion.

cc/ @McdonaldSeanp @ScottGarman @geoffnichols

mwaggett:
Thanks for adding the labels and context, @melissa 😄 
Forgot those things during lunch. 😅

melissa:
totes :-D

MosesMendoza:
but very much agree with this pr

underscorgan:
@melissa @mwaggett the duplication is my fault, was added when dealing with `FOSS_ONLY` mode. I think this is mostly used in this repo for nightlies so my concern would be accidentally shipping extra platforms for that.

melissa:
This list should be exactly the same as the one in build-data. Do we not load build data when shipping to nightlies?

underscorgan:
I don't know how the nightlies shipping works ¯\_(ツ)_/¯

mwaggett:
@underscorgan do you know where we can find out how that works?

melissa:
Yeah, it's all in the packaging repo. The best place to start will be to look at the nightly jobs to see what rake task they call

mwaggett:
@melissa @underscorgan It looks like the nightly repo tasks (https://github.com/puppetlabs/packaging/blob/master/tasks/nightly_repos.rake) all call pl:fetch, which is where build-data comes in, as far as I can tell: https://github.com/puppetlabs/packaging/blob/master/tasks/fetch.rake#L46

so does this PR still feel like a safe & sane idea?

McdonaldSeanp:
So, I am +1 to moving this information to a centralized place. *However*, that centralized place will need to support different versions of OS lists. 

Now that we have two streams: LTS and master, it is possible that the set of OSes to ship will be different between the two. I am 👍 to this PR if build-data is set up to handle different sets of OSes for the two streams, but otherwise we need to have further discussion.

melissa:
If we need to support different sets of platforms in different branches, then we should remove this from build-data. @underscorgan why did we add it there again?

mwaggett:
Another option would be to have separate branches in build-data.. We'd have to adjust packaging to deal with that, but aren't we planning on killing or at least super-refactoring packaging anyway?

underscorgan:
@melissa we added it to build-data to get foss_only mode working / a consistent set of platforms for shipping.

underscorgan:
👎 for adding moar branches in build data. we need to be moving things out of there not adding new things.

mwaggett:
@underscorgan Why moving things out? If that's the case, then why don't we use the list in this repo instead of overriding with build-data?

underscorgan:
@mwaggett because we wanted the list for things that aren't puppet-agent.

underscorgan:
@McdonaldSeanp are the actual platforms built derived from these lists? I thought they were intended to be the whitelist for shipping nightlies?

underscorgan:
The list in build-data is definitely meant as a whitelist not an authoritative list.

McdonaldSeanp:
@underscorgan ah, I believe I misunderstood what these were meant for.

If these are meant as whitelists, then putting them in build-data would make sense, and there are no problems with the two different streams.

mwaggett:
so where do we stand with this?

McdonaldSeanp:
I am +1 after the explanation from @underscorgan, but this PR should be targeted at the LTS branch of puppet-agent so we can easily do the appropriate mergeups.

PR #988, opened 2017-02-17
(PA-954) Pin Hiera, MCO, pxp-agent to tags

McdonaldSeanp:
For the upcoming 1.9.2 release


jcoconnor:
👍

Magisus:
Note: some platforms will fail on Facter tests with this tag due to acceptance changes that landed after the previous tag. With that caveat, 👍

McdonaldSeanp:
Facter has been moved back to a SHA so acceptance will pass. It is very low risk to ship the latest facter changes as a Z release and we will benefit from all green CI.

Magisus:
Sounds good to me 👍

PR #989, opened 2017-02-17
Revert "(PA-954) Pin Hiera, MCO, pxp-agent to tags"

McdonaldSeanp:
Reverts puppetlabs/puppet-agent#988

We are going to test with SHAs because there are acceptance changes.


ScottGarman:
:+1:

McdonaldSeanp:
Failure is just a commit check

Magisus:
we should fix that, looks like it's diffing against the wrong thing.

PR #993, opened 2017-02-23
(PA-768) Make puppetres.dll permanent

jcoconnor:
puppetres.dll is managed and installed using the default file harvest
function, so make a separate puppetres.wxs file to define this as a
permanent component along with its registry entry.

The puppetres.dll file also needs to be filtered out of the harvest
operation. This is achieved by using the "install_service" method in
puppet.rb which adds it to the filter xslt file.

A future implemntation of vanagon could add a "filter_file" method which
would better describe this use of the function.

### DIFF
--- a/resources/windows/wix/filter.xslt.erb
+++ b/resources/windows/wix/filter.xslt.erb
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsl:stylesheet version="1.0"
+ xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+ xmlns:wix="http://schemas.microsoft.com/wix/2006/wi">
+  <!-- https://ahordijk.wordpress.com/2013/03/26/automatically-add-references-and-content-to-the-wix-installer/ -->
+  <!-- http://www.chriskonces.com/using-xslt-with-heat-exe-wix-to-create-windows-service-installs/ -->
+  <xsl:output method="xml" indent="yes" />
+  <!--<xsl:strip-space elements="*"/>-->
+  <xsl:template match="@*|node()">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+  <!-- Filter out all Service File Executables from the Harvest (heat) run as these are specified in the transformed service.components.wxs file -->
+  <!-- The list of component service files are collected in an array so that a set of unique names can be extracted -->
+  <!-- we have to substitue the SourceDir\\etc. etc. for $(var.ProjectSourcePath) in order to facilitate heat -->
+  <!--   being run in two different places, specifically it no longer runs from SourceDir so we needed to pass -->
+  <!--   in ProjectSourcePath to the heat runs. Thus when this filter does its work the source attribute will -->
+  <!--   look like $(var.ProjectSourcePath)/puppet/bin/rubyw.exe for example, not  -->
+  <!--   SourceDir/Program64FilesFolder/Puppet/puppet/bin/ruby.exe -->
+  <%- service_files = Array.new -%>
+  <%- get_services.each do |service| -%>
+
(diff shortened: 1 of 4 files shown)


jcoconnor:
@Iristyle & @McdonaldSeanp This PR updates PA install to retain ```puppetres.dll``` after PA removal.

I have done basic testing and the behaviour appears to be correct.

Unfortunately I was unable to use the original service_file method that I discussed briefly with Sean yesterday, so instead have made local copy of the ```filter.xslt.erb``` to filter out ```puppetres.dll``` from the harvest operation.

Iristyle:
Also /cc @jpogran

resources/windows/wix/filter.xslt.erb
```diff
@@ -0,0 +1,34 @@
...
+    </xsl:copy>
+  </xsl:template>
+  <!-- Filter out all Service File Executables from the Harvest (heat) run as these are specified in the transformed service.components.wxs file -->
+  <!-- The list of component service files are collected in an array so that a set of unique names can be extracted -->
+  <!-- we have to substitue the SourceDir\\etc. etc. for $(var.ProjectSourcePath) in order to facilitate heat -->
+  <!--   being run in two different places, specifically it no longer runs from SourceDir so we needed to pass -->
+  <!--   in ProjectSourcePath to the heat runs. Thus when this filter does its work the source attribute will -->
+  <!--   look like $(var.ProjectSourcePath)/puppet/bin/rubyw.exe for example, not  -->
+  <!--   SourceDir/Program64FilesFolder/Puppet/puppet/bin/ruby.exe -->
+  <%- service_files = Array.new -%>
```
Iristyle: I'm a little confused by these bits because this seems quite complex.

Will have to dig in a bit more to try to understand better.

McdonaldSeanp: So, this is part of how we needed to implement WiX with vanagon. Normally, this functionality (filtering out files from a heat run) would exist in vanagon. However, the implementation we use in vanagon will *only* filter files listed as services in the component/project configuration (see [here](https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/puppet.rb#L48) for an example of something listed as a service file by vanagon)

All the following lines do is add things listed as service files to an XML filter that we feed to heat.exe when it's autogenerating all the WiX for the project.

As I said before, normally this functionality is located in vanagon, but since we are adding the new puppetres.dll file to the filter (and we cant list it as a service because that breaks things) we needed to copy over this functionality to the agent so we could add things to the filter.

We would like to go back to using the file in vanagon and remove this in the future, but https://tickets.puppetlabs.com/browse/VANAGON-46 needs to be resolved before we can do so

McdonaldSeanp:
👍 LGTM

resources/windows/wix/filter.xslt.erb
Iristyle: Ok - I see what's happening... this is unfortunate, but I understand what's going on.

A couple of other ideas here:

* Since `filter.xslt.erb` is already an ERB template, allow injecting an arbitrary string into this document through Vanagon OR
*  See if it's possible to pass additional XSLT transforms to heat.exe via `-t` in addition to the Vanagon default.

resources/windows/wix/puppetres.wxs
```diff
@@ -0,0 +1,52 @@
...
+            Type="string"
+            Name="EventMessageFile"
+            Value="[INSTALLDIR]puppet\bin\puppetres.dll"/>
+        </RegistryKey>
+      </Component>
+
+      <Component
+          Id="PuppetResDllFileComp"
+          Directory="PUPPETRESPUPPETBINDIR"
+          Guid="A8032753-9CC7-45DF-815A-EBF17D48BE3F"
```
Iristyle: @glennsarti has pointed out that should we ever change `puppetres.dll` to add additional messages, compile under 64, etc.... that we would be required to change the `Guid`.

In #1002, by relying on the heat generated `Guid`, we get a more maintainable behavior I think.

jcoconnor:
@Iristyle I'm closing the PR in favour of your one.

PR #1000, opened 2017-03-04
Vanagon 0.10.x is a breaking change but help is here

mckern:
5ca64d21b4057551c594c8e69b942efb

The new environment variable support in Vanagon is great! Except that env. var. scoping and the component env. var. interface has changed. This means that the change is a backwards incompatible change which will require updating numerous component configurations to take advantage of it. Of note:

- CMake projects seem to build correctly now if we use the Unix Makefile generator. This means we don't need to mangle `$PATH` quite as often or quite as much.
- All env. vars. (project, platform, or component) are now scoped for resolution by Make.
- I tried to cut down on calling `$(cygpath -u <some windows path>)` quite so much.

There's still room to simplify the P-A config but this is a good start. Here's the platforms I've built using this branch so far:

- [x] `el-5-i386`
- [x] `windows-2012r2-x64`
- [ ] `aix-7.1-ppc`


mckern:
This work is blocked while waiting for https://github.com/puppetlabs/vanagon/pull/459 (and an AIX build) but please review for obvious WTFery.

McdonaldSeanp:
@mckern can you test a build on windows x86 as well? Normally that's never an issue, but ruby changes have historically been SUPER flaky.

mckern:
Huh. This is surprising. AIX is using GNU Make 3.80 (released in 2002) but we decided to target GNU Make 3.81 (released in 2006) features as our baseline; that's (at least) what many (most?) platforms ship out of the box. Gonna track that over in Vanagon PR #459 but it'll impact this as we hammer it out.

geoffnichols:
Not sure if this helps, but looks like IBM may provide a GNU Make 4.1 rpm via their ["AIX Toolbox for Linux Applications"](http://www-03.ibm.com/systems/power/software/aix/linux/toolbox/alpha.html#M)

PR #1010, opened 2017-03-13
(PCP-719) Merge fix to use TLS v1.2 in cpp-pcp-client

MikaelSmith:


McdonaldSeanp:
👍

branan:
I really don't like seeing manually-updated SHAs. We either should be on released tags for a component (like leatherman), or have continuous SHA updates from Jenkins.

What's the plan to ensure we're keeping cpp-pcp-client at a reasonable commit at all times?

MikaelSmith:
Currently manual. I'm happy to tag cpp-pcp-client once we're settled on a release. This is a practice I've used for Leatherman before too (bump SHA, tag and bump to tag after release contents are settled). I'm not against a full continuous update pipeline for Leatherman or cpp-pcp-client either however.

PR #1014, opened 2017-03-24
(PA-437) Add wrapper scripts to clean up linker environment

branan:
Previously, we relied on symlinks from our public bin directory to the
internal one to indicate which executables were public. This worked for
its purpose, but could cause a user's linker environment variables to
affect our executables. Because we rely on our own versions of many
libraries, this can cause hard-to-debug failures for our customers.

We now use a small wrapper script that cleans up the linker environment
variables, instead of a symlink. This maintains our clean separation
between public and internal executables, and makes sure that users
global settings won't mess with Puppet.

Users that want to adjust Puppet's behaviors with linker variables can
directly call the internal path to the executables. This does not
require us to stabilize those internal locations, as any user
intentionally messing with our loaded libraries has already broken out
of being on a supported configuration.


branan:
I'm currently confirming that this builds and works as expected on {linux,solaris,aix}. Will remove the DNM tag when that validation is complete

McdonaldSeanp:
IMO it makes more sense to use erb to populate the environment changes and only use one script resource. I certainly wouldn't block on that if others disagree, but that's my 2 cents

Magisus:
This seems reasonable. It's simple enough that I can understand it even with little context, and shouldn't mess with the rest of their system.

Might consider a better name for the script itself though. The comment in the component makes its purpose clear, but a better name for the component would make it easier to understand from the outside (i.e. in the project).

branan:
@McdonaldSeanp do you have an example of a sane pattern for doing file generation in Vanagon? I couldn't find anything that didn't feel worse than this to me

branan:
@Magisus Yeah, naming is hard. I'm open to suggestions, but nothing comes to mind immediately

McdonaldSeanp:
@branan turns out we don't expose the vanagon's erb functionality externally, so we can't do that. Soooooo, nevermind.

Magisus:
Were you intending this to be super generalizable (i.e. a component that could point to a script that does _anything_)? If not, what about something as simple as "env-clean" or somesuch?

branan:
It looks like `mode` on `install_file` may be busted in vanagon. AFAICT we don't use this capability anywhere else in puppet-agent, so it's possible it never worked. This is blocked on fixing that in Vanagon

branan:
It looks like the `install_file` `mode` setting does work, but only in the final package, not on the system at build time. this is probably still kinda buggy, but not a blocker to this PR. After installing a generate RPM, everything seems to be working as expected:

'''
[root@xri3tc6otlmi6wb /]# for e in puppet facter hiera mco; do /opt/puppetlabs/bin/$e --version; done
5.0.0
4.0.0 (commit fdfec64eedc5031eb674c072ab850a636e97d813)
3.3.1
/opt/puppetlabs/puppet/bin/mco 2.10.2
[root@xri3tc6otlmi6wb /]# for e in puppet facter hiera mco; do /opt/puppetlabs/bin/$e --version; cat /opt/puppetlabs/bin/$e; done
5.0.0
#!/bin/sh

unset LD_LIBRARY_PATH
unset LD_PRELOAD

COMMAND=`basename "${0}"`
BIN_PATH=`dirname "${0}"`
INSTALL_PATH=`dirname "${BIN_PATH}"`

${INSTALL_PATH}/puppet/bin/${COMMAND} $@
4.0.0 (commit fdfec64eedc5031eb674c072ab850a636e97d813)
#!/bin/sh

unset LD_LIBRARY_PATH
unset LD_PRELOAD

COMMAND=`basename "${0}"`
BIN_PATH=`dirname "${0}"`
INSTALL_PATH=`dirname "${BIN_PATH}"`

${INSTALL_PATH}/puppet/bin/${COMMAND} $@
3.3.1
#!/bin/sh

unset LD_LIBRARY_PATH
unset LD_PRELOAD

COMMAND=`basename "${0}"`
BIN_PATH=`dirname "${0}"`
INSTALL_PATH=`dirname "${BIN_PATH}"`

${INSTALL_PATH}/puppet/bin/${COMMAND} $@
/opt/puppetlabs/puppet/bin/mco 2.10.2
#!/bin/sh

unset LD_LIBRARY_PATH
unset LD_PRELOAD

COMMAND=`basename "${0}"`
BIN_PATH=`dirname "${0}"`
INSTALL_PATH=`dirname "${BIN_PATH}"`

${INSTALL_PATH}/puppet/bin/${COMMAND} $@
'''

branan:
I'm running a broader ad-hoc pipeline now, and once it's green I'll call this ready to go

branan:
@Magisus My concern with a name like "env-clean" is that it sounds like it's part of the build environment, not the runtime. "Wrapper" implies (to me) something that we're laying down on the file system. "environment-cleaning-executable-wrapper" would be the most descriptive name, but feels too long. :man_shrugging:

Magisus:
Yeah I think you're right. Don't we use something like this on Windows? I think we call that one just "environment", which I don't really like, but maybe "environment-wrapper"? The only issue I have with _just_ "wrapper" is it feels super generic even though it's not supposed to be.

McdonaldSeanp:
I'm fine with the names. The windows .bat files have different names for each component that uses a binary, i.e. puppet.bat, facter.bat and environment.bat is another bat file called by each component file to mangle the environment, which is an approach I'm not a super fan of.

I would say I overall agree with the entire approach.

👍

branan:
It'd be nice to eventually refactor the various `.bat` files to be part of this as well, but I don't think that's worth doing until we also start fixiing some of the other windows idiosyncracies

MosesMendoza:
Is there any chance any of these projects will be running in a supported configuration where the invoking user/process is not allowed to also invoke `#!/bin/sh`?

MosesMendoza:
@branan would you mind sending a note to puppet-dev about this change as a heads up / asking for feedback?

MosesMendoza:
Is there any way to run this in an ad-hoc manner against PE smoke tests too, ie via frankenbuilder or something?

resources/files/sysv-wrapper.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+unset LD_LIBRARY_PATH
+unset LD_PRELOAD
```
Sharpie: Should we also unset DYLD_LIBRARY_PATH for the macOS folks?

branan: Ah, yup. That's another special case like AIX that I didn't even think of. I'll do some research on the OSX linker variables

branan:
@puppetlabs/agent-team  This is updated with an acceptance test, and to add Darwin support.

resources/files/sysv-wrapper.sh
branan: This should be all set with the latest revision

acceptance/tests/linker_variable_cleanup.rb
```diff
@@ -0,0 +1,115 @@
...
+            else
+              "LD_PRELOAD"
+            end
+      on agent, puppet("--version"), :environment => {var => libs[agent] }, :acceptable_exit_codes => [42]
+    end
+  end
+
+  step "Put the crashing library on the load path" do
+    agents.each do |agent|
+      next if dirs[agent] == nil
```
MosesMendoza: Falling through all the way to success on an unsupported platform opens up the possibility that we might stop or not be testing the platforms we think we are. I think it would be safer if we explicitly target only the platforms we want to test, always run on those platforms, and fail down the road if for some reason we won't have the right fixtures/pathing etc.

Magisus:
@MosesMendoza I wonder if we should look into adding a PEZ step to the ad-hoc pipeline. I don't know if that's feasible or not, but it seems like it'd be nice to have. /cc @McdonaldSeanp

acceptance/tests/linker_variable_cleanup.rb
branan: @MosesMendoza I don't think we have an easy way to confine to all the linuxes in Beaker. I'd rather just fall through here than miss updating the confine as we add new linux platforms.

Our non-linux platforms are all special enough that they'd require manual effort to port this test to them. Having this fall-through vs. be skipped by a confine in that case seems like a toss-up to me.

branan:
@Magisus This is updated to now only do the library path test, and not the preload test. That should avoid the issue we were seeing where my crasher library was crashing Bash. It does mean the test coverage isn't as good, but it will at least ensure that the wrapper scripts are being laid down sanely.

resources/files/sysv-wrapper.sh
MosesMendoza: It doesn't look we have a case for `darwin` in the acceptance test - can you add one or was that prohibitively costly? Seems like that platform seems to edge-case us all the time..

acceptance/tests/linker_variable_cleanup.rb
MosesMendoza: Hmm - alright. Could we instead then confine to !$platform? That's the pattern we follow in acceptance for unsupported platforms anyway. Looks like we're already confining to !windows, could we also do !Cisco & !Huawei?

resources/files/sysv-wrapper.sh
branan: It's prohibitive. I couldn't get it working in the LIBRARY_PATH test, and I ripped out the preload test entirely because it's way too complex to do effectively.

acceptance/tests/linker_variable_cleanup.rb
branan: I'd actually prefer to run on Cisco and Huawei - both of them are "weird" Linuxes, so getting coverage on them is good.

Confining against windows really says "This feature isn't expected to be implemented on that platform". Everything else here is just missing coverage within the test.

Magisus:
The tests seem to pass now 👍

ScottGarman:
:+1:  I'm going to merge this per our discussion in standup; further changes can always be made after the fact, and this is now known to pass testing.

PR #1017, opened 2017-04-01
Adjusting openssl configuration flags to disable following ciphers: D…

jsane:
Security enhancement: Updated flags to openssl Configure to disable following ciphers - DES/3DES, RC2/4/5, CAMELLIA, IDEA, SEED, GOST, CAST, MD2/4, BF.
The above list, of ciphers to be disabled, come from the SSL Implementation Reference located here (Recommended Ciphersuites section): https://docs.google.com/document/d/1KonBlqNY9pGY01PfjidPk3PAgKJG85huoOQW_LQ_tME/edit#heading=h.t17jjgd88ew7.

**Note**: With these switches ECDHE is getting enabled. Pl let know if that needs to be disabled as well though I would wonder why it need to be.

This was tested using nogotofail to ensure none of the above ciphers were presented by puppet-agent when establishing TLS connection. Expect a PR on nogotofail to enable checks for unacceptable ciphers and then on pe_acceptance_tests to have it use the puppetlabs/nogotofail. 

@joshcooper

### DIFF
--- a/configs/components/openssl.rb
+++ b/configs/components/openssl.rb
@@ -118,6 +118,7 @@ component "openssl" do |pkg, settings, platform|
     # --libdir ensures that we avoid the multilib (lib/ vs. lib64/) problem,
     # since configure uses the existence of a lib64 directory to determine
     # if it should install its own libs into a multilib dir. Yay OpenSSL!
+    # Disabled: rc4, bf, 
     "./Configure \
       --prefix=#{settings[:prefix]} \
       --libdir=lib \
@@ -127,19 +128,22 @@ component "openssl" do |pkg, settings, platform|
       #{target} \
       #{sslflags} \
       no-camellia \
-      enable-seed \
       enable-tlsext \
       enable-rfc3779 \
-      enable-cms \
       no-md2 \
       no-mdc2 \
-      no-rc5 \
+      no-weak-ssl-ciphers \
+      no-dtls \
+      no-dtls1 \
+      no-rc4 \
+      -DOPENSSL_NO_RC4 \
+      no-idea \
+      no-seed \
       no-ec2m \
-      no-gost \
-      no-srp \
       no-ssl2 \
       no-ssl2-method \
       no-ssl3 \
+      -DOPENSSL_NO_HEARTBEATS \
       #{cflags} \
       #{ldflags}"]
   end


configs/components/openssl.rb
```diff
@@ -127,19 +128,22 @@
       #{target} \
       #{sslflags} \
       no-camellia \
-      enable-seed \
       enable-tlsext \
       enable-rfc3779 \
-      enable-cms \
       no-md2 \
       no-mdc2 \
-      no-rc5 \
```
joshcooper: Should we preserve `no-rc5` or is it in the `no-weak-ssl-ciphers` bucket?

configs/components/openssl.rb
```diff
@@ -127,19 +128,22 @@
...
+      no-weak-ssl-ciphers \
+      no-dtls \
+      no-dtls1 \
+      no-rc4 \
+      -DOPENSSL_NO_RC4 \
+      no-idea \
+      no-seed \
       no-ec2m \
-      no-gost \
-      no-srp \
```
joshcooper: Should we continue to exclude these?

joshcooper:
I think it's okay to leave ECDHE enabled.

It would be great to add an acceptance test to verify the end result. It could be as simple as running `openssl cipher -v` and verify the result contains the expected cipher list.

configs/components/openssl.rb
jsane: Pardon the delay in getting back. I did/do not have notifications enabled in github.. 
Yes weak-ssl-ciphers seems to disables it. Strangely enough it would not disable RC4 IIRC even though the documentation says it would be so need to add no-rc4 separately.

configs/components/openssl.rb
jsane: These too seem to be getting disabled by no-weak-ssl-ciphers. In my testing these wont show up and our "SSL Implementation Reference" document recommends disabling these. Unless you are saying there are some developments/research suggesting these can be used.

jsane:
@joshcooper, regarding your above comment on adding a test to verify I have re-introduced checks in nogotofail to check the client (puppet-agent) only supports ciphers from the acceptable list (have included ECDHE in there) and that it is also not missing any ciphers from the acceptable list (flag as a warning).  I have a PR on it. Pl let know if that is sufficient or more is needed. 

Also not sure what the above error on "All checks have failed ... CI Travis-CI build failed". Does it mean the builds will start failing if this gets merged?

joshcooper:
@jsane If you click on the Travis CI build details, it shows that there's a rubocop issue with trailing whitespace (static code analysis) and the commit summary needs to be prefixed with the ticket number, e.g. "(PA-XXX) Disable weak ciphers". This latter part is necessary because we have automation to reconcile tickets vs git commits, to ensure we not including commits that don't have tickets, or ticket that are supposed to be fixed in the release, but don't have any commits.

joshcooper:
Superseded by PR #1024

PR #1018, opened 2017-04-03
(PA-762) Fix locale gem version

Magisus:
Previously the locale gem component referenced a gem version that does
not exist outside of our internal mirrors. This updates the gem version
to 2.1.2, which is present on rubygems.org.


McdonaldSeanp:
👍

ScottGarman:
:+1:

PR #1021, opened 2017-04-04
Ubuntu Power8 platform addition (ubuntu-16.04-ppc64el)

ScottGarman:
This builds an agent which can successfully run some smoke tests on the Power8 target I have access to, including using the file resource type and basic commands such as puppet --version and facter.


McdonaldSeanp:
👍

configs/components/openssl.rb
```diff
@@ -13,7 +13,7 @@
     # needed for the makedepend command
     pkg.build_requires 'imake' if platform.name =~ /^el/
     pkg.build_requires 'xorg-x11-util-devel' if platform.name =~ /^sles/
-    pkg.build_requires 'xutils-dev' if platform.is_huaweios?
+    pkg.build_requires 'xutils-dev' if platform.is_deb?
```
stahnma: Does this change impact armel and armhf debian builds negatively?

ScottGarman: I'm actually not sure how openssl could have built without the makedepend command, which is in the xutils-dev package for all Debian-based distros. I'll run a test build for each of those architectures just to make sure nothing's been broken.

kurtwall: Looks sane to me. Deferring the merge pending feedback on ^^

ScottGarman: @stahnma looks like these arm platforms haven't been built since we made the default switchover to ruby 2.3.3 - the build blows up immediately due to not being able to find the rbconfig file for the platform. That's something else that needs to be fixed. 

Both of these arm platforms are debian-8 only, so I have verified that our debian 8 amd64 pooler hosts do not have the makedepend command by default and installing xutils-dev will get you that command. IMO the worst case scenario would be that this would just pull in an extra package that might not be used.

ScottGarman:
Note: I've filed PA-1070 to update the debian-8-armel platform. I didn't have to create a ticket for debian-8-armhf, as it appears to be building without issues.

PR #1022, opened 2017-04-05
(PA-1044) Update all component SHAs to tags

branan:
This updates cpp-pcp-client, facter, leatherman,
marionette-collective, puppet, and pxp-agent to their tagged versions
for the agent 1.10.1 release.


stahnma:
All of those shas match the tags on checkout, so good work :)

PR #1024, opened 2017-04-05
(PA-1059) Adjusting openssl configuration flags to disable weak ciphers

jsane:
Moved the ticket to PA. Corresponds to this earlier PR https://github.com/puppetlabs/puppet-agent/pull/1017 (which had the ticket in Security project).

### DIFF
--- a/configs/components/openssl.rb
+++ b/configs/components/openssl.rb
@@ -118,6 +118,7 @@ component "openssl" do |pkg, settings, platform|
     # --libdir ensures that we avoid the multilib (lib/ vs. lib64/) problem,
     # since configure uses the existence of a lib64 directory to determine
     # if it should install its own libs into a multilib dir. Yay OpenSSL!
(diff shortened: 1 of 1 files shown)


joshcooper:
@jsane travis is complaining about trailing whitespace in the one of the commits. You can run `git show --check` on the commit and it should show you where the trailing whitespace is.

jsane:
Sorry for that and delay again (I seem to get github notifications intermittently as this did not show up till today am). Anyway right now my machine seems to have gotten in weird state since after a reboot yesterday night. It wont let me access my github keys, and now 1Password.  Will rectify as soon as I have that sorted out.

jsane:
Yes I tested the agent only sends the desired list of ciphers (at least per what we adjusted in openssl.rb) using nogotofail independently. There are PRs on nogotofail to detect/check for weak ciphers along with this so yes they would get checked regularly as part of pe_acceptance_tests.

PR #1025, opened 2017-04-06
(PA-1019) Vendor minitar on solaris

branan:
Previously, we only vendored minitar on Windows, which completely lacks a `tar` command. As it turns out, the version of `tar` on Solaris doesn't support all the options that the PMT wants to use, so we also need minitar there.


Magisus:
👍

ScottGarman:
:+1:

PR #1027, opened 2017-04-07
(packaging) Add windows to `platform_repos`

mwaggett:
Beaker will be using the list of `platform_repos` for testing, so this commit adds windows so that we have a complete list of platforms.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -141,6 +141,8 @@ platform_repos:
     repo_location: repos/solaris/11/**/*.i386.p5p
   - name: solaris-11-sparc
     repo_location: repos/solaris/11/**/*.sparc.p5p
+  - name: windows-2012-x86
+  - name: windows-2012-x64
 gpg_name: 'info@puppetlabs.com'
 gpg_key: '7F438280EF8D349F'
 deb_targets: 'precise-amd64 trusty-amd64 xenial-amd64'


mwaggett:
The fact that these entries are missing `repo_location` keys is handled here: https://github.com/puppetlabs/packaging/pull/634

ext/build_defaults.yaml
```diff
@@ -141,6 +141,8 @@ platform_repos:
     repo_location: repos/solaris/11/**/*.i386.p5p
   - name: solaris-11-sparc
     repo_location: repos/solaris/11/**/*.sparc.p5p
+  - name: windows-2012-x86
+  - name: windows-2012-x64
```
McdonaldSeanp: So, while we do only build for these two platforms (and we only ship two packages for windows, an x86 and an x64 version) we test those two packages on several different platforms.

In general, you can't use the platform_repos list as the list of platforms to test, as they are different

McdonaldSeanp:
So, I'm fine with people disagreeing with this and I welcome other opinions, but I think this is the wrong approach to abstracting this information out of beaker. 

There's a couple of changes that happen in this kind of data during the transition between building and subsequently testing packages on different operating systems. Paramount of which is the problem I pointed out above: On more than one platform, we build only one package that we test on multiple versions of an OS.

What concerns me is the supporting PR in packaging has allowed platforms with missing information not to fail, and if we only did that to support this PR I believe we are on the wrong path.

mwaggett:
After Sean's feedback, this is no longer the direction we're going.

PR #1028, opened 2017-04-07
(PA-1023) Add step to generate MO files for puppet-agent components

Magisus:
This PR contains 3 commits enabling i18n in the major puppet-agent components.
The first affects puppet, the second enables Leatherman's i18n, and the third updates the configuration of the other C++ project to use Leatherman's i18n tooling.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -8,6 +8,17 @@ component "puppet" do |pkg, settings, platform|
   if platform.is_windows?
     pkg.build_requires "rubygem-win32-dir"
   end
+  # Used to compile binary translation files
+  # i18n is not supported on Solaris
+  if platform.is_osx?
+    pkg.build_requires "gettext"
+  elsif platform.is_windows?
+    pkg.build_requires "pl-gettext-#{platform.architecture}"
+  elsif platform.is_aix?
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-gettext-0.19.8-2.aix#{platform.os_version}.ppc.rpm"
+  elsif !platform.is_solaris?
+    pkg.build_requires "pl-gettext"
+  end
 
   pkg.replaces 'puppet', '4.0.0'
   pkg.provides 'puppet', '4.0.0'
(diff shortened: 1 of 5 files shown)


Magisus:
This can't be merged till we decide what to do about gettext on EL5.

Magisus:
We also need to regenerate the AIX packages using a version of vanagon that includes https://github.com/puppetlabs/vanagon/pull/475,

ScottGarman:
@Magisus yeah, I can help with rebuilding/shipping packages when the vanagon work is done. Ping me when that's the case.

configs/components/puppet.rb
```diff
@@ -8,6 +8,16 @@
   if platform.is_windows?
     pkg.build_requires "rubygem-win32-dir"
   end
+  # Used to compile binary translation files
+  if platform.is_osx?
+    pkg.build_requires "gettext"
+  elsif platform.is_windows?
+    pkg.build_requires "pl-gettext-#{platform.architecture}"
+  elsif platform.is_aix?
+    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-gettext-0.19.8-2.aix#{platform.os_version}.ppc.rpm"
```
kurtwall: This is cool. We don't officially support it on AIX, so if i18n suddenly breaks, does this commit commit us (heh, see what I did there?) to unofficially supporting i18n on AIX?

Magisus: yeah honestly I'm indifferent on this one. I don't know enough about WHY we don't support i18n on AIX (though I believe it has something to do with not being able to detect the locale? /cc @branan) to say how likely it is that we will expand support there any time soon. If it's merely a matter of "We need to figure out how but haven't gotten around to it" then it might be time to make that dive. If it's more "we know exactly what's wrong and have no way forward", then I think you're right and we should drop this.

Magisus: We also have no automated testing around pulling in actual translations at runtime yet. That would be the next big step IMO to making this safer to do, to make sure that either Puppet and friends ignore the MO files on AIX, or that they actually work, and thereby make sure they're NOT broken.

branan: Boost and libstdc++'s locale stuff just bombs hard on AIX. We haven't taken the time to figure out how to actually do localization on that platform.

Magisus: Does that mean it might work in a Ruby codebase?

branan: ¯\_(ツ)_/¯ I have no idea what Ruby's locale management stuff looks like under the hood. I'm willing to commit to "might work"

Magisus: I'll do some manual validation to make sure that this at least won't break anything, and better yet, if it works.

Magisus: Update, if the locales we want are installed on AIX, this translation tooling works just fine!

Magisus:
pl-gettext packages for EL 5 and AIX 6 and 7 have been built, so this should be good to go now.

Magisus:
This needs https://github.com/puppetlabs/leatherman/pull/238 merged before the generated translation files can actually be found at runtime by the C++ projects, but nothing will fail in CI (since at the moment we don't have any translations).

PR #1031, opened 2017-04-13
(maint) Add explanation about elevate.exe

melissa:
I always forget why we have elevate.exe in puppet-agent. Since many new
projects use puppet-agent as a template, we should have a little more
detail as to why we include it. This way we'll know if we do or do not
want to include it in future projects.


melissa:
ping @joshcooper @McdonaldSeanp

melissa:
I've retargeted this PR at master!

PR #1032, opened 2017-04-13
(PUP-7425) bump net-ssh to 4.1.0

Iristyle:
- Note that net-ssh 3.3.0beta1 became a dependency of Beaker 3
   (see BKR-40) to address some UTF-8 issues. When that version was
   found to cause issues with bundler (based on its version) number,
   Beaker updated to 4.0+ in Beaker 3.11.0+.

   Given the problems with Unicode handling in the 2.9.2 version of
   this gem, bump the dependency before anything is reported by users.

 - Note that net-ssh 3 and higher require Ruby 2.  This has no bearing
   on agent packages which already ship Ruby 2.1, but could be
   problematic for gem installs - therefore only update packages here.
   The puppet ext/project_data.yaml has not been updated.

   This new requirement will not be a problem within Puppet server as
   the loading of the library is guarded behind a feature in the Puppet
   code. When net-ssh fails to load there due to a Ruby incompatibility
   the :ssh feature will be set to false inside Puppet.

### DIFF
--- a/configs/components/rubygem-net-ssh.rb
+++ b/configs/components/rubygem-net-ssh.rb
@@ -1,6 +1,6 @@
 component "rubygem-net-ssh" do |pkg, settings, platform|
-  pkg.version "2.9.2"
-  pkg.md5sum "ac7574a89e2b422468d98f5387ceb41e"
+  pkg.version "4.1.0"
+  pkg.md5sum "6af1ff8c42a07b11203058c9b74cbaef"
   pkg.url "https://rubygems.org/downloads/net-ssh-#{pkg.get_version}.gem"
 
   pkg.replaces 'pe-rubygem-net-ssh'


Iristyle:
/cc @joshcooper @MosesMendoza @shermdog @tphoney @DavidS 

Not sure if this is the right branch to land this one, but put up for discussion at the very least.

Relates to https://github.com/puppetlabs/puppet/pull/5801

shermdog:
This *could* cause issues for the Huawei agent / our NETCONF work.  Both rely on net-netconf which is currently using 

```
s.add_dependency('net-ssh', '>= 2.9', '< 3.3')
```
That gem goes into the agent's vendored ruby. 

We are in the process of updating that library, but still working out release stuff and all that.

/cc @ScottGarman

ScottGarman:
I'll ping Huawei and see if they're okay with upgrading this gem.

Iristyle:
Thanks for the info @shermdog / @ScottGarman - had a feeling `net-scp` or other `net-*` related gems could have issues here.

We probably need to vet this a bit more / potentially re-target at Puppet 5.  I don't want to blow up `puppet device` land with this change as we don't currently have to make it right now, and could plausibly defer.

I don't know how common it is to need to send commands `>= 128` bytes in length that contain UTF-8 in them, but that's the crux of the problem with the older version of the gem (and why it needed to be updated for Beaker).  See https://github.com/net-ssh/net-ssh/pull/394 for more info.

I believe there were also a number of other bugfixes that you might be interested in - changelog is at https://github.com/net-ssh/net-ssh/blob/master/CHANGES.txt

Iristyle:
@ScottGarman / @shermdog were you able to check in with Huawei?

ScottGarman:
@Iristyle I pinged them on Apr 13 and haven't heard back from them. I think it's unlikely they will have problems with this, and even if they do, it is possible for us to use the older version of this gem for the huaweios platform only. I'm inclined not to block this progress based on those considerations.

Iristyle:
Per @McdonaldSeanp request output from `bundle exec inspect puppet-agent windows-2012r2-x64` against 360db8a55c783f04b0f34789b0e4ffae4b17e11c - https://gist.github.com/Iristyle/dd669745a0d37134e2d7950ee16b213b

McdonaldSeanp:
From a vanagon perspective, based on the output of the gist this PR is functionally correct

👍

ScottGarman:
:+1:

Iristyle:
For future reference, the gem content should be mirrored locally to buildsources.delivery.puppetlabs.net to ensure this works properly.  I failed to do that.

PR #1034, opened 2017-04-17
(PA-1084) Add path to readlink command to $PATH for Solaris runtime builds

ScottGarman:
This removes the error: "runtime.sh: line 35: readlink: command not found"
seen when building runtime for Solaris build targets.


McdonaldSeanp:
👍

PR #1037, opened 2017-04-20
(PA-1035) Remove precise from deb targets

McdonaldSeanp:
This is used in PE promotion and that OS has gone EOL for support


ScottGarman:
:+1:

PR #1041, opened 2017-05-01
(maint) Add `needs :cxx14` support to older homebrew versions

kurtwall:
Prior to this commit, macOS 10.12 packaging is failing with the
following homebrew error when installing build dependencies:
```
 Error:` Invalid formula:
  /usr/local/Library/Taps/homebrew/homebrew-core/Formula/mame.rb
  Error: Cannot tap homebrew/core: invalid syntax in tap!
  "cxx14" is not a recognized standard
```
This appears to be related to an upstream change that landed recently at
https://github.com/Homebrew/homebrew-core/pull/12146/commits/3a830725fda3b63e7f2753a8de5d68079caf7b1b.

New versions of Homebrew add `needs :cxx14`, but don't allow execution
as root, which is typically how we currently build puppet-agent.

As a workaround, this commit patches our homebrew to backport support
for `needs :cxx14` in brew forumlae.

Shamelessly copypasta-ed from @geoffnichols


Magisus:
Thanks for putting this up, we'll catch the other branches when we do the mergeup after the release 👍

Iristyle:
This was separately pulled into stable at 2e2e8056bab70a382412c55f76f1d222c58aece9

PR #1042, opened 2017-05-02
(PA-1023) Update LEATHERMAN_LOCALE_VAR to leatherman install root

Magisus:
Improvements to the Leatherman tooling around installing and using
locale files require the LEATHERMAN_LOCALE_VAR point to the Leatherman
install root, which on Windows corresponds to `<user-defined install
location>/puppet`. In `environment.bat` we set an environment variable
PUPPET_DIR, which points to this directory (`$PL_BASEDIR/puppet`). This
commit updates the leatherman config to use that environment variable
when determine where to install and load locale files.


joshcooper:
Not sure why the travis failure?

Magisus:
Hmm, I even rekicked it once already...

McdonaldSeanp:
👍

PR #1044, opened 2017-05-04
(maint) Update gettext-setup gem to 0.24

Magisus:
This version includes the `negotiate_locale!` method that we would like
to use in Puppet to detect the user's locale.

### DIFF
--- a/configs/components/rubygem-gettext-setup.rb
+++ b/configs/components/rubygem-gettext-setup.rb
@@ -1,6 +1,6 @@
 component "rubygem-gettext-setup" do |pkg, settings, platform|
-  pkg.version "0.20"
-  pkg.md5sum "53bbb01f01e68fbfa9f2ff11c466e60b"
+  pkg.version "0.24"
+  pkg.md5sum "f766a5e12bbad9f85905638c500e08f6"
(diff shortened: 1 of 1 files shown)


Iristyle:
https://tickets.puppetlabs.com/browse/PA-1077 strikes again!

Looks like the `.gemspec` is already using `[">= 0.10", "< 1"]`, so should be using `0.24` in TravisCI / AppVeyor / local dev already per https://github.com/puppetlabs/puppet/blob/master/.gemspec#L42

configs/components/rubygem-gettext-setup.rb
```diff
@@ -1,6 +1,6 @@
 component "rubygem-gettext-setup" do |pkg, settings, platform|
-  pkg.version "0.20"
-  pkg.md5sum "53bbb01f01e68fbfa9f2ff11c466e60b"
+  pkg.version "0.24"
+  pkg.md5sum "f766a5e12bbad9f85905638c500e08f6"
```
Iristyle: verified sha locally 👍

Magisus:
That is correct, and that's why I didn't catch it till after it was merged. This will allow us to un-revert https://github.com/puppetlabs/puppet/pull/5830.

PR #1045, opened 2017-05-09
(PA-793) Correct PA package selection

jcoconnor:
Some platforms (e.g. osx, sles), use the SUITE_VERSION instead of SHA in
package name - update package_version to use this.

This is the Commit for Stable to match master


ScottGarman:
:+1:

Magisus:
👍

PR #1047, opened 2017-05-10
(maint) remove unused Ruby patch files

Magisus:
The patches removed here were originally copied over from the Ruby 2.1
files, but are not being applied in the Ruby 2.3 config because they
ended up being unnecessary. As such, they are dead files and should be
removed.


McdonaldSeanp:
👍

Magisus:
@MosesMendoza updated.

PR #1049, opened 2017-05-12
(maint) Update md5sum for precompiled nokogiri

geoffnichols:
This commit updates the md5um for the precompiled nokogiri gem
for huaweios.


configs/components/rubygem-nokogiri.rb
```diff
@@ -9,7 +9,7 @@
     if settings[:ruby_version] == "2.1.9"
       pkg.md5sum "2ad8b4dd69f6e2b296dfaba293e9cb11"
     elsif settings[:ruby_version] == "2.3.3"
-      pkg.md5sum "443e5b3bea092cd955e125c8e8bda0d6"
+      pkg.md5sum "65620f5c36061fbb07076cdbe086fd35"
```
Iristyle: Will merge, but as a side note I think this is wrong to not bump the version / path of this precompiled asset at the same time.  It guarantees we won't be able to go back in time and build against an old puppet-agent SHA.

PR #1053, opened 2017-05-22
(PA-1132) Update Windows installer message

Iristyle:
- Note in the Windows installer that master config is optional and
   that Puppet will work standalone using this installer.


Iristyle:
All set now @McdonaldSeanp

PR #1055, opened 2017-05-24
(PA-1124) Add Nano Server Ruby 2.4 patch

Iristyle:
- This patch is from Ruby and has landed on their "trunk", but has
   not been added to the 2.4 source tree.  It's possible that it may
   not see the light of day until 2.5.

 - In earlier technical preview versions of Microsoft Nano Server,
   the COM internals were setup to allow calls to OleInitialize to
   succeed, despite it being unsupported on Nano Server, for the sake
   of vendors recompiling / testing their code with Nano.

   OleInitialize, according to MDSN documentation always calls
   CoInitializeEx(NULL, COINIT_APARTMENTTHREADED) per
   https://msdn.microsoft.com/en-us/library/windows/desktop/ms690134%28v=vs.85%29.aspx
   given OLE is not thread-safe.

   All COINIT values are specified at:
   https://msdn.microsoft.com/en-us/library/windows/desktop/ms678505

 - Ruby has traditionally initialized COM by calling OleInitialize,
   which is unsupported on Nano given only the MTA is supported on
   Nano. The solution is to instead call CoInitializeEx with
   COINIT_MULTITHREADED when initializing the COM subsystem.

   This necessarily means Nano won't support some OLE functions, but
   general purpose COM calls can still go through the WIN32OLE class
   in Ruby, as long as they conform to other rules with Nano and
   actually exist.

 - Should this functionality ever be ported to Ruby 2.4 this patch can
   be removed.

   See:

   Filed Ruby bug - https://bugs.ruby-lang.org/issues/12371
   MWrock PR for Chef patch - https://github.com/ruby/ruby/pull/1423
   Nobu commit to Ruby trunk -
   https://github.com/ruby/ruby/commit/8feb9779182bd4285f3881029fe850dac188c1ac

### DIFF
--- a/resources/patches/ruby_241/PA-1124_add_nano_server_com_support-8feb9779182bd4285f3881029fe850dac188c1ac.patch
+++ b/resources/patches/ruby_241/PA-1124_add_nano_server_com_support-8feb9779182bd4285f3881029fe850dac188c1ac.patch
@@ -1,0 +1,94 @@
+commit e37ab5a673c9d121208d80a0a1128de346f46a19
+Author: suke <suke@b2dd03c8-39d4-4d8f-98ff-823fe69b080e>
+Date:   Sat Feb 25 22:28:22 2017 +0000
+
+    ext/win32ole/win32ole.c(ole_initialize): avoid to fail in Windows nano server.
+    This is experimental. Thanks to mwrock, Ethan Brown.
+    
+    git-svn-id: svn+ssh://ci.ruby-lang.org/ruby/trunk@57715 b2dd03c8-39d4-4d8f-98ff-823fe69b080e
+    
+    Original patch at 8feb9779182bd4285f3881029fe850dac188c1ac was errant
+    and included an additional line.
+    
+    This one has been updated to be correct
+
+diff --git a/ext/win32ole/win32ole.c b/ext/win32ole/win32ole.c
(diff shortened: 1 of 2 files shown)


resources/patches/ruby_241/PA-1124_add_nano_server_com_support-8feb9779182bd4285f3881029fe850dac188c1ac.patch
```diff
@@ -0,0 +1,91 @@
...
+@@ -817,16 +819,23 @@ ole_initialize(void)
+     }
+ 
+     if(g_ole_initialized == FALSE) {
++        if(g_running_nano) {
++            hr = CoInitializeEx(NULL, COINIT_MULTITHREADED);
++        } else {
++            hr = OleInitialize(NULL);
++        }
+         hr = OleInitialize(NULL);
```
MosesMendoza: Looks like this was accidentally left in - meaning the hresult is always from OleInitialize instead of from the conditional. Possibly misapplication of the original patch, as its not in https://github.com/ruby/ruby/pull/1423

McdonaldSeanp:
LGTM 👍

Iristyle:
Great catch on the bad patch @MosesMendoza - totally missed it.

I think we're good to go now.

PR #1056, opened 2017-05-24
(PA-1138) Add el-7-ppc64le (Power8) platform

ScottGarman:
Two things worth noting here:

1. There was an existing platform definition file, that was originally a community contribution. It has been absorbed into the official way we build things now.

2. This commit includes multiple rbconfigs for different ruby versions (2.1.9/2.3.3/2.4.1), to facilitate the backport to stable.


configs/components/facter.rb
```diff
@@ -110,7 +110,7 @@
     skip_blkid = 'OFF'
   elsif platform.is_rpm?
     if (platform.is_el? && platform.os_version.to_i >= 6) || (platform.is_sles? && platform.os_version.to_i >= 11) || platform.is_fedora?
-      pkg.build_requires "libblkid-devel" unless platform.architecture == "s390x"
+      pkg.build_requires "libblkid-devel" unless platform.is_cross_compiled?
```
Magisus: This won't catch any additional things, will it? Are we not cross-compiling anything else besides s390x and now this new one?

McdonaldSeanp: We cross compile many things, including solaris, AIX, huawei, cisco....

ScottGarman: I'll update this with a comment - the reason this is necessary is if we have libblkid-devel installed, facter will try to link against the native (x86_64) library and not the one in our sysroot. So we want to avoid having this package installed in all cross-compiled situations.

Magisus:
Why are there two rbconfigs for 2.1.9? What is the difference between the `gnu` and not `gnu` versions?

ScottGarman:
@Magisus I snuck an additional rbconfig in there for the Ubuntu-based Power8 builds, since we're going to want to enable stable builds for these platforms. I can move that out into a separate PR if that seems clearer.

Magisus:
Ah gotcha. Maybe not a separate PR, but maybe a separate commit describing the fact that it's for a different Power8 OS?

McdonaldSeanp:
Yeah, either a second commit, or a longer description (including justification for the second rbconfig) will do

ScottGarman:
@Magisus good call, I've separated that out into its own commit now.

McdonaldSeanp:
From a vanagon perspective the changes look sound

McdonaldSeanp:
👍

Magisus:
Thanks! 👍

configs/components/facter.rb
Magisus: Wait, I thought we did not cross compile AIX? (this is unrelated to this line of code)

jcoconnor:
@ScottGarman Looks ok based on our recent pairing/intro on same.

PR #1057, opened 2017-05-26
(PA-516) Update augeas to 1.8.0.

jtappa:
- Update augeas from 1.4.0 to 1.8.0
- Removes two patches that have been incorporated to upstream since 1.4.0


Magisus:
Verified that this builds on all platforms 👍

PR #1058, opened 2017-05-26
(PA-1139) Move Hiera config to puppet

jcoconnor:
*tentative - working on this*
Hiera now lives in puppet, so hiera.yaml is pulled in from the puppet
component/repo instead of hiera.

### DIFF
--- a/configs/components/hiera.rb
+++ b/configs/components/hiera.rb
@@ -30,12 +30,11 @@ component "hiera" do |pkg, settings, platform|
     --man \
     --mandir=#{settings[:mandir]} \
     --no-batch-files \
+    --no-configs \
     #{flags}"]
   end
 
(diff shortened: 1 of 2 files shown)


jcoconnor:
This is the *new instance* of former PR https://github.com/puppetlabs/hiera/pull/396

Iristyle:
Ping when the corresponding PR is up against the Puppet repo - we'll need to cross-check against that.

jcoconnor:
See PR https://github.com/puppetlabs/puppet/pull/5919 for the other half.

jcoconnor:
@Iristyle This appears to be installing the correct ```hiera.yaml``` file.

jcoconnor:
Closing in preference for #1061

PR #1059, opened 2017-05-26
(PA-1140) Add support for downloading puppetserver from nightlies

Magisus:
We recently started trying to test against puppetserver nightly builds
instead of released tags. These builds must be downloaded from a
different url than released builds. This commit checks the
SERVER_VERSION param and switches to the correct url when 'latest' has
been specified.


McdonaldSeanp:
👍

PR #1060, opened 2017-05-30
(PA-1128) Switch to building with Ruby 2.4.1

Magisus:
The config files for Ruby 2.4.1 were adding in a previous commit. This
commit switches to building and using Ruby 2.4.


McdonaldSeanp:
👍

PR #1062, opened 2017-05-30
(maint) Update md5 sums for rbconfigs

Magisus:
It is unclear whether these sums are actually being used for anything given
successful builds even when they were wrong. But this commit updates
them to be correct for the files they point to, for good measure.


ScottGarman:
:+1: the updated md5s look correct based on what I can run locally from master in that directory.

ScottGarman:
Merging as this is a trivial change and I need it for the next PR to add the redhat-power8 entry.

PR #1071, opened 2017-06-06
(PA-1168) Patch STOMP to detect failed connections and reconnect

MikaelSmith:
Previously failed connections would take minutes or hours to detect,
then leave MCO in an unusable state. Patch STOMP to handle failed
connections as soon as the timeout limits are reached and reconnect
cleanly.

The specific problem is an issue with Ruby/OpenSSL interaction, where
it's not thread-safe to close a socket while reading from it (which is
precisely what the Stomp gem does). A simple reproduction of that issue
can be found at https://gist.github.com/richardc/b5a1d4d28c8bbae24cd5.

### DIFF
--- a/resources/patches/ruby-stomp/connection_loss_reconnect_fix.patch
+++ b/resources/patches/ruby-stomp/connection_loss_reconnect_fix.patch
@@ -1,0 +1,125 @@
+From ed5da4d1d857c427bf995e50c9baa25befaa3175 Mon Sep 17 00:00:00 2001
+From: Reid Vandewiele <reid@puppetlabs.com>
+Date: Thu, 1 Jun 2017 16:22:08 -0700
+Subject: [PATCH] Ensure IO#gets exits in a reasonable time
+
+According to the notes on heartbeats, `max_hbrlck_fails` should be
+enabled in order to determine when heartbeats stop arriving. This is
+required because the receive thread of the MCO daemon spends most of its
+time waiting for something to come in over the wire. That ties up the
+"read lock" used to provide exclusive access to the underlying network
+socket.
+
+However, the blocking receive should consume heartbeats and update the
+"last received" timestamp (`@lr`). The heartbeat read thread also
+consumes pings coming in over the wire in case the process is not
+blocking on receive.
+
+Normal operation means that a missed hbrlck is normal, but two in a row
+should be abnormal (the heartbeat thread may try to read and get locked,
+but the receive thread should read the heartbeat before the heartbeat
+thread resumes - networking timing could impact this in bad ways - and
+reset the "last received" timestamp which will trigger resetting the
+`read_lock_count` and `lock_fail_count` the next time the heartbeat
+thread wakes).
+
+When the `lock_fail_count` exceeds `@max_hbrlck_fails`, the heartbeat
+thread closes the socket and kills the heartbeat threads. The
+expectation is that this should cause the `IO#gets` call in the receive
+call to return. That doesn't happen until TCP timeouts occur, which can
+take minutes or hours. When the TCP timeout occurs, it appears to leave
+the main thread in an unexpected state that sometimes results in SIGABRT
+and is unreachable via the broker.
+
+To exit `gets` early, trigger an exception in the `receive` thread.
+That interrupts the `gets` call and is caught in the `transmit` or
+`__old_receive` method and initiates a reconnect.
+---
+ lib/connection/heartbeats.rb |  6 +++---
+ lib/connection/netio.rb      | 19 +++++++++++++++----
+ lib/stomp/connection.rb      |  1 +
+ 3 files changed, 19 insertions(+), 7 deletions(-)
+
+diff --git a/lib/connection/heartbeats.rb b/lib/connection/heartbeats.rb
+index ae350dd..dc0cb15 100644
+--- a/lib/connection/heartbeats.rb
++++ b/lib/connection/heartbeats.rb
+@@ -232,9 +232,9 @@ module Stomp
+             # Retry on max lock fails.  Different logic in order to avoid a deadlock.
+             if (@max_hbrlck_fails > 0 && lock_fail_count >= @max_hbrlck_fails)
+               # This is an attempt at a connection retry.
+-              begin
+-                @socket.close # Attempt a forced close
+-              rescue
++              @gets_semaphore.synchronize do
++                @getst.raise(Errno::EBADF.new) if @getst rescue nil # kill the socket reading thread if exists
++                @socket.close rescue nil # Attempt a forced close
+               end
+               @st.kill if @st   # Kill the sender thread if one exists
+               Thread.exit       # This receiver thread is done            
+diff --git a/lib/connection/netio.rb b/lib/connection/netio.rb
+index f9e84be..d0e2d2f 100644
+--- a/lib/connection/netio.rb
++++ b/lib/connection/netio.rb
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Renamed the patch to make more sense. Also submitted upstream at https://github.com/stompgem/stomp/pull/141.

resources/patches/ruby-stomp/connection_loss_reconnect_fix.patch
```diff
@@ -0,0 +1,125 @@
...
++++ b/lib/connection/heartbeats.rb
+@@ -232,9 +232,9 @@ module Stomp
+             # Retry on max lock fails.  Different logic in order to avoid a deadlock.
+             if (@max_hbrlck_fails > 0 && lock_fail_count >= @max_hbrlck_fails)
+               # This is an attempt at a connection retry.
+-              begin
+-                @socket.close # Attempt a forced close
+-              rescue
++              @gets_semaphore.synchronize do
++                @getst.raise(Errno::EBADF.new) if @getst rescue nil # kill the socket reading thread if exists
```
MosesMendoza: This form of blanked rescue is discouraged because we are rescuing _everything_ (ruby doesn't have a one-line rescue of a specific exception class). A safer but less succinct form is explicit:
```ruby
begin
  @getst.raise(Errno::EBADF.new) if !@getst.nil?
rescue Errno::EBADF
end
```

Also explicitly testing for `.nil?` instead of if defined avoids unexpected behavior if `@getst` somehow gets set to false

MosesMendoza: Why `Errno::EBADF` (bad file descriptor)?

Also [despite what ruby says](https://ruby-doc.org/core-2.3.0/Errno.html) about this class offering platform-specific error numbers, this resolves to error code 9 on both posix and windows. On Windows error 9 is `Invalid Block`

MikaelSmith: @reidmv have a minute to respond to this?

MikaelSmith: The expectation is that https://github.com/stompgem/stomp/blob/dev/lib/connection/netio.rb#L176 or https://github.com/stompgem/stomp/blob/dev/lib/connection/utils.rb#L256 will catch this. However, EBADF is what we'd expect from trying to read from a closed socket, which is close to the situation here. An alternative might be IOError or EOFError.

MikaelSmith: Generally yes. Maybe you'd like to spend some time in https://github.com/stompgem/stomp.

Less snarkily, `.nil?` seems overly defensive since there's only 2 places it's set, but I don't see a reason against it. Also, the next step after this is to exit the thread, so the blanket rescue is to ensure cleanup happens correctly.

reidmv: Sure. EBADF comes from the segfault reproduction code, where when things don't segfault (race condition apparently, doesn't always happen) and the socket is closed out from under a reader, the error raised is [Errno::EBADF](https://gist.github.com/richardc/b5a1d4d28c8bbae24cd5#file-threaded_openssl_disconnect-rb-L34). We could raise something else too. What we're trying to do is mimic whatever exception is raised when the socket is closed while being read from (except not segfaulting).

@MosesMendoza do you have a suggestion for what to do here instead?

MosesMendoza: ha sounds good. I should have couched that comment in `this is really a nitpick its just things like this seem to propagate themselves once they've landed`

MikaelSmith: I don't see a great way to handle cleanup via ensure while also handling the socket in the semaphore. Otherwise I'd try to do that.

MosesMendoza: Oh man the same blanket rescues in those links @MikaelSmith.

IOError seems fine. So does EBADF really - I was mostly just curious what the background was there.

reidmv: If an exception gets raised here it'll have something to do with not being able to call Thread#raise on `@getst`. I can't actually think of any kind of expected exception; the important thing here is just that it doesn't matter if something goes wrong trying to Thread#raise, the method should continue. Because we don't expect anything to go wrong, does that mean we should omit the `rescue nil`?

An alternative might be some kind of ensure. e.g.

```ruby
@gets_semaphore.synchronize do
  @getst.raise(Errno::EBADF.new) if !@getst.nil?
ensure
  @socket.close rescue nil
end
```

Note that we're still one-line rescuing here. Point is though, I don't know what exceptions might be raised here. The original code did a blanket rescue on the `@socket.close`.

MosesMendoza: Just to confirm - the cleanup is always ensuring we close the socket? I.e., no matter what happens in our attempt to interrupt the main thread, we always call `@socket.close` ?

reidmv: Honestly the ensure isn't really cleanup and so probably shouldn't be there. If we want to insert our Thread#raise but preserve pre-patch behavior of `@socket.close`, the closest thing to do would probably just be this:

```ruby
@gets_semaphore.synchronize do
  @getst.raise(Errno::EBADF.new) if !@getst.nil?
  @socket.close rescue nil
end
```

Forget ensure, it's not needed. Don't try to rescue anything from Thread#raise, we don't expect any exceptions. Use a blanket rescue on the Socket#close to leave pre-patch behavior unchanged.

MosesMendoza: Ok, so we're trying to always raise an exception on the main thread, and always call @socket.close, and we don't care if either of those things fails but we always want to try the latter regardless of the success of the former. @reidmv I think the alternative you propose seems fine given that as long as we don't expect any trouble calling `Thread#raise`, but now that I've talked myself in circles I ultimately think the code is fine as-is.

MikaelSmith: `@getst` should only be set during the `_interruptible_gets` call, and the only operation there that doesn't hold the `@gets_semaphore` is `read_socket.gets`. That has very deterministic behavior that we'll only throw the exception from within `Connection#_receive`,  which is either called during `Connection#connect` or `Connection#__old_receive`.

`__old_receive` wraps it in a blanket rescue, and is the main behavior we test. After rescuing, it'll attempt a reconnect.

`connect` generally also happens within a `Connection#socket`, where it's also wrapped in a blanket rescue. It's not explicitly private, but seems intended to be used primarily during `Connection.new`. It's definitely not called from mcollective.

So I would expect an exception thrown in either of these instances to attempt a reconnect. For `connect`, the expectation is the socket is closed when it tries to reconnect. For `__old_receive`, it tries to close the socket before reconnect but rescues and ignores all exceptions when doing so.

So preserving the prior behavior of closing the socket no matter its implications on the main thread and ignoring failures seems ok. Also, the implications of interrupting the main thread are limited.

MikaelSmith: I'll propose dropping the rescue in the PR to stomp. I don't think it's worth changing the patch on its own here.

PR #1074, opened 2017-06-08
(maint) merge stable -> master, update nokogiri tarball md5 for ruby 2.4.1

McdonaldSeanp:


configs/components/rubygem-nokogiri.rb
```diff
@@ -5,13 +5,13 @@
     # including its dependencies (mini_portile2), installing it on the target, and
     # installing the nokogiri gem using our gem command. The added files from
     # :gem_home were then tar'ed up and packed into this container tarball.
-    pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-precompiled-huaweios-#{pkg.get_version}-for-ruby-#{settings[:ruby_version]}.tar.gz"
+    pkg.url "http://buildsources.delivery.puppetlabs.net/nokogiri-precompiled-huaweios-#{pkg.get_version}-patch1-for-ruby-#{settings[:ruby_version]}.tar.gz"
     if settings[:ruby_version] == "2.1.9"
-      pkg.md5sum "4bf7a53bbfa5593041f7698f4e56a8d1"
+      pkg.md5sum "016db16f78d9e6d0c3ead89093e610a2"
     elsif settings[:ruby_version] == "2.3.3"
       pkg.md5sum "65620f5c36061fbb07076cdbe086fd35"
```
MikaelSmith: Will this SHA still be correct? Does it matter?

Magisus: Since we're not building with 2.3.3 anywhere, no, it shouldn't matter. But technically speaking we probably do need to rebuild that one too and update this at some point.

McdonaldSeanp: Right now it doesn't matter. I will open a PR later (today) hopefully to fix this md5 as well. For now we don't build with 2.3.3 so this won't effect any building/testing

PR #1075, opened 2017-06-08
(PA-1117) Update nokogiri tarball for ruby 2.3.3

McdonaldSeanp:
Update the md5sum of the nokogiri tarball for ruby 2.3.3, now that the source
tarball has been created and added to buildsources


Magisus:
That sum matches up with what I see 👍

PR #1061, opened 2017-05-30
(PA-1139) Move Hiera config to puppet

jcoconnor:
Hiera now lives in puppet, so hiera.yaml is pulled in from the puppet
component/repo instead of hiera.

This is #1058 retargeted at ```stable```

PR #1076, opened 2017-06-09
(PA-1139 PA-1169) Install Hiera 5 config files 

branan:
This supersedes #1061 

This switches to using Puppet's version of hiera.yaml, instead of hiera's. It also adds new logic for upgrades to make sure we don't accidentally lay down part of a hiera5 config when a user has their own Hiera configuration already.


jcoconnor:
@branan Going to continue this on with some fixes for windows in #1078

Should this not be going into ```stable``` rather than ```master```
/cc @geoffnichols @McdonaldSeanp

PR #1078, opened 2017-06-12
(PA-1139 PA-1169) Install Hiera 5 config files

jcoconnor:
This supersedes #1076 

Adding an additional commit to this to address preserving the hiera yaml files for windows.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -243,60 +243,44 @@ component "puppet" do |pkg, settings, platform|
 
   unless platform.is_windows?
     old_hiera = File.join(settings[:puppet_codedir], 'hiera.yaml')
-    new_hiera = File.join(configdir, 'hiera.yaml')
+    cnf_hiera = File.join(configdir, 'hiera.yaml')
     env_hiera = File.join(settings[:puppet_codedir], 'environments', 'production', 'hiera.yaml')
+    rmv_hiera = File.join(configdir, 'remove_hiera5_files.rm')
 
     # Pre-Install script to save copies of existing hiera configuration files.
     # This also creates "marker files to indicate that the files pre-existed."
     preinstall = <<-PREINST
-# Backup the old hiera location, so that we
-# can drop it back in place if the package manager
+# Backup the old hiera configs if present, so that we
+# can drop them back in place if the package manager
 # tries to remove it.
 if [ -e #{old_hiera} ]; then
   cp #{old_hiera}{,.pkg-old}
+  touch #{rmv_hiera}
 fi
-# Same treatment for conf/new hiera location
-if [ -e #{new_hiera} ]; then
-  cp #{new_hiera}{,.pkg-old}
-fi
-
-# Flag the global and production hiera.yaml files for
-# removal, if the user doesn't have them configured.
-if [ \\\( ! -e #{new_hiera} \\\) -a -e #{env_hiera} ]; then
-  touch #{new_hiera}.rm
-fi
-if [ \\\( ! -e #{env_hiera} \\\) -a \\\( -e #{new_hiera} -o -e #{old_hiera} \\\) ]; then
-  touch #{env_hiera}.rm
+if [ -e #{cnf_hiera} ]; then
+  cp #{cnf_hiera}{,.pkg-old}
+  touch #{rmv_hiera}
 fi
 PREINST
 
     # Post-install script to restore old hiera config files if the have been saved.
     # and remove any extre hiera configuration files that we laid down
     postinstall = <<-POSTINST
+# Remove any extra hiera config files we laid down if prev config present
+if [ -e #{rmv_hiera} ]; then
+  rm -f #{cnf_hiera}
+  rm -f #{env_hiera}
+  rm -f #{rmv_hiera}
+fi
+
 # Restore the old hiera, if it existed
 if [ -e #{old_hiera}.pkg-old ]; then
   cp #{old_hiera}{.pkg-old,}
   rm -f #{old_hiera}.pkg-old
 fi
-
-# Remove any extra hiera config files we laid down
-if [ -e #{new_hiera}.rm ]; then
-  rm -f #{new_hiera}.rm
-  if [ -e #{new_hiera} ]; then
-    rm -f #{new_hiera}
-  fi
-fi
-if [ -e #{env_hiera}.rm ]; then
-  rm -f #{env_hiera}.rm
-  if [ -e #{env_hiera} ]; then
-    rm -f #{env_hiera}
-  fi
-fi
-
-# Finally restore the old conf/env hiera, if it existed
-if [ -e #{new_hiera}.pkg-old ]; then
-  cp #{new_hiera}{.pkg-old,}
-  rm -f #{env_hiera}.pkg-old
+if [ -e #{cnf_hiera}.pkg-old ]; then
+  cp #{cnf_hiera}{.pkg-old,}
+  rm -f #{cnf_hiera}.pkg-old
 fi
 POSTINST
 
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -223,6 +223,7 @@ component "puppet" do |pkg, settings, platform|
   pkg.directory File.join(settings[:puppet_codedir], 'environments', 'production')
   pkg.directory File.join(settings[:puppet_codedir], 'environments', 'production', 'manifests')
   pkg.directory File.join(settings[:puppet_codedir], 'environments', 'production', 'modules')
+  pkg.directory File.join(settings[:puppet_codedir], 'environments', 'production', 'data')
   pkg.install_configfile 'conf/environment.conf', File.join(settings[:puppet_codedir], 'environments', 'production', 'environment.conf')
 
   if platform.is_windows?
@@ -235,4 +236,9 @@ component "puppet" do |pkg, settings, platform|
   if platform.is_eos?
     pkg.link "#{settings[:sysconfdir]}", "#{settings[:link_sysconfdir]}"
   end
+
+  pkg.install_file "ext/hiera/hiera.yaml", File.join(settings[:puppet_codedir], 'environments', 'production', 'hiera.yaml')
+  pkg.configfile File.join(settings[:puppet_codedir], 'environments', 'production', 'hiera.yaml')
+  pkg.configfile File.join(configdir, 'hiera.yaml')
+
 end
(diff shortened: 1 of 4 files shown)


configs/components/puppet.rb
```diff
@@ -236,4 +237,56 @@
   if platform.is_eos?
     pkg.link "#{settings[:sysconfdir]}", "#{settings[:link_sysconfdir]}"
   end
+
+  pkg.install_file "ext/hiera/hiera.yaml", File.join(settings[:puppet_codedir], 'environments', 'production', 'hiera.yaml')
+  pkg.configfile File.join(settings[:puppet_codedir], 'environments', 'production', 'hiera.yaml')
+  pkg.configfile File.join(configdir, 'hiera.yaml')
+
+  unless platform.is_windows?
```
branan: I'm not sure it's necessary to wrap this? The old Hiera pre/post-install actions didn't have this conditional and it seemed to work fine over there.

jcoconnor: Would prefer the wrapping as the pre/post actions aren't relevant to windows at all and think its better to make it explicit.

branan: Reasonable. I'd actually like to fix Vanagon to throw an error on this sometime in the future. Silently ignoring things makes me sad.

jcoconnor:
Artefacts are collecting at: http://builds.delivery.puppetlabs.net/puppet-agent/b170ded5a106a9d7c82fcf8ad6c205ac211abebe/artifacts/

jcoconnor:
@branan I have had to redo the Windows/MSI installation part as it didn't behave [as per the agreed specifications](https://tickets.puppetlabs.com/browse/PA-1139?focusedCommentId=451542&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-451542) 

I think I've got the conditional testing correct (particularly "_If we detect a hiera.yaml in either $confdir or $environment, don’t install a new file in either location_")

This took a few attempts to get it working, so I resorted to fixing the Wix on the fly on the build machine.

I have also taken the opportunity to tidy/squash the commits so I'm running the whole lot through a Dynamic Vanagon build to make sure I haven't broken anything.

Will update this PR with the build locations once this is done.

I have also added  @Iristyle @McdonaldSeanp as reviewers to ensure my Wixology is non-heretical.

jcoconnor:
Added ```Don't merge``` until commit https://github.com/puppetlabs/puppet-agent/pull/1078/commits/beb476725bd2a864225761f2a4dc3715e0bb9ece is removed.

jcoconnor:
Artefacts are gathering at: http://builds.delivery.puppetlabs.net/puppet-agent/beb476725bd2a864225761f2a4dc3715e0bb9ece/artifacts/

jcoconnor:
Had to queue a rebuild - artefacts will be at: http://builds.delivery.puppetlabs.net/puppet-agent/b9053cb6b0603258a42391989981c63263d1690c/artifacts/

jcoconnor:
I have carried out the following testing on ```Windows-2012r2```

1. Fresh Installation - Hiera 5 files installed in ```environment``` & ```confdir```
2. Upgrade from PA 1.10 - Hiera 5 files are not installed because pre-existing ```hiera.yaml``` in ```confdir```
3. Upgrade from PA 1.4 - Hiera 5 files are not installed because pre-existing ```hiera.yaml``` in ```codedir```

jcoconnor:
I have also done a quick test using ```sles-12```

The upgrade operation also preserves the ```/etc/puppetlabs/puppet/hiera.yam``` file if it has been changed.

The Linux upgrade seems to be a bit cleverer in that it is able to detect if the file has actually been changed from the default, whereas the Windows upgrade will not put in the new files if if finds a version ```hiera.yaml``` there.

MosesMendoza:
It would be nice to have an acceptance test added that validates the install (and ideally upgrade) case - not sure if that's possible especially in the timeframe we have left..

configs/components/puppet.rb
```diff
@@ -236,4 +237,56 @@
...
+    old_hiera = File.join(settings[:puppet_codedir], 'hiera.yaml')
+    new_hiera = File.join(configdir, 'hiera.yaml')
```
MosesMendoza: @jcoconnor per @geoffnichols's comment on PA-1139, we should not be overwriting `/etc/puppetlabs/puppet/hiera.yaml` if it already exists. So this needs to also backup `$new_hiera` if it exists on pre-install, and then copy it back on post-install, unconditionally

branan: We use `config(noreplace)` on RPM, so the package manager handles that for us. I can double-check Debian. For windows we might need additional logic.

branan:
@MosesMendoza We don't currently have any infrastructure in place for puppet-agent upgrade tests outside of PE. I'm not sure that we have the time to set that up, or that it's worth doing when we only have one test case that would use it.

configs/components/puppet.rb
jcoconnor: Windows ignores this logic completely and implements it in Wix.
I've designed it to check if either of the ```hiera.yaml``` files are pre-existing - if that is so, the hiera5 files are *not* installed.

We could make backup copies etc., but this is more complex so would prefer to stick with the simpler implementation here.

resources/windows/wix/properties.wxs.erb
```diff
@@ -173,5 +173,26 @@
...
+    <!--
+      Check if there is a pre-existing Hiera 3 format file in either:
+        1. C:\ProgramData\PuppetLabs\code
+        2. C:\ProgramData\PuppetLabs\puppet\etc
+      If either file exists, then we need to block the installation of the new format
+      Hiera 5 default files.
+      This Programdata directory is fixed, so its not influenced by a change in
+      INSTALLDIR, so we can use the util:search property instead of a custom action.
+      These properties are used in appdatafiles.wxs.
+    -->
```
jcoconnor: For reference @branan & @MosesMendoza This is the detection/search for pre-existing files in windows which causes ```PUP_PREEXISTING_CODE_HIERA``` or ```PUP_PREEXISTING_CONF_HIERA``` to be populated.

These are used to drive a conditional for the components (see above) to block them installing if either of these properties exist.

resources/windows/wix/appdatafiles.wxs
```diff
@@ -160,6 +156,32 @@
...
+        Permanent="yes"
+        NeverOverwrite="yes"
+        Guid="998A7670-BF42-4F03-B144-4528EADD0CE3"
+        Directory="PuppetConfDir">
+        <CreateFolder />
+        <File
+          Id="HieraConf"
+          KeyPath="yes"
+          Source="SourceDir\CommonAppDataFolder\Puppetlabs\puppet\etc\hiera.yaml" />
+        <Condition>NOT PUP_PREEXISTING_CODE_HIERA AND NOT PUP_PREEXISTING_CONF_HIERA</Condition>
```
jcoconnor: This conditional blocks the installation of this component (file) if either of these properties exist.

configs/components/puppet.rb
MosesMendoza: Per conversation in hipchat, - config(noreplace) will still ovewrite the file if the user hasn't edited it. So we still need logic to backup and then restore any existing `/etc/puppetlabs/puppet/hiera.yaml` file

jcoconnor:
@MosesMendoza @branan Does this change look better ?
Putting it through build and test.

jcoconnor:
New artefacts at: http://builds.delivery.puppetlabs.net/puppet-agent/3cada94a12a4c01e0f7f7a1fc94f7a6420a6b77e/artifacts/

jcoconnor:
Rebuilding again - artefacts will be available at: http://builds.delivery.puppetlabs.net/puppet-agent/76d067c903519ff313c15c11c0358e7328a72979/artifacts/

configs/components/puppet.rb
```diff
@@ -235,4 +236,55 @@
...
+
+    # Post-install script to restore old hiera config files if the have been saved.
+    # and remove any extre hiera configuration files that we laid down
+    postinstall = <<-POSTINST
+# Remove any extra hiera config files we laid down if prev config present
+if [ -e #{rmv_hiera} ]; then
+  rm -f #{cnf_hiera}
+  rm -f #{env_hiera}
+  rm -f #{rmv_hiera}
+fi
```
MosesMendoza: With the logic in this script, if the file `/etc/puppetlabs/code/hiera.yaml` exists, but the file `/etc/puppetlabs/puppet/hiera.yaml` does not, we will:
- back up, overwrite, and remove, and restore `/etc/puppetlabs/code/hiera.yaml`
- write and then remove `/etc/puppetlabs/hiera.yaml` 

Leaving the host in a state without any `/etc/puppetlabs/hiera.yaml` (it had none to begin with). I was under the impression we wanted to install the new hiera 5 format yaml to `/etc/puppetlabs/puppet/hiera.yaml` if it does not exist. Is that right?

configs/components/puppet.rb
```diff
@@ -235,4 +236,55 @@
...
+    cnf_hiera = File.join(configdir, 'hiera.yaml')
+    env_hiera = File.join(settings[:puppet_codedir], 'environments', 'production', 'hiera.yaml')
+    rmv_hiera = File.join(configdir, 'remove_hiera5_files.rm')
+
+    # Pre-Install script to save copies of existing hiera configuration files.
+    # This also creates "marker files to indicate that the files pre-existed."
+    preinstall = <<-PREINST
+# Backup the old hiera configs if present, so that we
+# can drop them back in place if the package manager
+# tries to remove it.
```
MosesMendoza: Can and should we back up and restore `/etc/puppetlabs/code/environments/production/hiera.yaml` ? Ie for users going from 5.0.0 to 5.0.1 ?

configs/components/puppet.rb
MosesMendoza: Confirmed in discussion on hipchat that the generic rule here is _if any existing hiera.yaml file exists in any of the known locations, do not install any of the new files_

jcoconnor:
Moses,

Can certainly add this to the list of files in case the package manager
should overwrite it.

configs/components/puppet.rb
geoffnichols: That sounds right to me. /cc @branan

jcoconnor:
👍 to Branan's last commit.

configs/components/puppet.rb
```diff
@@ -235,4 +236,61 @@
...
+
+# Restore the old hiera, if it existed
+if [ -e #{old_hiera}.pkg-old ]; then
+  mv #{old_hiera}{.pkg-old,}
+fi
+if [ -e #{cnf_hiera}.pkg-old ]; then
+  mv #{cnf_hiera}{.pkg-old,}
+fi
+if [ -e #{env_hiera}.pkg-old ]; then
+  mv #{cnf_hiera}{.pkg-old,}
```
MosesMendoza: I thinks this was supposed to be `mv #{env_hiera}{.pkg-old,}` ?

PR #1080, opened 2017-06-12
(maint) Update vanagon version to automatically pull in new Z versions

Magisus:
This commit updates the Gemfile to use Vanagon 0.12.1, while changing
the version pin to allow new Z releases to be pulled in automatically.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ def vanagon_location_for(place)
   end
 end
 
-gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '0.11.3')
+gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '~> 0.12.1')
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -10,7 +10,7 @@ def vanagon_location_for(place)
   end
 end
 
-gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '0.11.3')
+gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '~> 0.12.1')
```
McdonaldSeanp: Won't this just continue to act in the same way (where it doesn't auto update) because we specified the Z release ?

Magisus: According to this https://www.devalot.com/articles/2012/04/gem-versions.html the last-specified number is allowed to vary.

McdonaldSeanp: Ah, yes

PR #1068, opened 2017-06-02
(PA-783) Localise Administrators and Everyone ids

jcoconnor:
Localise the "Administrators" Group and "Everyone" User id using the
Wix OSInfo service and a custom action, so that file permissions and
user group associations are correct in a localised (e.g. German/Chinese)
environment.

PR #1082, opened 2017-06-14
(PA-1189) Add GetLocalisedAccountNames to ex Seq

jcoconnor:
Ensure that this Custom Action is regardless of whether UI is used, i.e.
runs in execute sequence if it hasn't already run in the UI sequence.

Otherwise the properties for localised Admin/Everyone aren't set during
unattended installs causing incorrect permission on the facts.d
directory.

### DIFF
--- a/resources/windows/wix/sequences.wxs.erb
+++ b/resources/windows/wix/sequences.wxs.erb
@@ -14,7 +14,6 @@
         CMDLINE_INSTALLDIR
       </Custom>
       <Custom Action='SetIniPropertyValues' After='AppSearch' />
-      <Custom Action='GetLocalisedAccountNames' After='AppSearch' />
       <!-- PUPPET_MASTER_SERVER -->
       <Custom Action='SetFromIniPuppetMasterServer' Before='FileCost'>
         INI_PUPPET_MASTER_SERVER
@@ -69,6 +68,7 @@
 
     <InstallExecuteSequence>
       <Custom Action='SetIniPropertyValues' After='AppSearch' />
+      <Custom Action='GetLocalisedAccountNames' After='AppSearch' />
       <!-- INSTALLDIR -->
       <Custom Action='SaveCmdLineInstallDir' Before='AppSearch' />
       <Custom Action='SetFromCmdLineInstallDir' After='AppSearch'>
--- a/resources/windows/wix/customactions.wxs.erb
+++ b/resources/windows/wix/customactions.wxs.erb
@@ -57,7 +57,7 @@ End If
     <CustomAction
       Id="GetLocalisedAccountNames"
       Script="vbscript"
-      Execute="firstSequence"
+      Execute="immediate"
       Return="check">
         <![CDATA[
       On Error Resume Next


resources/windows/wix/sequences.wxs.erb
```diff
@@ -14,7 +14,9 @@
         CMDLINE_INSTALLDIR
       </Custom>
       <Custom Action='SetIniPropertyValues' After='AppSearch' />
-      <Custom Action='GetLocalisedAccountNames' After='AppSearch' />
+      <Custom Action='GetLocalisedAccountNames' After='AppSearch'>
```
Iristyle: Is there no other way of configuring this action?  `SetIniPropertyValues` runs in unattended mode no problem, so I'm not sure this makes sense to me?

Iristyle: What if we changed the `execute` type?
http://wixtoolset.org/documentation/manual/v3/xsd/wix/customaction.html

jcoconnor:
I have built this successfully for 32 and 64 bit platforms and installed on both using the following install commands:
64bit: ```msiexec.exe /i "C:\cygwin64\var\tmp\puppet-agent-x64.msi" /qn /L*V .\install-puppet.log PUPPET_AGENT_STARTUP_MODE=Manual```
32bit:```msiexec.exe /i "C:\cygwin\var\tmp\puppet-agent-x86.msi" /qn /L*V .\install-puppet.log PUPPET_AGENT_STARTUP_MODE=Manual```

In both cases, the installation completed and the ```facts.d``` directory was set to the correct permissions.

I also examined the MSI install log and verified that the ```GetLocalisedAccountNames``` CA was running and the properties were being set, e.g. 32 bit log extract:
```
MSI (s) (00:84) [17:12:37:438]: Doing action: GetLocalisedAccountNames
MSI (s) (00:84) [17:12:37:438]: Note: 1: 2205 2:  3: ActionText 
Action ended 17:12:37: AppSearch. Return value 1.
MSI (s) (00:28) [17:12:37:438]: Generating random cookie.
MSI (s) (00:28) [17:12:37:438]: Created Custom Action Server with PID 3556 (0xDE4).
MSI (s) (00:8C) [17:12:37:454]: Running as a service.
MSI (s) (00:8C) [17:12:37:454]: Hello, I'm your 32bit Impersonated custom action server.
MSI (s) (00!84) [17:12:37:501]: PROPERTY CHANGE: Adding PUP_LCL_ACCOUNT_EVERYONE property. Its value is 'Everyone'.
MSI (s) (00!84) [17:12:37:501]: PROPERTY CHANGE: Adding PUP_LCL_ACCOUNT_ADMINISTRATORS property. Its value is 'Administrators'.
Action start 17:12:37: GetLocalisedAccountNames.
```

resources/windows/wix/sequences.wxs.erb
Iristyle: I completely misread what was going on here - I see that the condition is there only to prevent the action from running multiple times.  We probably don't need that bit? (or should leave a comment about what the condition is about)

Making sure `GetLocalisedAccountNames` appears in `InstallExecuteSequence` so it's always run in silent installation mode seems like the right thing to do.

jcoconnor: ```execute=immediate``` is the default - and is probably the best mode to run it in.
Will set this explicitly though (should really be set elsewhere too).

Do you think its ok to lift it completely from the UI sequence - its not needed there (whereas the IniProperties is probably needed in both sequences).

Iristyle:
Relates to #1068

jcoconnor:
Another ```ad-hoc``` build in the works.

geoffnichols:
Passing ad-hoc CI run (including on windows10ent-32a)  at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-daily-facter-ad-hoc/8/.

PR #1085, opened 2017-06-15
(maint) Use exec in command wrapper scripts

branan:
Without using exec, the shell process for the wrapper script
will just hang around. Using exec replaces the shell process,
saving a bit of memory and cleaning up `ps` output.


Magisus:
Looks good to me, pending ad-hoc verification that it fixes the pxp-agent test that examines puppet-agent process IDs.

MikaelSmith:
👍

PR #1089, opened 2017-06-19
(maint) Hardcode /opt/puppetlabs in command wrapper scripts

branan:
Previously, the wrapper scripts tried to calculate where they
were based on $0. Unfortuntely, executing them via a symlink
in an arbitrary path would break this heuristic. Since
we have no plans to change our paths any time soon, this hardcoding
is better than adding even more complex logic to try to track
down the path of the agent.


Magisus:
Ah, I see. Okay 👍

johnduarte:
@cthorn42 this should address the issue you are seeing. Hopefully, this will promote tonight.

PR #1090, opened 2017-06-20
(PA-1242) merge 1.10.4-release in to stable

McdonaldSeanp:


Magisus:
We don't want to use either of those tags. Both SHAs are ahead of them.

McdonaldSeanp:
Ah, yes

McdonaldSeanp:
@Magisus updated

PR #1092, opened 2017-06-20
(packaging) Update to allow PA to ship to puppet5

melissa:
This commit updates all references to PC1 to puppet5. These are the new
repos for the new puppet5 series. This commit also removes a bunch of
instances where `output_dir` had been set unnecessarily. Vanagon stages
the packages at an identical path, so this extra setting is unnecessary.

This commit also adds in settings for a nonfinal repo name and a
nonfinal gpg key. These two things allow us to ship development builds
to the puppet5-nightly repos.


melissa:
Please don't merge this yet. This PR is mostly to stage these changes until we're ready. Beaker will not work yet with these changes in place

underscorgan:
@melissa this looks right to me.

configs/projects/puppet-agent.rb
```diff
@@ -153,7 +153,6 @@
   proj.license "See components"
   proj.vendor "Puppet Labs <info@puppetlabs.com>"
   proj.homepage "https://www.puppetlabs.com"
-  proj.target_repo "PC1"
```
melissa: We no longer need to set this in vanagon. We don't care how vanagon stages packages anymore, since we always reorder them with packaging

melissa:
Would this be better to target at the 5.1.x branch?

Magisus:
Yes, the 5.0.x branch should really be deleted now that we've shipped 5.1.0. That's probably pending moving all the PRs targeted there.

PR #1093, opened 2017-06-22
(maint) Set facter and pxp-agent SHA to release branch HEADs

Magisus:
An additional commit had landed on Facter and pxp-agent beyond what
should go out for the 5.0.0 release. This commit sets those component
SHAs to the correct branching point for the release.


McdonaldSeanp:
👍

PR #1094, opened 2017-06-22
(PA-1249) Clean up failed service restart on SLES 11 upgrade

MikaelSmith:
From pxp-agent 1.5.3 to 1.6.0, pxp-agent used a versioned binary with
pxp-agent symlinked to it. When upgrading, the binary used to start the
pxp-agent service would change. On SUSE 11, this prevented restarting
the service after upgrade because killproc would not kill the process
started with the previous binary. Use kill instead - with appropriate
guards to check it was the expected service process - to ensure we can
restart the service after upgrading from one of those versions.


McdonaldSeanp:
Seems reasonable to me 👍

MikaelSmith:
If there's a good way to guard this for only SLES 11, I'd use it.

MikaelSmith:
Nevermind, it already is. Happens inside a block checking for `sysv`.

MikaelSmith:
Slightly cleaner to fix in the init script: https://github.com/puppetlabs/pxp-agent/pull/598

PR #1096, opened 2017-06-23
(PA-1169) add version check for hiera.yaml files

ericldelaney:
Check that the puppet and production hiera.yaml files
list version 5 for a new install

### DIFF
--- a/acceptance/tests/ensure_hiera_config_version.rb
+++ b/acceptance/tests/ensure_hiera_config_version.rb
@@ -1,0 +1,23 @@
+test_name 'C100304: Ensure hiera.yaml file contains version 5 format on new install' do
+  tag 'risk:high'
+
+  def environment_hiera_conf(agent)
+    environment_path = on(agent, puppet("config print environmentpath")).stdout.chomp
+    File.join(environment_path, 'production', 'hiera.yaml')
+  end
+
+  def puppet_hiera_conf(agent)
+    on(agent, puppet("config print hiera_config")).stdout.chomp
+  end
+
+  agents.each do |agent|
+    [puppet_hiera_conf(agent), environment_hiera_conf(agent)].each do |file|
+      step "hiera.yaml contains 'version 5'" do
+        on(agent, "cat #{file}", ) do |cat_results|
+          assert_match(/^\s*version: 5/, cat_results.stdout, "Expected to find specific hiera version")
+          assert_empty(cat_results.stderr, "Expected to not produce an error")
+        end
+      end
+    end
+  end
+end


acceptance/tests/ensure_hiera_config_version.rb
```diff
@@ -0,0 +1,38 @@
+test_name 'C100304: Ensure hiera.yaml file contains version 5 format on new install' do
+
+  require 'puppet/acceptance/temp_file_utils'
+  extend Puppet::Acceptance::TempFileUtils
+
+  def environment_hiera_conf(agent)
```
johnduarte: Rather than hard-coding the values for the hiera config locations in the test, I would recommend using puppet to introspect them.

The hiera config location is available as an atomic configuration value via `puppet config print hiera_config`.
The environment directory is also available via this mechanism and can be combined with the desired environment:
```
env_dir = on(agent, puppet('config', 'print', 'environmentpath'))
prod_hiera_conf = File.join ( env_dir, 'production', 'hiera.yaml')
```

acceptance/tests/ensure_hiera_config_version.rb
```diff
@@ -0,0 +1,38 @@
...
+      'C:/ProgramData/PuppetLabs/puppet/etc/hiera.yaml'
+    else
+      '/etc/puppetlabs/puppet/hiera.yaml'
+    end
+  end
+
+  agents.each do |agent|
+    [puppet_hiera_conf(agent), environment_hiera_conf(agent)].each do |file|
+      step "puppet agent #{file} exists" do
+        if !file_exists?(agent, file)
```
johnduarte: nit pick: In ruby, the `if !` pattern is usually expressed using `unless`

acceptance/tests/ensure_hiera_config_version.rb
```diff
@@ -0,0 +1,38 @@
...
+          fail_test("Failed to find file #{file} on agent #{agent}")
+        end
+      end
+
+      step "hiera.yaml contains 'version 5'" do
+        on(agent, "cat #{file}") do |cat_results|
```
johnduarte: Is the `file_exists?` call needed if you are going to try and cat the file? You could add an assert in this step that STDERR should be empty and remove the extra ssh call to the SUT.

acceptance/tests/ensure_hiera_config_version.rb
ericldelaney: Changed

acceptance/tests/ensure_hiera_config_version.rb
ericldelaney: removed

acceptance/tests/ensure_hiera_config_version.rb
ericldelaney: changed

acceptance/tests/ensure_hiera_config_version.rb
```diff
@@ -0,0 +1,26 @@
+test_name 'C100304: Ensure hiera.yaml file contains version 5 format on new install' do
+
+  require 'puppet/acceptance/temp_file_utils'
+  extend Puppet::Acceptance::TempFileUtils
```
johnduarte: I don't believe this module is required any more.

ericldelaney:
@johnduarte Updated version

Magisus:
LGTM 👍

Magisus:
Should this be targeted at 5.0.x?

ericldelaney:
Yes, it should be 5.0.x

PR #1097, opened 2017-06-23
(PA-1251) Call puppet agent directly in AIX service

branan:
Previously, the puppet-agent invocation in our service was
calling Ruby, with the path to puppet passed as the script.
Since PA-437, this no longer works, since the path to puppet
is a shell script now, not a ruby file.

There's no good reason for us to be calling puppet indirectly
this way, so we should instead simplify the service config.


geoffnichols:
👍

joshcooper:
Should this be targeted at 5.0.0-release? Or is it ok to target 5.0.x since that's what will promote into PE?

PR #1098, opened 2017-06-26
(PA-1254) Remove huaweios from foss_platforms

ScottGarman:
HuaweiOS is only being supported in our LTS series of puppet-agent.


Magisus:
👍

MosesMendoza:
It would be nice to have a ticket to track the decision making behind this

ScottGarman:
@MosesMendoza good point. I've created ticket PA-1254 to track this work.

ScottGarman:
The Travis failure is because I marked the ticket as internal.

PR #1099, opened 2017-06-26
(PA-1274) Fix deprecation warnings from the vanagon upgrade

Magisus:
We recently updated the vanagon version used from 0.8.2 to 0.12.1. In this gap, several pieces of functionality were changed, and deprecation warnings were added to the old usages. This PR contains three commits which clean up the bulk of the warnings:
1) `is_osx?` was renamed to `is_macos?`
2) environment variables should now be specified as two separate arguments (name and value) rather than as a hash
3) interpolation of environment variables in strings destined for the shell are now being handled differently by the intermediate steps, and a new syntax is preferred: `$$PATH` becomes `$(PATH)`

There is just one warning left, concerning the specification of a mirroring URL. However, the current alternative to `register_rewrite_rule` as used [here](https://github.com/puppetlabs/puppet-agent/blob/master/configs/projects/puppet-agent.rb#L301) is currently quite clunky, and comments in the Vanagon codebase indicate that the changes might still be under active development, so I'm leaving it for now.

On Windows there is also a warning about attempting to set file modes, which apparently still doesn't work. But that seems like a different sort of problem.

### DIFF
--- a/test.rb
+++ b/test.rb
@@ -1,1 +1,0 @@
-pkg.environment "THING", "value"
--- a/test.rb
+++ b/test.rb
@@ -1,0 +1,1 @@
+pkg.environment "THING", "value"
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -73,7 +73,7 @@ component 'augeas' do |pkg, settings, platform|
       pkg.environment "PKG_CONFIG_PATH" => "/usr/lib/pkgconfig"
       pkg.environment "PKG_CONFIG" => "/opt/pl-build-tools/bin/pkg-config"
     end
-  elsif platform.is_osx?
+  elsif platform.is_macos?
     pkg.environment "PATH" => "$$PATH:/usr/local/bin"
     pkg.environment "CFLAGS" => settings[:cflags]
   end
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -11,14 +11,14 @@ component 'augeas' do |pkg, settings, platform|
   pkg.build_requires "libxml2"
 
   # Ensure we're building against our own libraries when present
-  pkg.environment "PKG_CONFIG_PATH" => "/opt/puppetlabs/puppet/lib/pkgconfig"
+  pkg.environment "PKG_CONFIG_PATH", "/opt/puppetlabs/puppet/lib/pkgconfig"
 
   if platform.is_aix?
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-gcc-5.2.0-1.aix#{platform.os_version}.ppc.rpm"
     pkg.build_requires "http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/pkg-config-0.19-6.aix5.2.ppc.rpm"
-    pkg.environment "CC" => "/opt/pl-build-tools/bin/gcc"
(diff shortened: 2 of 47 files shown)


Magisus:
Travis failure is due to an internal ticket (which once again makes me think we still don't check quite the right commit range).

mckern:
> There is just one warning left, concerning the specification of a mirroring URL. However, the current alternative to register_rewrite_rule as used here is currently quite clunky, and comments in the Vanagon codebase indicate that the changes might still be under active development, so I'm leaving it for now.

Can you elaborate?

Magisus:
Looking through the code again, I'm not sure what I saw that led me to believe this API would continue changing. I'll make the updates.

Side note, I guess I'm just curious why the option to set a mirroring URL globally (that would implicitly affect all components) was removed. I imagine there must have been some drawbacks to that model, but in this case it seems like one more way for a developer writing a component to accidentally mess up by forgetting to add a mirror.

mckern:
@Magisus The TL;DR [is here](https://github.com/puppetlabs/vanagon/commit/42fd983354acad31b9a8c595989bc59597d36f80) but if I might expand on it a bit:

When I was working out what should be tightened up on the march to Vanagon 1.0.0, I split a lot of features into "good for everyone" and "good for Puppet," and then began excising things that were only good for Puppet in favor of things that were good for everyone.

Many package building tools provide the option to define a mirror URL for an archive and a canonical URL to fall back to. Homebrew is actually the project that Vanagon tracks most closely against as far as the user API is concerned, and Homebrew does this. Mirror URLs are explicit, and they are a property of the component itself as opposed to the project. This aligns with the mental model most users have for distributing source code and that most package maintainers have for building packages.

The URL rewriter was a property of the project, not the components. It was complex to maintain (as anything that effectively rewrites intent tends to be), full of subtle bugs (as anything that manipulates raw-but-specifically-formatted strings tends to be), and not every kind of source artifact had access to it (for a long time, we could not/would not rewrite Git URLs). If you were unaware that it was there, then it became a sort of "magic number" if you tried to build P-A or any other project that had a hardcoded rewrite rule in the project.

So, while the replacement feature is ultimately more verbose, it is also significantly more explicit and more intuitive for a user who is getting started with a Vanagon project or who may have inherited a Vanagon project from another maintainer.

I hope that provides some helpful background. If you've got any questions or want to talk it through, @melissa or @underscorgan or even @shrug should be pretty well versed here. I'm kind of a lurker these days since I lost all my commit bits.

Magisus:
Cool, thanks for the background!

Magisus:
I think the only ones left now are these from Windows:
```
You're trying to set the mode, owner, or group for windows. I don't know how to do that, ignoring!
```
Not sure that's worth special casing to avoid the warning. So long as we know trying to do that does nothing on Windows.

Magisus:
We can merge this after https://github.com/puppetlabs/puppet-agent/pull/1110 gets in.

Magisus:
Green full matrix build: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent_pkg-van-ship_daily-ad-hoc/53/

test.rb
```diff
@@ -0,0 +1 @@
+pkg.environment "THING", "value"
```
McdonaldSeanp: What is this file for?

Magisus: Oh, derp, I was using it to test my Ruby rewriter script. I wasn't crazy enough to redo all those environment hashes by hand. :P I'll delete it.

PR #1102, opened 2017-06-27
(PA-1258) Delete pre-vanagon windows logic and configs

joshcooper:
Prior to vanagon having support for windows, the puppet-agent repo
relied on powershell scripts and the puppet-win32-ruby and
puppet_for_the_win repos to build MSIs. Now that vanagon knows how to
build MSIs, they are no longer needed.


Magisus:
Aww yes, getting rid of cruft 👍

ScottGarman:
:+1:

PR #1103, opened 2017-06-30
(CPR-391) Fedora packages have odd distro tag

mwaggett:
This PR removes the 'f' that prefixes fedora platform versions and adds an explicit 'fc<version>' distro tag that is consistent with upstream packages.


shrug:
Should probably hold off a merge until we get back on Wednesday, just in case...

PR #1104, opened 2017-07-05
(PA-1263) Do not install curl when building solaris 10 packages

branan:
Curl is an external dependency (opencsw), and is already installed
on the builders. Not trying to re-install it every time will help
us to avoid potential issues with the opencsw package repo (such
as the one in the ticket above)


Magisus:
This sounds like a nice clean workaround. Are there other platforms where we are similarly doing unnecessary work?

branan:
@Magisus almost certainly, but auditing that will be an effort unto itself

PR #1105, opened 2017-07-06
(PA-1269) Enable EC2 master install

johnduarte:
This commit updates the aio install pre-suite to allow installing
an ec2 master dynamically on a el based system without intervention.
Due to the expectations of the nightly puppetserver repo, the nightly
puppet-agent is installed first and then over-written with the
puppet-agent version under test. The specification of a SERVER_VERSION
is also supported.

### DIFF
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -24,26 +24,62 @@ step "Install puppet-agent..." do
 end
 
 step "Install puppetserver..." do
-  if ENV['TESTING_RELEASED_PACKAGES']
-    # unlike the install_puppet_agent_* beaker methods,
-    # install_puppetlabs_release_repo_on does not use pc1 by default. Have to
-    # supply pc1 arg to install from pc1 rather than latest in pre-puppet
-    # collection repos which was version 1.2.0.
-    install_puppetlabs_release_repo_on(master, 'pc1')
+  if ENV['SERVER_VERSION'].nil? || ENV['SERVER_VERSION'] == 'latest'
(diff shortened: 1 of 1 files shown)


MosesMendoza:
Would you mind updating the README at https://github.com/johnduarte/puppet-agent/tree/59e540264f6bc87ef9607810465cd9bb0cf3c738/acceptance to include a note about SERVER_VERSION being supported?

joshcooper:
Also can we hold off on merging this until we're sure the version in puppet is correct? Based on https://github.com/puppetlabs/puppet/pull/6043#issuecomment-313401513 it sounds like it's not quite right?

geoffnichols:
Relates to https://github.com/puppetlabs/pxp-agent/pull/603.

johnduarte:
@MosesMendoza the [README](https://github.com/puppetlabs/puppet-agent/blob/master/acceptance/README.md#settings-and-environment-variables) already includes the definition of SERVER_VERSION.
I could add a README for ec2 in the same manner as [in puppet](https://github.com/puppetlabs/puppet/blob/4.10.x/acceptance/README-ec2.md) if that is desirable.

MosesMendoza:
Oops sorry @johnduarte my mistake - I read it quickly and don't know how I missed that.

joshcooper:
This works for me locally, except that by default, this switches puppet-agent to installing the nightly 5.0.x puppetserver, which is not compatible with puppet-agent 1.10.x, so the install fails. So we need to PR ci-job-configs first to `export SERVER_VERSION=2.7.2` like we do for the other component tests in the puppet-agent#1.10.x pipeline. Marking this as blocked for now.

PR #1109, opened 2017-07-09
(PA-1264) Add support to build PA against system openssl on FIPS platforms

jsane:
This is being done to allow using puppet agent on FIPS enabled systems.
Vendored openssl is not allowed on FIPS systems as it is not FIPS certified.
System openssl linking is done when building on redhat_fips-7-x86_64 platform.
The rhel7 fips vmpooler image used for building will be FIPS enabled and have 
the appropriate version of openssl-devel package installed on it.
    
Also adding redhat-fips-7-x86_64 as a build platform.

### DIFF
--- a/configs/components/cpp-pcp-client.rb
+++ b/configs/components/cpp-pcp-client.rb
@@ -11,7 +11,12 @@ component "cpp-pcp-client" do |pkg, settings, platform|
     pkg.environment "PATH", "#{settings[:bindir]}:/opt/pl-build-tools/bin:$(PATH)"
   end
 
-  pkg.build_requires "openssl"
+  if settings[:vendor_openssl] == "no"
+    pkg.build_requires 'openssl-devel'
+  else
+    pkg.build_requires 'openssl'
+  end
+
   pkg.build_requires "leatherman"
 
   if platform.is_aix?
--- a/configs/components/puppet-ca-bundle.rb
+++ b/configs/components/puppet-ca-bundle.rb
@@ -1,7 +1,12 @@
 component "puppet-ca-bundle" do |pkg, settings, platform|
   pkg.load_from_json("configs/components/puppet-ca-bundle.json")
 
-  pkg.build_requires "openssl"
+  if settings[:vendor_openssl] == "no"
+    pkg.build_requires 'openssl-devel'
+  else
+    pkg.build_requires 'openssl'
+  end
+
 
   java_available = true
   case platform.name
@@ -19,7 +24,12 @@ component "puppet-ca-bundle" do |pkg, settings, platform|
     java_available = false
   end
 
-  openssl_cmd = "#{settings[:bindir]}/openssl"
+  if settings[:vendor_openssl] == "no"
+    openssl_cmd = "/usr/bin/openssl"
+  else
+    openssl_cmd = "#{settings[:bindir]}/openssl"
+  end
+
   if platform.is_cross_compiled_linux?
     # Use the build host's openssl command, not our cross-compiled one
     openssl_cmd = "/usr/bin/openssl"
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -253,7 +253,15 @@ project "puppet-agent" do |proj|
   end
   proj.component "ruby-shadow" unless platform.is_aix? || platform.is_windows?
   proj.component "ruby-augeas" unless platform.is_windows?
-  proj.component "openssl"
+
+  if platform.name =~ /^redhat-fips-7-.*/
+    # Link against system openssl and not package openssl
+    proj.setting(:vendor_openssl, "no")
+  else
+    proj.setting(:vendor_openssl, "yes")
+    proj.component "openssl"
+  end
+
   proj.component "puppet-ca-bundle"
   proj.component "libxml2" unless platform.is_windows?
   proj.component "libxslt" unless platform.is_windows?
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -9,7 +9,12 @@ component 'curl' do |pkg, settings, platform|
     pkg.apply_patch 'resources/patches/curl/curl-7.55.1-aix-poll.patch'
   end
 
-  pkg.build_requires "openssl"
+  if settings[:vendor_openssl] == "no"
+    pkg.build_requires 'openssl-devel'
+  else
+    pkg.build_requires 'openssl'
+  end
+
   pkg.build_requires "puppet-ca-bundle"
 
   if platform.is_cross_compiled_linux?
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -15,7 +15,12 @@ component "facter" do |pkg, settings, platform|
   pkg.replaces 'pe-facter'
 
   pkg.build_requires "ruby-#{settings[:ruby_version]}"
-  pkg.build_requires 'openssl'
+  if settings[:vendor_openssl] == "no"
+    pkg.build_requires 'openssl-devel'
+  else
+    pkg.build_requires 'openssl'
+  end
+
   pkg.build_requires 'leatherman'
   pkg.build_requires 'runtime'
   pkg.build_requires 'cpp-hocon'
@@ -28,7 +33,6 @@ component "facter" do |pkg, settings, platform|
 
   # Running facter (as part of testing) expects augtool are available
   pkg.build_requires 'augeas' unless platform.is_windows?
-  pkg.build_requires "openssl"
 
   if platform.is_windows?
     pkg.environment "PATH", "$(shell cygpath -u #{settings[:gcc_bindir]}):$(shell cygpath -u #{settings[:ruby_bindir]}):$(shell cygpath -u #{settings[:bindir]}):/cygdrive/c/Windows/system32:/cygdrive/c/Windows:/cygdrive/c/Windows/System32/WindowsPowerShell/v1.0"
--- a/configs/components/pxp-agent.rb
+++ b/configs/components/pxp-agent.rb
@@ -10,7 +10,12 @@ component "pxp-agent" do |pkg, settings, platform|
     pkg.environment "PATH", "#{settings[:bindir]}:/opt/pl-build-tools/bin:$(PATH)"
   end
 
-  pkg.build_requires "openssl"
+  if settings[:vendor_openssl] == "no"
+    pkg.build_requires 'openssl-devel'
+  else
+    pkg.build_requires 'openssl'
+  end
+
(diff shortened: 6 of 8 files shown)


configs/projects/puppet-agent.rb
```diff
@@ -241,7 +241,12 @@
   end
   proj.component "ruby-shadow" unless platform.is_aix? || platform.is_windows?
   proj.component "ruby-augeas" unless platform.is_windows?
-  proj.component "openssl"
+
+  fips_build = ENV["USE_SYSTEM_OPENSSL"]
+  if fips_build != '1'
+    proj.component "openssl"
+  end
```
mckern: Hi Jayant. Since `ENV` is treated as a Ruby Hash, calling `ENV["USE_SYSTEM_OPENSSL"]` will return either `nil` ([which is `false`, along with `false`](http://www.skorks.com/2009/09/true-false-and-nil-objects-in-ruby/)) or any other object, which will then be treated as `true`. So you can simplify almost all of these conditionals:

```ruby
proj.component "openssl" unless ENV["USE_SYSTEM_OPENSSL"]
```

You could also use this to set a project-wide setting for OpenSSL so you can skip the Env. checks in all of the components and use the `Project#settings` Hash instead:

```ruby
if ENV["USE_SYSTEM_OPENSSL"]
  proj.component "openssl"
  proj.setting(:vendor_openssl, false)
else
 proj.setting(:vendor_openssl, true)
end
```

One of these approaches will probably be more flexible than specifically checking for the String `1` in every component.

jsane: Agree. Thanks for the feedback. I am relatively new to puppet-agent's build system. Will push changes incorporating your suggestions.

jsane: WIP - pushed updates not tested yet. Pl do not merge.

mckern: I believe you can add a `do not merge` label to this PR. You can also add "(WIP)" to the PR title.

jsane:
Changes tested. ok to review and merge if appropriate.

mckern:
Offhand, I'd be concerned about the implications for the vendored TLS cert bundle and any potential ramifications for PE. I'd probably also want to check all of the dynamic linkages for binaries (libraries and executable) to make sure that everything is linked as expected.

OpenSSL was vendored for very specific reasons in 2014 (I did a decent amount of that work initially), but maybe those reasons are less applicable in 2017. I've been gone for about 3 months and things change. Has anyone from the Release Engineering team (@shrug, @melissa, @underscorgan, @mwaggett, @sarameisburger, @demophoon) seen this yet?

mckern:
(Oh, and your commit subject is fine but you'll want to write up a commit message that summarizes the reason for this change.)

configs/components/cpp-pcp-client.rb
```diff
@@ -11,7 +11,12 @@
     pkg.environment "PATH", "#{settings[:bindir]}:/opt/pl-build-tools/bin:$(PATH)"
   end
 
-  pkg.build_requires "openssl"
+  if !settings[:vendor_openssl]
+    pkg.build_requires 'openssl-devel'
```
underscorgan: You might need to do some switching based on OS, this isn't going to be the package name across all the supported OSes (or if it only supports EL*, we should fail if this is set for those platforms)

configs/components/puppet-ca-bundle.rb
```diff
@@ -19,7 +23,12 @@
     java_available = false
   end
 
-  openssl_cmd = "#{settings[:bindir]}/openssl"
+  if !settings[:vendor_openssl]
+    openssl_cmd = "/usr/bin/openssl"
```
underscorgan: Is this path consistent across platforms?

configs/projects/puppet-agent.rb
```diff
@@ -241,7 +241,14 @@
   end
   proj.component "ruby-shadow" unless platform.is_aix? || platform.is_windows?
   proj.component "ruby-augeas" unless platform.is_windows?
-  proj.component "openssl"
+
+  if ENV["USE_SYSTEM_OPENSSL"]
```
underscorgan: I'm not sure this is quite doing what you want. This will evaluate to `true` even if USE_SYSTEM_OPENSSL is set to false, which seems like it could lead to unintended results.

```
morgan@river.corp.puppetlabs.net % FOO=true BAR=false irb                                                                               (git)-[remove_utopic]
irb(main):001:0> if ENV['BAR']
irb(main):002:1> puts 'true'
irb(main):003:1> end
true
```

mckern: Oh yeah, we never plumbed in "coerce these to pure booleans", did we?

configs/components/cpp-pcp-client.rb
jsane: Since this is primarily meant for allowing PA for FIPS, I will have limit to just RHEL family. It will fail for other OS's. Unless we want it to silently ignore the environment variable.

configs/components/puppet-ca-bundle.rb
jsane: This actually came from the cross-compiled linux case, that may have had similar issue I guess, where we want to use the host system openssl instead of using puppetlabs's openssl (which gets used with #{settings[:bindir]}/openssl). Also since now it is limited to el/fedora's this should be ok but can change it to make it platform based if desired.

configs/projects/puppet-agent.rb
jsane: One of the prior versions I submitted explicitly checked the environment variable's value for "1" but had changed it per some earlier suggestion/comments then.

configs/components/cpp-pcp-client.rb
jsane: I have adjusted the commit message to include the reason for the change. 

Yes I have been aware of the reasons for vendor-ing openssl and we are not intending to undo/remove that broadly. This is only for generating fips compliant builds for certain targets - would be very conscious.

configs/projects/puppet-agent.rb
```diff
@@ -241,7 +241,20 @@
   end
   proj.component "ruby-shadow" unless platform.is_aix? || platform.is_windows?
   proj.component "ruby-augeas" unless platform.is_windows?
-  proj.component "openssl"
+
+  if ENV["USE_SYSTEM_OPENSSL"] == '1'
```
mckern: @underscorgan @jsane I provided this guidance last time; I thought we'd added support for coercing those values but I realize now that was in the `packaging` repo, not Vanagon so it won't work here. I also expected that this value would _only_ ever be set in the positive, e.g. "USE_SYSTEM_OPENSSL=true bundle exec build", never "USE_SYSTEM_OPENSSL=false bundle exec build". YMMV, but my opinion is that using an extremely specific value is probably a bad idea (I can get into why but it boils down to Shell UX being _*dodgy*_), but here's a compromise (which uses an anonymous function, which I don't love but I could live with) that I think will work OK:

```ruby
def use_system_openssl?
  # explicitly catch the expected "don't use 
  return false if ENV["USE_SYSTEM_OPENSSL"] =~ /false/f/0/i
  !!ENV["USE_SYSTEM_OPENSSL"]
end

if use_system_openssl?
  # do the thing
end
```

configs/projects/puppet-agent.rb
```diff
@@ -241,7 +241,20 @@
...
+    if (platform.is_el? || platform.is_fedora?)
+      # Currently we plan to support linking against system openssl
+      # only on these platforms.
+      proj.setting(:vendor_openssl, false)
+    else
+      fail "Currently linking against system openssl is supported only on RHEL family platforms"
```
mckern: Nitpick: I'd explicitly state "Enterprise Linux or Fedora platforms". Abbreviations are OK in the code paths but the UX should be as explicitly precise as possible.

configs/projects/puppet-agent.rb
```diff
@@ -241,7 +249,20 @@
   end
   proj.component "ruby-shadow" unless platform.is_aix? || platform.is_windows?
   proj.component "ruby-augeas" unless platform.is_windows?
-  proj.component "openssl"
+
+  if use_system_openssl?
+    if (platform.is_el? || platform.is_fedora?)
```
underscorgan: rubocop doesn't like these parens

underscorgan: is this going to work reasonably on all of the fedora/el platforms we support or do we need to limit scope more? Fedora we only have relatively recent, and it doesn't look like any of those have moved to openssl 1.1.0 yet, but will PA work with that? Will it work with 0.9.8 installed on EL5?

Just trying to make sure we're explicit where we need to be.

jsane:
Sorry this languished far too long. Primarily, apart from other reasons, due to challenges in testing my custom agent builds on some of the platforms we intend to support this on. 
I plan to limit it to EL 6, 7 and Fedora 24, 25. Would you be amenable to merging in with that? With the changes checked in, I plan to run them thru various CI pipelines; whatever testing I had been doing and/or would have done say on some of these targets was still going to be limited. If there are any failures in CI that appear to be due to these changes, then I can do the needful.

Unless if it is possible to run my custom builds, generated locally, thru all relevant CI tests I will do that but don't know at this point.

jsane:
I am in process of testing it in CI - will update with job details. Pl hold off merging (re: above comment).

jsane:
Latest set of updates. One of the main changes is switch from using a boolean for the dictionary item "vendor_openssl" to string type since boolean does not work. 
The changes have been tested on rhel7, rhel6 thru PA CI but that required this PR https://github.com/puppetlabs/puppet-agent/pull/1218 (related to PA-1580) to workaround issues with rhel7 images used by CI/vmpooler. The "Create PEZ build" in step 9 however fails for reasons unknown; there was no response about its cause and need in Jenkins room. Will deal with it if it does become an issue. Attached find logs of successful CI run.   
Fedora 24, 25 will likely work but experiencing some transience where one attempt/build succeeded till step 9 for both f24/f25 but later failed at Step 6 for f24. Last CI build aborted due to lack for rhel7 instances. Will attach logs as soon as that tests can be completed.  
[Openssl-PA-CI-rhel6and7-logs.tar.gz](https://github.com/puppetlabs/puppet-agent/files/1369869/Openssl-PA-CI-rhel6and7-logs.tar.gz)

jsane:
Pushed latest changes - built & tested in CI on rhel7 fips platforms

jsane:
Link to the latest successful adhoc PA CI run when building against the new fips platform: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/79/

PR #1110, opened 2017-07-12
(PA-611) Fix the libselinux tarball md5 to match upstream

ScottGarman:
Our locally mirrored copy of the 2.0.94 libselinux tarball has a different md5 from the upstream one because we had unpacked/recompressed it. It's time to update our local mirror version to match the upstream one and fix this. 

Note: let me know right before you're ready to merge this and I can make the file switchover, otherwise builds will be broken.


Magisus:
This should be retargeted at the 1.10.x branch.

PR #1111, opened 2017-07-12
(PA-611) Fix md5 for libselinux tarball and nssm zip archive

ScottGarman:
This updates the md5 to reflect what the upstream tarball has. Once this is merged I'll overwrite the file on buildsources (after making a backup of it).


Magisus:
These now match what is on buildsources 👍

PR #1121, opened 2017-07-27
(PA-1320) rbconfig fixes for Power8 platforms

ScottGarman:
(This is a PR for the 5.0.x branch, which uses ruby 2.4.1. The 1.10.x commits can be safely merged up to this branch when we need to do that)

The previous rbconfig files included for Power8 platforms were badly broken; I had taken them from the packaged versions of ruby for the OS, instead of rebuilding from sources. This PR includes rbconfig files from a native compile from source I've performed on the target, which is what I should have done in the first place.

I've tested this by rebuilding the RHEL and Ubuntu agents with these fixes, and they now successfully show installed gems and I can require the gems using our irb command.

Additionally, I can install a gem with native extensions (sqlite3) with our gem command.


ScottGarman:
I think there may be a second issue to fix with installing gems with native extensions. Don't merge this just yet.

ScottGarman:
All issues are now resolved and this is ready for final review and merge.

branan:
:+1: from me

PR #1122, opened 2017-07-27
(maint) Remove redundant PATH envvar in curl config

johnmccabe:


MikaelSmith:
I tried to do a resolve and now there are no changes. Something weird happened.

caseywilliams:
With #1378 merged, the curl component is now built as part of [puppet-runtime](https://github.com/puppetlabs/puppet-runtime) and consumed in a tarball as part of the puppet-runtime component. I do notice that this path duplication is repeated in that project, though - I can make a quick PR to remove it.

PR #1123, opened 2017-07-30
(PA-1339) Add Debian 9 "Stretch" x86 platforms

ScottGarman:
It blends, I've done pretty thorough smoke testing with the agent packages this can generate.


ScottGarman:
The Travis check is because BUILD-17 is not a public ticket. I've checked the ticket settings and the security level is set to None, so I'm not sure why the ticket is private.

McdonaldSeanp:
👍

PR #1131, opened 2017-08-14
(packaging) Pin cpp-pcp-client and pxp-agent to 5.0.1 release

branan:
We are not shipping these components as part of the 5.1.0 release


McdonaldSeanp:
👍

PR #1133, opened 2017-08-14
(PA-1406) Enable DEP support in Windows version of Puppet agent binaries

jsane:
Enable nxcompat and dynamicbase flags in Windows builds of all puppet and
3rd party exe's, dll's built using puppet-agent.
This is being done to meet some customer's security audit requirements.

**Note**: In addition to this change, git sha references in following component's project files (json files) need to be updated appropriately so as to enable DEP linker settings: cpp-pcp-client, facter, leatherman, pxp-agent. 
Refer to following PRs: 
https://github.com/puppetlabs/facter/pull/1624
https://github.com/puppetlabs/cpp-pcp-client/pull/228
https://github.com/puppetlabs/pxp-agent/pull/620
https://github.com/puppetlabs/leatherman/pull/247

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -183,7 +183,7 @@ project "puppet-agent" do |proj|
     proj.setting(:tools_root, "C:/tools/pl-build-tools")
     proj.setting(:cppflags, "-I#{proj.tools_root}/include -I#{proj.gcc_root}/include -I#{proj.includedir}")
     proj.setting(:cflags, "#{proj.cppflags}")
-    proj.setting(:ldflags, "-L#{proj.tools_root}/lib -L#{proj.gcc_root}/lib -L#{proj.libdir}")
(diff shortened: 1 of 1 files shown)


MikaelSmith:
This could be done for all CMake-related projects by setting CMAKE_EXE_LINKER_FLAGS and CMAKE_SHARED_LINKER_FLAGS. Not sure it makes sense to do in all the individual projects.

jsane:
I am trying to remember if I tried doing that as well (and it did not work?). But I can check again and if it works then we can do it properly once we are past this deadline. btw, CMAKE_SHARED_LINKER_FLAGS definitely does not work for dll's (link_libraries does as I found out).

MikaelSmith:
CMAKE_SHARED_LINKER_FLAGS should work for dll's, strange. Well, most of the PRs are merged.

jsane:
You are right; thats what I thought and had tried first but it won't (on cpp-pcp-client).

PR #1134, opened 2017-08-16
(packaging) Set component tags for 5.1.0 release

branan:


configs/components/cpp-pcp-client.json
```diff
@@ -1 +1 @@
-{"url": "git://github.com/puppetlabs/cpp-pcp-client.git", "ref": "c929ce5eb6dccd37dccb7db1f71189a330d16a9d"}
+{"url": "git://github.com/puppetlabs/cpp-pcp-client.git", "ref": "refs/tags/1.5.3"}
```
Magisus: This tag and SHA are not equivalent, but I think I saw some chatter that we're intentionally pinning back?

McdonaldSeanp: Yes

PR #1136, opened 2017-08-16
(PA-1414) Update nssm to be build with msbuild

McdonaldSeanp:


configs/components/nssm.rb
```diff
@@ -1,14 +1,19 @@
...
-  pkg.environment "PATH" => "/cygdrive/c/ProgramData/chocolatey/tools:$$PATH" if platform.is_windows?
+  # pkg.environment "PATH" => "/cygdrive/c/ProgramData/chocolatey/tools:$$PATH" if platform.is_windows?
 
-  win_arch = platform.architecture == "x64" ? "win64" : "win32"
+  win_arch = platform.architecture == "x64" ? "x64" : "Win32"
 
-  pkg.install_file "nssm-2.24/#{win_arch}/nssm.exe", "#{settings[:service_dir]}/nssm.exe"
+  pkg.install do
+    [
+      "/cygdrive/c/Program\\ files\\ \\(x86\\)/Microsoft\\ Visual\\ Studio/2017/BuildTools/Common7/Tools/vsdevcmd.bat",
```
Iristyle: The `.bat` adds `msbuild` to path, and also configures all the other paths needed to find things properly.... but as mentioned, it will have to be run in the same process as `msbuild` itself.

configs/components/nssm.rb
```diff
@@ -1,14 +1,19 @@
...
+      "/cygdrive/c/Program\\ Files\\ \\(x86\\)/Microsoft\\ Visual\\ Studio/2017/BuildTools/MSBuild/15.0/Bin/msbuild.exe nssm.vcxproj /detailedsummary /p:Configuration=Release /p:Platform=#{win_arch} /p:PlatformToolset=v141 /p:TargetPlatformVersion=8.1",
```
Iristyle: By calling `vsdevcmd.bat`, the path to `msbuild` should not be qualified.

configs/platforms/windows-2012r2-x64.rb
```diff
@@ -9,10 +9,12 @@
   plat.add_build_repository "http://nexus.delivery.puppetlabs.net/service/local/nuget/nuget-pl-build-tools/"
 
   # We don't want to install any packages from the chocolatey repo by accident
+  plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe update -y chocolatey"
```
Iristyle: We can now remove this since we have `0.10.7` in Nexus

configs/platforms/windows-2012r2-x64.rb
```diff
@@ -9,10 +9,12 @@
...
   plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe sources remove -name chocolatey"
 
   plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe install -y mingw-w64 -version 5.2.0 -debug"
   plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe install -y Wix310 -version 3.10.2 -debug -x86"
+  plat.provision_with "cmd.exe /C C:/ProgramData/chocolatey/bin/choco.exe install msvc.2017-win8.1.sdk.en-us -y --cache-location=\"C:\\msvc\""
```
Iristyle: Executed this way, is the `cmd.exe /C` necessary?  Or given it's in cygwin can use something like `--cache-location=/cygdrive/c/msvc`?

configs/components/nssm.rb
McdonaldSeanp: Yeah, the way this is set up will do that.

When vanagon goes to actually execute the commands in the pkg.install array, it will do so by joining each command with a `&&` so all commands in this array will be in the same process.

configs/platforms/windows-2012r2-x64.rb
McdonaldSeanp: As we discussed this morning, updating everything to 0.10.7 broke builds, so we should keep this here until we figure out how to fix things.

configs/platforms/windows-2012r2-x64.rb
McdonaldSeanp: I think you're right, we can likely move to that.

configs/platforms/windows-2012r2-x64.rb
```diff
@@ -9,10 +9,12 @@
   plat.add_build_repository "http://nexus.delivery.puppetlabs.net/service/local/nuget/nuget-pl-build-tools/"
 
   # We don't want to install any packages from the chocolatey repo by accident
+  plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe update -y chocolatey"
   plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe sources remove -name chocolatey"
 
   plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe install -y mingw-w64 -version 5.2.0 -debug"
   plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe install -y Wix310 -version 3.10.2 -debug -x86"
+  plat.provision_with "C:/ProgramData/chocolatey/bin/choco.exe install msvc.2017-win8.1.sdk.en-us -y --cache-location=\"C:\\msvc\""
```
Iristyle: Can we add a comment about why we're using `--cache-location`?

configs/components/nssm.rb
```diff
@@ -1,14 +1,14 @@
...
-  win_arch = platform.architecture == "x64" ? "win64" : "win32"
+  pkg.install do
+    [
+      "/cygdrive/c/Program\\ files\\ \\(x86\\)/Microsoft\\ Visual\\ Studio/2017/BuildTools/Common7/Tools/vsdevcmd.bat",
+      "/cygdrive/c/Program\\ Files\\ \\(x86\\)/Microsoft\\ Visual\\ Studio/2017/BuildTools/MSBuild/15.0/Bin/msbuild.exe nssm.vcxproj /detailedsummary /p:Configuration=Release /p:OutDir=./out /p:Platform=#{build_arch} /p:PlatformToolset=v141 /p:TargetPlatformVersion=8.1",
+    ]
+  end
 
-  pkg.install_file "nssm-2.24/#{win_arch}/nssm.exe", "#{settings[:service_dir]}/nssm.exe"
+  pkg.install_file "out/nssm.exe", "#{settings[:service_dir]}/nssm.exe"
```
Iristyle: What are we going to do with the `pdb`?

configs/components/nssm.rb
```diff
@@ -1,14 +1,14 @@
...
-  # We don't automatically have this set on windows, unfortunately. We need to set the
-  # path to have access to 7za.
-  pkg.environment "PATH" => "/cygdrive/c/ProgramData/chocolatey/tools:$$PATH" if platform.is_windows?
+  build_arch = platform.architecture == "x64" ? "x64" : "Win32"
 
-  win_arch = platform.architecture == "x64" ? "win64" : "win32"
+  pkg.install do
+    [
+      "/cygdrive/c/Program\\ files\\ \\(x86\\)/Microsoft\\ Visual\\ Studio/2017/BuildTools/Common7/Tools/vsdevcmd.bat",
+      "/cygdrive/c/Program\\ Files\\ \\(x86\\)/Microsoft\\ Visual\\ Studio/2017/BuildTools/MSBuild/15.0/Bin/msbuild.exe nssm.vcxproj /detailedsummary /p:Configuration=Release /p:OutDir=./out /p:Platform=#{build_arch} /p:PlatformToolset=v141 /p:TargetPlatformVersion=8.1",
```
Iristyle: A few things here:

* `msbuild.exe` shouldn't be fully qualified
* Can we use the env vars for `%ProgramFiles(x86)%` location?  I'm actually surprised that works on the 32-bit host as `vsdevcmd` is under `%ProgramFiles%` there.
* Can we parameterize `PlatformToolset` and `TargetPlatformVersion` and move them up in the file by `build_arch`?
* Note that the the binary is auto-versioned with `version.cmd` in the nssm repo -
 so we can't use `/p:ApplicationVersion=" here and should instead rely on the `version.cmd` behavior I think (which uses `git describe` - https://github.com/puppetlabs/nssm/blob/master/version.cmd)

configs/platforms/windows-2012r2-x64.rb
Iristyle: I *think* one alternative here is to use the newest choco and install `7zip.commandline` from our choco feed at http://nexus.delivery.puppetlabs.net/service/local/repositories/temp-build-tools/content/7zip.commandline/9.20.0.20150210/7zip.commandline-9.20.0.20150210.nupkg

That will get `7za` back into PATH, which is I believe what broke the builds.

Iristyle:
Follow-on cleanup PR is at https://github.com/puppetlabs/puppet-agent/pull/1139

PR #1137, opened 2017-08-17
(MAINT) Pin Chocolatey to Version 0.9.9.11

jcoconnor:
The latest version of chocolatey has changed the 7za command to 7za
which causes nssm extract to fail.

This is a temporary fix until a vanagon fix to make the local url
handler (unzip) more robust.

### DIFF
--- a/configs/platforms/windows-2012r2-x64.rb
+++ b/configs/platforms/windows-2012r2-x64.rb
@@ -3,8 +3,11 @@ platform "windows-2012r2-x64" do |plat|
 
   plat.servicetype "windows"
 
+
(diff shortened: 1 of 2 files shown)


jcoconnor:
Adding @Iristyle as reviewer as it looks like ```chocolatey``` was upgrade in connection with another issue.

Iristyle:
We've removed 0.10.7 from Nexus for now... let's close this as I think we have better ways of addressing the problem than adding another new script to buildsources.

FWIW - I'm not a fan of the scripts at http://buildsources.delivery.puppetlabs.net/windows/ - I'd like to see them incorporated into the actual project. By putting them up on a shared drive that makes them immutable because we don't know what projects consume them and what might break if we need to change them due to inflexibility.

PR #1140, opened 2017-08-21
(PA-1427) Configure logging on nssm for pxp-agent

MikaelSmith:
Configure logging with nssm to capture any errors that happen during
startup (i.e. before pxp-agent has configured logging to file).


MikaelSmith:
Doing a build to make sure this works.

MikaelSmith:
Should be good to go.

McdonaldSeanp:
👍

PR #1142, opened 2017-08-24
(maint) double-escape NSSM outdir

branan:


branan:
Closing in favor of #1143

PR #1144, opened 2017-08-24
(PA-1408) Fix test failures with pxp-agent

MikaelSmith:
pxp-agent requires that the modules directory exists, so restore
creating it. Removing it caused test failures on all non-Windows
platforms, as pxp-agent would fail to start.


McdonaldSeanp:
👍

MikaelSmith:
Changed this to just restoring the "modules" directory. Curl is going to take some more work, so going to plan that out and revert the pxp-agent change that required it.

PR #1145, opened 2017-08-24
(PA-1447) Build curl on all platforms

MikaelSmith:
Supports pxp-agent work that requires curl.

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -4,6 +4,11 @@ component 'curl' do |pkg, settings, platform|
   pkg.url "https://curl.haxx.se/download/curl-#{pkg.get_version}.tar.gz"
   pkg.mirror "http://buildsources.delivery.puppetlabs.net/curl-#{pkg.get_version}.tar.gz"
 
+  if platform.is_aix?
+    # Patch to disable _ALL_SOURCE when including select.h from multi.c. See patch for details.
+    pkg.apply_patch 'resources/patches/curl/curl-7.55.1-aix-poll.patch'
+  end
+
   pkg.build_requires "openssl"
   pkg.build_requires "puppet-ca-bundle"
 
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -1,6 +1,6 @@
 component 'curl' do |pkg, settings, platform|
-  pkg.version '7.51.0'
-  pkg.md5sum '490e19a8ccd1f4a244b50338a0eb9456'
+  pkg.version '7.55.1'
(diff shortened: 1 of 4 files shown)


MikaelSmith:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent_init-van-int_suite-daily-ad-hoc/ job #102

MikaelSmith:
I've also done local builds with Solaris 10, AIX 7.1, and EL 7 that all look good.

MikaelSmith:
Almost: cisco-wrlinux-7 and osx failed. I have a fix for osx, still figuring out cisco.

MikaelSmith:
Updated to use a newer version of curl, which fixed the LDFLAGS issue I was running into on cisco-wrlinux-7. Build at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent_pkg-van-ship_daily-ad-hoc/89/.

MikaelSmith:
I believe this is ready to go. Tests failed because Jenkins was unable to load the Digest module, but that's a separate problem we'll need to resolve before merging https://github.com/puppetlabs/pxp-agent/pull/634.

configs/components/curl.rb
```diff
@@ -1,24 +1,31 @@
...
+  pkg.version '7.55.1'
+  pkg.md5sum '3b832160a8c9c40075fd71191960307c'
   pkg.url "https://curl.haxx.se/download/curl-#{pkg.get_version}.tar.gz"
   pkg.mirror "http://buildsources.delivery.puppetlabs.net/curl-#{pkg.get_version}.tar.gz"
 
+  if platform.is_aix?
+    # On AIX, sys/poll.h defines event and revent when _ALL_SOURCE is 1. CURL sets _ALL_SOURCE,
+    # but those collide with definitions in multi.h, causing multi.c to fail because it includes
+    # sys/poll.h (via select.h) and thereby redefines names declared in multi.h.
+    # Patch to disable _ALL_SOURCE when including select.h from multi.c.
```
McdonaldSeanp: Because there's a longer description in the patch, can you actually just write:

```
# Patch to disable _ALL_SOURCE when including select.h from multi.c. see patch for details.
```

in the comments, instead of the whole reasoning again? I actually prefer shorter comments in the component configs (since the component configs can get very large)

MikaelSmith: I'll update the comment, then merge.

PR #1147, opened 2017-08-28
(PA-1452) Add SLES 12 Power8 (ppc64le) as a supported platform

ScottGarman:
I have successfully built an agent with these changes, and validated it on a SLES 12 Power8 server (including the Ruby environment).


McdonaldSeanp:
👍

ScottGarman:
I've also added a commit to enable these builds in nightlies.

PR #1151, opened 2017-08-31
(PA-1460) Add libwhereami as a facter component

caseywilliams:
No need to merge yet - still testing builds.


branan:
:+1: as soon as the test builds finish green!

PR #1157, opened 2017-08-31
(PA-1462) rbconfig fixes for RHEL 7 Power8

ScottGarman:
This resolves problems with RHEL 7 Power8 agents being unable to
build native gem extensions.

I've tested this by building an agent package and smoke testing the agent, including its ruby environment and testing building a gem with native extensions (sqlite3).


ScottGarman:
Travis is complaining as this is a private PE ticket.

PR #1158, opened 2017-09-01
(PUP-7881) Update gettext to fix Japanese error output

Magisus:
This updates gettext-setup to a version where English is the initial
state for the locale, rather than a random locale based on the order in
which translation files were found.


Magisus:
Ugh, how have we still not gotten range of commits checked by Travis right? :( The commit with the failing message is not part of this PR.

ScottGarman:
md5 matches what I get when I download the gem from buildsources. :+1:

PR #1160, opened 2017-09-01
(PUP-7881) Bump gettext-setup to version 0.28

Magisus:
This update includes a fix for incorrect locale loading.


ScottGarman:
md5 matches the gem file I can download from buildsources. :+1:

PR #1163, opened 2017-09-06
(PA-1362) Add fedora 26 definition to the agent

McdonaldSeanp:
This commit adds support for fedora 26 to the agent. The only abnormal change
made to make things work was to specify that pl-binutils was _not_ a
requirement for openssl builds on fedora 26 in particular

### DIFF
--- a/configs/platforms/fedora-f26-x86_64.rb
+++ b/configs/platforms/fedora-f26-x86_64.rb
@@ -1,0 +1,10 @@
+platform "fedora-f26-x86_64" do |plat|
+  plat.servicedir "/usr/lib/systemd/system"
+  plat.defaultdir "/etc/sysconfig"
+  plat.servicetype "systemd"
+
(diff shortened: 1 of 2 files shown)


configs/platforms/fedora-f26-x86_64.rb
```diff
@@ -0,0 +1,10 @@
+platform "fedora-f26-x86_64" do |plat|
+  plat.servicedir "/usr/lib/systemd/system"
+  plat.defaultdir "/etc/sysconfig"
+  plat.servicetype "systemd"
+
+  plat.provision_with "/usr/bin/dnf install -y --best --allowerasing autoconf automake rsync gcc make rpmdevtools rpm-libs"
```
caseywilliams: @McdonaldSeanp you already noticed this while we were talking earlier, but for posterity: gcc won't need to be installed here

McdonaldSeanp: This uncovered a separate issue, so gcc will need to be taken out in a new PR: See https://tickets.puppetlabs.com/browse/PA-1487 for details

PR #1165, opened 2017-09-12
(PA-1504) Change NSSM versioning scheme

Iristyle:
- Update NSSM code so that it now versions by

   Major.Minor.Patch.CommitsPastTag

   This is reproducible within our build system


Iristyle:
I think we'll need to actually cut a `1.10.8` release branch based on `1.10.7` and retarget this PR at that branch.

This PR also has a temporary ref to my fork and a SHA, instead of the `puppetlabs` fork and the eventual `2.24.2` tag.

Iristyle:
Closing given the NSSM component will auto-promote

PR #1168, opened 2017-09-12
(PA-1498) Update pl-boost for aix 7.1

ScottGarman:
This updates the version of pl-boost to one which includes a patch
we've applied to fix boost::filesystem::unique_path failures.

### DIFF
--- a/configs/components/leatherman.rb
+++ b/configs/components/leatherman.rb
@@ -15,7 +15,11 @@ component "leatherman" do |pkg, settings, platform|
     pkg.build_requires "pl-cmake"
   elsif platform.is_aix?
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-cmake-3.2.3-2.aix#{platform.os_version}.ppc.rpm"
-    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-boost-1.58.0-1.aix#{platform.os_version}.ppc.rpm"
+    if platform.name =~ /aix-7.1/
+      pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-boost-1.58.0-7.aix#{platform.os_version}.ppc.rpm"
+    else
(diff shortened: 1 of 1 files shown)


configs/components/leatherman.rb
```diff
@@ -15,7 +15,11 @@
     pkg.build_requires "pl-cmake"
   elsif platform.is_aix?
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-cmake-3.2.3-2.aix#{platform.os_version}.ppc.rpm"
-    pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-boost-1.58.0-1.aix#{platform.os_version}.ppc.rpm"
+    if platform.name =~ /aix-7.1/
```
MikaelSmith: Is this package also used on AIX 7.2?

MikaelSmith:
Have you validated this in a puppet-agent build somewhere?

configs/components/leatherman.rb
ScottGarman: Sorry, that was a typo. I've updated the commit header and PR title to reflect that it's for aix 7.1.

MikaelSmith: Do we need separate code to support AIX 7.2? Because that also appears to be failing: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/master/view/Non-Standard/job/platform_puppet-agent_intn-van-sys_suite-secondary-pxp-agent-master/10/SLAVE_LABEL=beaker,TEST_TARGET=aix72-POWERfa/testReport/(root)/tests_tasks/task_download_rb/

ScottGarman: No, I just confirmed with Branan that the packages we build for aix 7.1 are what we use for 7.2 targets as well. In the p-a platform list you'll see there is only a file for aix 7.1, so this code should work as intended.

ScottGarman:
I've tried testing this in the ad-hoc pipeline but we have other problems with the repo command not being found by bundler:

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent_pkg-van-repo_daily-ad-hoc/88/console

Given the time-sensitive nature of this work, if we merge this now and find problems we can revert it if necessary, and I'm pretty sure there's very little risk of this PR causing problems.

MikaelSmith:
I'm ok with that.

PR #1175, opened 2017-09-14
(PA-1527) Allow non-root agents reading modules

MikaelSmith:
Relax permissions on modules dir so non-root agents can read installed
modules, such as pxp-module-puppet. Specific modules could still use
more restrictive permissions to prevent access.


MikaelSmith:
Verbal thumbs-up on this change from @adreyer.

PR #1179, opened 2017-09-15
(PCP-511) Add hocon dependency for pxp-agent

MikaelSmith:
Ensures hocon will be built before pxp-agent when we add hocon support.


MikaelSmith:
Should be merged before https://github.com/puppetlabs/pxp-agent/pull/660. Harmless on its own.

PR #1166, opened 2017-09-12
(PA-1502) Set NoAppConsole=1 for Win-10 Creators

jcoconnor:
Win-10 Creators has an issue with nssm startup that is address by
setting NoAppConsole=1

PR #1180, opened 2017-09-15
(PA-1502) Set NoAppConsole=1 for Win-10 Creators

jcoconnor:
Win-10 Creators has an issue with nssm startup that is address by
setting NoAppConsole=1


jcoconnor:
This replaces #1166

PR #1195, opened 2017-09-21
(PA-1466) Exclude dmidecode component for Power8 architectures

ekinanp:
Previously when running facter on a machine with a Power8
architecture, dmesg would show the following line of error:
Program dmidecode tried to access /dev/mem between f0000->100000

The fix is to exclude dmidecode for Power8 architectures.


ekinanp:
I tested this on the relevant OSes outlined in the ticket, and it looks like this works.

ekinanp:
@geoffnichols Done.

PR #1199, opened 2017-09-24
(PA-1547) Update ruby to address vulnerabilities fixed in v2.4.2

ekinanp:


McdonaldSeanp:
The changeset looks good, but we'll need an adhoc run with building testing of at least the following:

aix 7.1
ciscowrlinux 7
EL 7
debian 9
osx 10.12
windows x64 ( you can test on just windows 2012 for the single windows package)

ekinanp:
Ran an ad-hoc build with the following OSes
{code}

{code}

PR #1201, opened 2017-09-26
Merge 5.3.0_release -> 5.3.x

caseywilliams:


branan:
I'd have expected at least some of the tags to roll forward without conflicts. How'd we end up with no changes in this PR?

PR #1205, opened 2017-09-29
(PA-1555) Set component tags for 5.3.1 release

caseywilliams:
Sets tags for updated facter 3.9.2 and puppet 5.3.1.

Also pins pxp-agent back to 1.8.0. The diff between 1.8.0 and the version of 1.8.x that went through the pipelines today is [just an acceptance change](https://github.com/puppetlabs/pxp-agent/compare/1.8.0...1.8.x), and doesn't seem worth tagging a 1.8.1 for.


branan:
:+1:

PR #1207, opened 2017-10-02
(maint) Build libwhereami on Solaris and AIX

caseywilliams:
libwhereami#master now has hypervisor detectors for Solaris and AIX; this commit builds libwhereami on those platforms - I was able to build the agent successfully on both.


ScottGarman:
:+1:

PR #1210, opened 2017-10-02
(PA-1547) Update Ruby 2.3.3 -> 2.3.5, 2.4.1 -> 2.4.2

ekinanp:


ekinanp:
@McdonaldSeanp Thanks for pointing that out. The 2.3.5 files also needed the change. About to run the ad-hoc build now.

ekinanp:
Updates:

--RUBY 2.3.5 AD-HOC:

(1) Failure on Windows 2012r2-x86_64. This was due to the fact that LEATHERMAN_RUBY was searching for ruby235.dll when it is ruby230.dll instead. The proposed fix is to change RUBY_SO_NAME to x64-msvcrt-$(RUBY_BASE_NAME)$(MAJOR)$(MINOR)0 in the mingw rbconfig for Ruby 2.3.5.

--RUBY 2.4.2 AD-HOC:

(1) Failure on el-7-x86_64. Here is the relevant part of the output from the Jenkins job:
`21:12:37 /opt/pl-build-tools/bin/ruby --disable=gems -I/var/tmp/tmp.YYYlfFGJGz/ruby-2.4.2 -rpowerpc64le-linux-fake  -I`cd ./lib; pwd` --disable-gems -r./powerpc64le-linux-fake ./tool/rbinstall.rb --make="/usr/bin/make" --dest-dir="" --extout=".ext" --mflags="-w" --make-flags="w" --data-mode=0644 --prog-mode=0755 --installed-list .installed.list --mantype="doc"
21:12:37 ./tool/rbinstall.rb:23:in `require': /var/tmp/tmp.YYYlfFGJGz/ruby-2.4.2/lib/ostruct.rb:191: syntax error, unexpected '.' (SyntaxError)
21:12:37     @table&.key?(mname) || super`

^This failure is because of the "&." syntax, which was introduced in Ruby 2.3.0. pl-build-tools uses Ruby 2.1.9 based on what's in the pl-build-tools-vanagon repo. The reason we did not see this failure on Ruby 2.4.1 is because that part of the code did not have the "&." operator in it. So my guess is there's two ways to fix this:

(1) Update pl-build-tools-vanagon to use Ruby 2.3.0 instead OR
(2) Patch lib/ostruct.rb to remove that use of the "&." operator.

I think I will choose Option (2), since it seems like the simplest fix.

Both ad-hocs also failed to build on AIX 7.1 due to some configuration error (sizeof(long long) cannot be computed) -- haven't been able to look into this yet because I could not build it locally on my machine due to the demand on the AIX builder VMs.

Anyways, will update the PR shortly to introduce the patch for lib/ostruct.

ekinanp:
Patch for lib/ostruct.rb has been added.

ekinanp:
The AIX 7.1 failures are caused by a change in the configure.in file from Ruby 2.3.4 => 2.3.5 and Ruby 2.4.1 => 2.4.2. Specifically, the following error occurs when running the command `bash configure --enable-shared --enable-bundled-libyaml --disable-install-doc --disable-install-rdoc --prefix=/opt/puppetlabs/puppet --with-opt-dir=/opt/puppetlabs/puppet --build=powerpc-ibm-aix7.1.0.0` after running `autoconf`:

`checking size of long... 0
checking size of long long... configure: error: in `/Users/enis.inan/Downloads/ruby-2.3.5':
configure: error: cannot compute sizeof (long long)`

It looks like the commits that modified configure.in are (in order of most recent):

(1) https://github.com/ruby/ruby/commit/edda79257537a37a679b7c3a066243ff4123b0d5
(2) https://github.com/ruby/ruby/commit/b7f7b0a2d25071dd8fe09b3f313168bf7f939cc7#diff-3b3a6ec97232deb43dc14319a73872c1

When reverting these two commits, the bash configure command proceeds without error.

ekinanp:
Created patches for the reverts, now re-running the ad-hoc builds.

ekinanp:
Looks like the patches worked for Windows on Ruby 2-3-5 and RHEL on Ruby 2-4-2; packages were built. The failure is still happening on AIX 7.1 and this is b/c the configure.in file in the TAR'd up version of ruby is different from what's on the repo, so the patches fail to apply correctly. Will look into this more later.

ekinanp:
Update: patches for AIX have been updated and tested locally, re-running the ad-hoc builds for Ruby 2.3.5 and 2.4.2. Once they pass, then this PR should be good for a merge.

ScottGarman:
Looks like there is a minor syntax tweak needed to make rubocop happy. I agree we should eventually update pl-ruby to a newer version. Assuming this all blends, it LGTM. Another (but more tedious) runtime check you can do for each package is validate that the rubygems environment is still working. I do this when developing new agent platforms to avoid getting packages that build but are still broken at runtime:

## Check the Ruby Environment

/opt/puppetlabs/puppet/bin/gem list

(should show a decent list of gems)

/opt/puppetlabs/puppet/bin/irb

You should be able to:

require 'hocon'
require 'deep_merge'
exit

and they should return true, not throw an error.

Finally, you should be able to install a gem with native extensions:

Install the sqlite3 headers using one of:

apt-get update && apt-get install gcc make libsqlite3-dev
yum install gcc sqlite-devel
zypper in gcc sqlite3-devel

find /usr -name sqlite3.h

/opt/puppetlabs/puppet/bin/gem install sqlite3

...and the sqlite3 gem should show up in:

/opt/puppetlabs/puppet/bin/gem list

configs/components/ruby-2.4.2.rb
```diff
@@ -207,6 +216,9 @@
     end
   end
 
+  # TODO: Remove this once PA-1607 is resolved.
+  pkg.configure { [ "/opt/pl-build-tools/bin/autoconf" ] } if platform.is_aix?
+
```
geoffnichols: It looks like Travis is failing in a couple of spots with a rubocop [style error](https://travis-ci.org/puppetlabs/puppet-agent/jobs/284379356):
 
`Style/SpaceInsideBrackets: Space inside square brackets detected.`

ekinanp:
@ScottGarman @geoffnichols Thanks for the PR feedbacks, Travis should pass now. The ad-hoc builds have finished for Ruby 2.4.2 (go to https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/ and search for "2-4-2" to find the pipeline) and it looks like everything except the smoke tests passed.

For Ruby 2.3.5 (same link, but search for 2-3-5), the Puppet tests failed on RHEL7-POWER (https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_puppet-agent_intn-van-sys_suite-daily-puppet-PA-1547-5.3.x-ruby-2-3-5/1/SLAVE_LABEL=beaker,TEST_TARGET=redhat7-POWERa/console is the console output), specifically the check_puppet_run_message.rb. I believe this isn't something with ruby, but rather a translation issue -- my puppet-agent PR  probably is using an older version of puppet in the components/puppet.json file. However after matrix reloading (skipping the RHEL-power), the tests passed up to the Facter tests, which failed on the AIX machines (https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_puppet-agent_intn-van-sys_suite-daily-facter-PA-1547-5.3.x-ruby-2-3-5/SLAVE_LABEL=beaker,TEST_TARGET=aix71-POWERa/) -- the reason was b/c the ruby so file could not be located.

I believe that the failure of the Facter tests for Ruby 2.3.5 is an issue with the AIX rbconfig. I compared it with the rbconfig for Ruby 2.4.2 and it looks like they match up well, so I'm going to investigate a bit further by installing a Ruby 2.3.5 package on an AIX 6.1 and a Ruby 2.4.2 package on an AIX 7.1, see if I can replicate the test failures and then seeing where the ruby so files end up. I know that a simpler solution is probably to just remove the "TEENY" version from the LIBRUBY_SO entry in the config file, but this would be inconsistent with what's in 2.4.2 (as LIBRUBY_SO there matches up), so I'd rather not do it unless I don't have a choice.

I'm going to look at this tomorrow though. Right now, I did a matrix reload on the 2.3.5 pipeline to skip the AIX stuff and see if the PXP-agent tests pass.

I'll also run the tests Scott suggested w/ the gems then as well.

ekinanp:
Update: PXP-agent tests also passed on Ruby 2.3.5.

ekinanp:
Talked to @McdonaldSeanp about it, looks like the TEENY versions for Ruby 2.3.0 should be 0 b/c Ruby 2.3 does not use the Z release, only Ruby 2.4 does.

ekinanp:
On second thought, it might be better to implement @ScottGarman 's ideas as a series of acceptance tests in the puppet-agent module. They seem to fit under that category. @McdonaldSeanp and @geoffnichols , what do you both think?

McdonaldSeanp:
@ekinanp there's an open ticket to add gem/ruby acceptance:

https://tickets.puppetlabs.com/browse/PA-1319

McdonaldSeanp:
Note that we should not add those tests as part of the puppet-agent module. They should just be included with the regular puppet-agent acceptance tests at https://github.com/puppetlabs/puppet-agent/tree/master/acceptance

ekinanp:
@McdonaldSeanp Awesome! I think I will pick that one up once this ticket has been merged in.

ekinanp:
Update: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/

Search for "ruby-2-3-5". It looks like the ad-hoc build passed, save for:

1. A failed puppet test on RHEL-POWER, as I outlined above. This is something with the translation that I don't think is an issue with ruby

2. PXP-agent test fail on AIX-71, specifically the run_echo.rb test. Looking at the output, this is a child io failure, i.e. a transient that occurred before in the past.

So I think the CI failures are not caused by any of the ruby changes here. Should be good for a merge.

ScottGarman:
I'm wary of merging this during after-hours, anyone else should feel free to do so if they're ready.

PR #1212, opened 2017-10-03
(PA-1487) Update broken ruby/curl builds

McdonaldSeanp:
Recently we updated the agent to build using only pl-gcc as we found out many
C components were building with a system GCC. During the mergeup for that work
to the 5.3.x branch, some things were missed

This commit fixes ruby 2.4.1 and curl builds.


McdonaldSeanp:
build (with failing eos-4 which is a pl-gettext issue): https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent_pkg-van-ship_daily-ad-hoc/205/

McdonaldSeanp:
@ekinanp see my comment above :)

ekinanp:
Reviewed too soon. LGTM

PR #1215, opened 2017-10-06
(maint) Remove pl-autoconf for AIX machines

ekinanp:
Work on PA-1547 requires an ad-hoc build on AIX which requires
pl-autoconf. Unfortunately, pl-autoconf has conflicting files with
pl-gettext that will cause the package installation to fail. This is
handled in the fork containing the work for PA-1547, but it is not
handled here. Thus, this PR ensures that pl-autoconf is erased whenever
a build is about to proceed on an AIX builder -- it is a temporary
fix to get CI passing until PA-1547 has been merged.


McdonaldSeanp:
👍

PR #1218, opened 2017-10-09
(PA-1580) Enable testing of system openssl linked puppet agent on rhel7 platforms

jsane:
The install script in all relevant components including puppet-agent repo
needs to be updated so it upgrades the version of openssl package on rhel7 platforms.
RHEL7 images used by CI/VMpooler uses an old openssl version, compared to that on
Centos7 which is where PA is built on,  which causes agent setup and tests to fail.
This has been preventing PA built against openssl to be tested in CI pipelines
which require one test target to be at least RHEL7.

Following components have been updated with above changes that puppet-agent now
refers to:
  - puppet-agent
  - puppet
  - hiera
  - facter
  - pxp-agent

Has been tested in PA CI - logs would be attached to PR for PA-1264 (enables linking PA against system openssl) which is required to do this testing. This does not impact regular agent build using vendor-ed openssl which disregard system openssl.

### DIFF
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -9,6 +9,21 @@ step "Install puppet-agent..." do
   }
   agents.each do |agent|
     next if agent == master
+    
+    # Move the openssl libs package to a newer version on redhat platforms
+    use_system_openssl = ENV['USE_SYSTEM_OPENSSL']
+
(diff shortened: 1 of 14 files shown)


jsane:
This PR needs to change after reaching a broader consensus on the approach to identify FIPS targets using platform names instead of using environment variables. We may also not have the need to do the following - currently being done via scripts:
- Upgrade openssl patch version on rhel7
- Move the agent hosts into FIPS mode

As the plan is to create VM images  that will have the above in place.

PR #1221, opened 2017-10-13
Merge 5.3.x -> master

ekinanp:


ScottGarman:
I'm out of the loop on why the ruby bump was reverted in the 1.10.x branch, but if that reason applies to 5.3.x as well, you probably want to revert it before the merge-up. Otherwise, I'm :+1:

PR #1222, opened 2017-10-13
(PA-1624) Update dmidecode to to 3.1 for el-7-aarch64

caseywilliams:
el-7-aarch64 uses SMBIOS 3.0, which dmidecode 2.1 can't understand; I included dmidecode 3.1 here, since its git history contains all the recommended patches for 3.0. I did a test build of this today and it ran successfully (as did e.g. `facter dmi`) on an aarch64 centos 7 host.


branan:
:+1:

On master, I'd like to see us just move to the new dmidecode globally

PR #1226, opened 2017-10-17
(PA-1503) Update to latest curl

caseywilliams:
Curl 7.56.0 is the latest version of curl, and there are no vulnerabilities listed for it (yet) on curl's site. I built this for several platforms and architectures (including AIX, which requires a special patch) and all seemed to be in order.


ScottGarman:
As a best practice, it's always nice to have an ad-hoc build url to point to in a PR to show which platforms this has been tested against - something I recommend for future PRs.

:+1:

PR #1227, opened 2017-10-18
(PA-1319) Add tests that validate the vendored ruby

ekinanp:


ekinanp:
Ran the tests locally using the following environment variables:
`export SHA=8cc295fa39ff6fa53d3045c5f1ecd8ea312e1db7
  export SUITE_VERSION=1.10.8.55.g8cc295f
  export BEAKER_HOSTS="${results_dir}/${HOSTS_NAME}"
  export TESTS=tests/validate_vendored_ruby.rb
  export OPTIONS='--test-tag-exclude=server'
  export ABS_RESOURCE_HOSTS=`get-abs-resources "${host_layout}"`
  export SERVER_VERSION=2.8.0`

where the SHA and SUITE_VERSIONS were obtained from https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/1.10.x/view/Suite/job/platform_puppet-agent_intn-van-sys-suite-puppet-agent-daily-1.10.x/SLAVE_LABEL=beaker,TEST_TARGET=sles12-64a/65/parameters/.

The test command I used was:
`bundle exec rake acceptance:development`

and the host-layout used was:
`redhat7-64m-redhat7-64a-redhat7-POWERa-aix71-POWERa-windows2012r2-64a`

and the tests all passed.

acceptance/tests/validate_vendored_ruby.rb
```diff
@@ -0,0 +1,54 @@
...
+      irb = irb_command(agent)
+      ['hocon', 'deep_merge'].each do |gem|
+        stdout = on(agent, "echo \"require '#{gem}'\" | #{irb}").stdout.chomp
+        assert_match(/true/, stdout, "'irb' failed to require the #{gem} gem")
+      end
+    end
+  end
+
+  step "'gem install' successfully installs a gem with native extensions" do
+    dependencies = {
```
ekinanp: It would be great if more providers and their relevant dependencies could be added here (e.g. dpkg). Better yet, we could have a hash of "provider => [<dependencies>, <gem>]" so we're not limited to sqlite. Also would be nice to have this test for Windows agents.

@ScottGarman any suggestions? I based this list off your comment on https://github.com/puppetlabs/puppet-agent/pull/1210.

McdonaldSeanp: native extensions are extremely tricky on windows because they require the ruby devkit for windows. I'm not sure anyone is (or should be) using native extensions with our vendored ruby for windows

acceptance/tests/validate_vendored_ruby.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  step "'gem list' displays a decent list of gems" do
+    minimum_number_of_gems = 3
+    # \d+\.\d+\.\d+ describes the gem version
+    gem_re = /[\w-]+ \(\d+\.\d+\.\d+(?:, \d+\.\d+\.\d+)*\)/
+
+    agents.each do |agent|
+      gem = gem_command(agent) 
+      listed_gems = on(agent, "#{gem} list").stdout.chomp.scan(gem_re)
+      assert(listed_gems.size >= minimum_number_of_gems, "'gem list' fails to list the available gems")
```
ScottGarman: I'd find it more helpful here if the error message was a bit more verbose about the context of this test. We're running it to make sure the gem environment is sane, so maybe update the error message to explain that the ruby gem environment doesn't appear to be working, or rename the test step name to explain that the test is intended to validate the ruby gem environment is working. When a test case blows up I find it helpful to get some context as to the purpose behind the test.

acceptance/tests/validate_vendored_ruby.rb
ScottGarman: @McdonaldSeanp does that mean users can't build nokogiri on windows? If not, do we include it in the agent? nokogiri is typically the main vendored extension gem that our users need in my experience.

ekinanp:
@ScottGarman I fixed the error message based on our discussion.

I'm going to try to see if sqlite3 can be built on other platforms (e.g. AIX, Solaris) so that we can test their rubygems environment as well.

ekinanp:
Updated PR to include AIX and OSX. Working on building a native gem on cisco windriver linux and arista

ekinanp:
Ok, this PR should be read to merge.

ScottGarman:
:+1:

PR #1228, opened 2017-10-19
Mergeup 1.10.x ->  5.3.x

caseywilliams:
* upstream/1.10.x:
  bumping puppet to 3eff4193ed5620ae88feaf5850f1d94f625225c4
  (PA-1503) Update to latest curl
  (PA-1624) Update dmidecode to to 3.1 for el-7-aarch64

### DIFF
--- a/configs/components/dmidecode.rb
+++ b/configs/components/dmidecode.rb
@@ -1,17 +1,22 @@
 component 'dmidecode' do |pkg, settings, platform|
-  pkg.version '2.12'
-  pkg.md5sum '02ee243e1ecac7fe0d04428aec85f63a'
-  pkg.url "http://download.savannah.gnu.org/releases/dmidecode/dmidecode-#{pkg.get_version}.tar.gz"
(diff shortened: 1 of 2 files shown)


configs/components/dmidecode.rb
```diff
@@ -1,17 +1,22 @@
 component 'dmidecode' do |pkg, settings, platform|
-  pkg.version '2.12'
-  pkg.md5sum '02ee243e1ecac7fe0d04428aec85f63a'
-  pkg.url "http://download.savannah.gnu.org/releases/dmidecode/dmidecode-#{pkg.get_version}.tar.gz"
-  pkg.mirror "http://buildsources.delivery.puppetlabs.net/dmidecode-#{pkg.get_version}.tar.gz"
```
ScottGarman: Seems like we're losing the pkg.mirror setting in this merge-up. Was there a merge conflict that you had to manually resolve? If so, that's a great opportunity to add the pkg.mirror line back in.

PR #1229, opened 2017-10-19
(maint) Add package mirror for dmidecode

caseywilliams:
Vanagon has been updated to the same version in 1.10.x as in 5.3.x, so this is possible now.


branan:
:+1:

PR #1235, opened 2017-10-23
(PA-1503) Update curl to v7.56.1

caseywilliams:
7.56.1 came out just after we merged the bump to 7.56.0

### DIFF
--- a/configs/components/curl.rb
+++ b/configs/components/curl.rb
@@ -1,6 +1,6 @@
 component 'curl' do |pkg, settings, platform|
-  pkg.version '7.56.0'
-  pkg.md5sum '65351b9df687ed539852ae7b9464006c'
+  pkg.version '7.56.1'
+  pkg.md5sum '48ba7bd7b363b40cd446d1e7b4be9920'
   pkg.url "https://curl.haxx.se/download/curl-#{pkg.get_version}.tar.gz"
 
   pkg.build_requires "openssl"


configs/components/curl.rb
```diff
@@ -1,5 +1,5 @@
 component 'curl' do |pkg, settings, platform|
-  pkg.version '7.56.0'
+  pkg.version '7.56.1'
   pkg.md5sum '65351b9df687ed539852ae7b9464006c'
```
MikaelSmith: Need to update the checksum. I suggest testing builds on a few platforms: a linux, windows, maybe solaris/aix.

ScottGarman:
Yep, this will need a link to a successful ad-hoc build on pretty much all of our platforms.

configs/components/curl.rb
caseywilliams: Gah, thanks @MikaelSmith, I'll update and push again once I've successfully built this on a few platforms.

caseywilliams:
Okay, PR updated with the md5sum, and here's an ad-hoc build where curl succeeded for a selection of Linux, Solaris, AIX, Mac, and Windows: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent_puppet-agent-packaging_daily-ad_hoc/2/

PR #1239, opened 2017-10-23
(maint) Update semantic_puppet gem to latest version

haus:
The vendored version in Puppet is already at 1.0.1, so this commit brings the version in puppet-agent in line with that.

### DIFF
--- a/configs/components/rubygem-semantic_puppet.rb
+++ b/configs/components/rubygem-semantic_puppet.rb
@@ -1,6 +1,6 @@
 component "rubygem-semantic_puppet" do |pkg, settings, platform|
-  pkg.version "0.1.2"
-  pkg.md5sum "192ae7729997cb5d5364f64b99b13121"
+  pkg.version "1.0.1"
+  pkg.md5sum "8ce85a0206a640654e0019eadf615672"
   pkg.url "https://rubygems.org/downloads/semantic_puppet-#{pkg.get_version}.gem"
   pkg.mirror "http://buildsources.delivery.puppetlabs.net/semantic_puppet-#{pkg.get_version}.gem"
 


joshcooper:
I thought we tried to do this earlier but had to revert due to compatibility with older versions of puppetserver that expect the pre-1.0 behavior. I could be misremembering... /cc @thallgren

haus:
(not sure why travis is failing. the PR does start with `(maint)`)

@joshcooper that's surprising as I would expect puppetserver to be using the vendored version?

haus:
@joshcooper I think last time the issue was it caused r10k headaches when it was out of sync. I believe that has been addressed by hard coding versions into the gemspecs that puppet/hiera ship. I may not have all of the past context though.

branan:
Didn't we run into issues from the community when we tried to unvendor from Puppet? And if so, how is that related to the version we have shipping here compared to the newer one that's vendored in Puppet?

Additionally, what are the semver implications of revving this from 0.1 to 1.0?

haus:
I'm going down the route of putting this in pe-r10k for hoyt for now, but I would like to know how to move this forward. Will semantic_puppet be unvendored from puppet itself at some point? At that point I presume it will be part of puppet-agent and be usable by pe-r10k and we can remove semantic_puppet from pe-r10k again.

justinstoller:
This went into pe-r10k for Hoyt. Don't know if we care that r10k uses different versions of semantic puppet though (FOSS r10k gets semantic_puppet ~> 1.0 transitively).

If this goes in we should probably update server as well.

geoffnichols:
@justinstoller and @joshcooper is this still something we want to land in 5.3.x?

caseywilliams:
If so, we should update this in puppet-runtime (I can make a quick PR to do that) - Now that #1378 is merged, semantic_puppet will be extracted from a runtime tarball instead of built with the agent each time.

justinstoller:
:noidea:

But Haus' removal from the Puppetlabs Github org (and hence part of his offboarding) is pending resolution of his open PRs. I have a feeling that if this PR isn't good to go as-is we can close it and add on whatever is required to https://tickets.puppetlabs.com/browse/PA-1880

joshcooper:
After this PR was filed we had to revert puppet back to pre-1.0. I'm going to close this PR. We have a ticket filed https://tickets.puppetlabs.com/browse/PA-1881 to update the puppet-agent version as part of the unvendoring semantic puppet epic. I'm hesitant to make any changes in 5.x due to the fragile semantic_puppet dependencies among puppet, puppet-agent and puppetserver.

PR #1241, opened 2017-10-24
(maint) stop hardcoding :libdir in rpath settings

DavidS:


caseywilliams:
With the merging of #1378, these ruby configs have moved into the puppet-runtime project - we did have to apply the changes in this PR there to allow reuse of ruby in multiple projects that consume the runtime - [here's the equivalent section of the shared ruby configs](https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/base-ruby.rb#L37).

PR #1243, opened 2017-10-26
(maint) Facilitate EC2 testing using always-be-scheduling

mchllweeks:
(maint) Always-be-scheduling needs to use a different keyfile for beaker specified by an environment variable in order to facilitate ec2 testing.


mchllweeks:
retest this please

PR #1245, opened 2017-10-27
(PA-1522) Set REINSTALLMODE=amus

jcoconnor:
This defaults to always replacing files regardless of their version or
checksum. This really only works for future upgrades, so the
REINSTALLMODE needs to be specified manually when downgrading to a
version of Puppet Agent without REINSTALLMODE set.

### DIFF
--- a/resources/windows/wix/properties.wxs.erb
+++ b/resources/windows/wix/properties.wxs.erb
@@ -12,6 +12,12 @@
       Id="ARPHELPLINK"
       Value="<%= settings[:links][:HelpLink] %>" />
 
+    <!-- Reinstallmode default to always replace files -->
+    <!-- https://msdn.microsoft.com/en-us/library/aa371182%28v=vs.85%29.aspx -->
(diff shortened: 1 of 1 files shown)


jcoconnor:
Assuming this needs placed against ```stable``` in the first instance.

jcoconnor:
Have done a quick smoke test by building and installing puppet-agent and can confirm that the REINSTALLMODE is showing up in the log as:
```
Property(C): ARPHELPLINK = http://puppet.com/services/customer-support
Property(C): REINSTALLMODE = amus
Property(C): VersionUIString = 5.3.2.234
```

Iristyle:
@jcoconnor / @glennsarti I think we wanted to target the 1.10 LTS series of releases, but this was merged into master which didn't target either 1.10.x or 5.3.x releases.

Iristyle:
Follow-on PR for correct 1.10.x branch is #1253

PR #1255, opened 2017-11-06
(PA-1699) Use gcc 4.x to compile sqlite3 gem on Solaris

ekinanp:
Previously, the vendored ruby tests failed on i386 and SPARC for Solaris 10,
and only i386 for Solaris 11 on the puppet-agent master pipelines. This
was due to the fact that the 5.3.x and master pipelines ship Ruby 2.4.2
with the agent. Unfortunately gcc 3.4.x, used on the failing Solaris
machines, does not work with Ruby 2.4.2. However gcc 4.x, used on
Solaris 11 SPARC (the passing case), does. This commit uses gcc 4.x for
all the Solaris platforms.

### DIFF
--- a/acceptance/tests/validate_vendored_ruby.rb
+++ b/acceptance/tests/validate_vendored_ruby.rb
@@ -27,7 +27,12 @@ def setup_build_environment(agent)
     # PA-1638
     gem_install_sqlite3 += " -v 1.3.11" 
   when /solaris-11-i386/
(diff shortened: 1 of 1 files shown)


ekinanp:
Cherry-pick commit to get this from master into 1.10.x

McdonaldSeanp:
👍

ScottGarman:
:+1:

PR #1259, opened 2017-11-09
(packaging) Update shipping defaults to work with new puppet5 repos

mwaggett:


ext/build_defaults.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 project: 'puppet-agent'
-packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
+packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=1.0.x'
```
smcclellan: Why does this branch name change?

mwaggett:
This is a cherry-pick of a change we already merged into master.
We've added ~*New and Improved*~ automation related to the puppet5 work, which was done on a separate branch as we worked out the kinks in our automation. But we've been shipping with this branch since the release of puppet5. And now that puppet-agent#master has gone green, we feel confident enough to cut over other branches/projects.

Does that help, @smcclellan ?

PR #1265, opened 2017-11-21
(maint) merge up 5.3.x to master

Iristyle:


configs/components/facter.rb
```diff
@@ -19,7 +19,11 @@
   pkg.build_requires 'leatherman'
   pkg.build_requires 'runtime'
   pkg.build_requires 'cpp-hocon'
-  pkg.build_requires 'libwhereami'
+
+  unless platform.is_aix? || platform.is_solaris?
+    # libwhereami doesn't support AIX or Solaris yet
+    pkg.build_requires "libwhereami"
+  end
```
caseywilliams: Master should actually have libwhereami - its master includes support for solaris/aix

McdonaldSeanp:
👍

configs/components/facter.rb
caseywilliams: (Other than that these changes look fine to me)

Iristyle: So master should be completely unguarded then, like?

```ruby
pkg.build_requires 'libwhereami'
```

caseywilliams: Right, just include for all platforms

ScottGarman:
:+1:

PR #1266, opened 2017-11-21
(PA-1623) Update AIX builds to use pooler engine

McdonaldSeanp:
Update builds of AIX to not use static builders anymore


configs/platforms/aix-6.1-ppc.rb
```diff
@@ -6,12 +6,9 @@
...
-  plat.provision_with "rpm --quiet -e autoconf binutils coreutils diffutils expect gcc gcc-c++ libgcc libstdc++ libstdc++-devel libxml2 libxml2-devel m4 make mktemp pkg-config pl-boost pl-cmake pl-gcc pl-yaml-cpp readline readline-devel screen sed tcl tk which zlib-devel cdrecord mkisofs; rm -rf /opt/pl-build-tools /opt/puppetlabs /etc/puppetlabs /var/log/puppetlabs /var/tmp/tmp.* /var/tmp/*root-root /var/tmp/rpm* /root/*.jam; test -f /opt/freeware/bin/granlib  && chmod 755 /opt/freeware/bin/granlib|| true"
-
   # Basic vanagon operations require mktemp, rsync, coreutils, make, tar and sed so leave this in there
   plat.provision_with "rpm -Uvh --replacepkgs http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/mktemp-1.7-1.aix5.1.ppc.rpm http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/rsync-3.0.6-1.aix5.3.ppc.rpm http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/coreutils-5.2.1-2.aix5.1.ppc.rpm http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/sed-4.1.1-1.aix5.1.ppc.rpm http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/make-3.80-1.aix5.1.ppc.rpm http://osmirror.delivery.puppetlabs.net/AIX_MIRROR/tar-1.22-1.aix6.1.ppc.rpm"
```
ekinanp: I wonder if it would be worthwhile to add pl-gcc-5.2.0-11.aix#{platform.os_version}.ppc.rpm here, so it can be removed from all those components.

McdonaldSeanp: That's a larger PR than this.

We have talked before about figuring out the best way to list deps, and still have not landed on a good solution. When we do we will refactor then.

ScottGarman:
:+1:

PR #1268, opened 2017-11-22
(maint) Merge 5.3.x -> master

mwaggett:


mwaggett:
#yolo I'm self-merging because this has two official thumbs, another set of unofficial thumbs, and it's just a merge-up.

PR #1272, opened 2017-11-30
(PA-1619) Update ruby 2.4.2 to include zlib for AIX

McdonaldSeanp:
The AIX builds need to include the zlib package to correctly install
zlib-devel.


McdonaldSeanp:
This is incomplete, there are still issues configuring the deps

PR #1280, opened 2017-12-04
(CPR-391) Fedora packages have odd distro tags

mwaggett:
I have been "working" on this ticket for like a year..


configs/components/facter.rb
```diff
@@ -61,7 +61,7 @@
   java_home = ''
   java_includedir = ''
   case platform.name
-  when /(el-(6|7)|fedora-f24)/
```
underscorgan: Did we just forget to add fedora 25 for this case? Should we have a java build requires for fc25?

configs/components/puppet-ca-bundle.rb
```diff
@@ -5,7 +5,7 @@
 
   java_available = true
   case platform.name
-  when /(el-(6|7)|fedora-f24)/
```
underscorgan: same question for fc25/java

configs/components/facter.rb
underscorgan: or should we have a generic fedora entry until there's a version it breaks for?

mwaggett:
@underscorgan I'm in the process of opening PRs to update tests to handle this. Right now, we have NOT been manually renaming the packages, but we HAVE been manually moving them to directories without the f (because apparently we're terrible and inconsistent). In theory, beaker should know where to get the packages because of the stuff in the <ref>.yaml.

mwaggett:
@ekinanp oops, yep, will fix!

configs/components/facter.rb
mwaggett: @McdonaldSeanp @branan ^^

branan: We need the java build-requires for Master platforms so that libfacter.so has JNI support.

configs/components/puppet-ca-bundle.rb
branan: I have no idea what we need the JDK for in the ca-bundle :man_shrugging:

configs/components/facter.rb
branan: It looks like we build with JNI support on a ton of platforms below - I'm not certain what we're shipping puppetserver on these days, but we should probably clean this up so we're not building against Java when we don't need to.

mwaggett: That all seems like work separate from this PR though.

branan: For sure. I don't think there's anything we need to do as part of this PR

branan:
Is there any way of validating that this will roll through cleanly without merging it? We can (and should) run the puppet-agent ad-hoc pipeline, but that won't test any of the PE-related bits - promotion, pe-repo, and the puppet-agent module.

configs/components/puppet-ca-bundle.rb
branan: Oh, looks like it's just to make the CA bundle available to puppetserver... because of course we use the bundle in puppet-agent for that :weary:

mwaggett:
We've decided to target this change at Platform 6 - closing this.

PR #1283, opened 2017-12-06
(PA-1402) Avoid loading custom facts during upgrade

npwalker:
Prior to this commit, we saved the output of
`puppet resource service puppet` and then applied it again after
upgrade with `puppet apply`.  This causes custom facts to be
loaded by `puppet apply` which is undesirable when your custom
fact wants to query the package manager.

After this commit, we blank out some puppet settings so that the
agent won't be able to find any custom facts that have been
pluginsynced or if its the agent on a master then custom facts
in the modulepath.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -58,7 +58,7 @@ component "puppet" do |pkg, settings, platform|
       [<<-HERE.undent
         mkdir -p  #{rpm_statedir} && chown root #{rpm_statedir} && chmod 0700 #{rpm_statedir} || :
         if [ -x #{puppet_bin} ] ; then
-          #{puppet_bin} resource service puppet > #{service_statefile} || :
+          #{puppet_bin} resource service puppet | awk -F "'" '/ensure =>/ { print $2 }' > #{service_statefile} || :
         fi
         HERE
       ]
@@ -66,7 +66,7 @@ component "puppet" do |pkg, settings, platform|
     pkg.add_postinstall_action ["upgrade"],
       [<<-HERE.undent
         if [ -f #{service_statefile} ] ; then
-          #{puppet_bin} apply #{service_statefile} > /dev/null 2>&1 || :
+          #{puppet_bin} resource service puppet ensure=$(cat #{service_statefile}) > /dev/null 2>&1 || :
           rm -rf #{rpm_statedir} || :
         fi
         HERE


geoffnichols:
ping @shrug for review of this one

McdonaldSeanp:
jenkins please test this,puppet

geoffnichols:
jenkins please test this,puppet

npwalker:
jenkins please test this,puppet

branan:
I have no idea what that test failure is. Since this doesn't affect puppet itself, just the P-A install/upgrade process, I think it's safe to ignore

branan:
I don't think this a correct solution, since it's not possible to completely turn off Facter's external and Ruby fact loading. Even with Puppet's `factpath` and `pluginfactdest` set to empty strings, we would still try to load facts from:

* Modules
* Ruby's load path
* Facter's default search paths
* Any custom paths defined in Facter's configuration file

Essentially, this PR in its current state only protects against loading pluginsunk facts.

The original idea of parsing out the state of the service and using `puppet resource` is the best I can come up with. It would be nice to have that extraction be cleaner than a series of `sed` invocations, tough.

npwalker:
@branan okay I updated the PR back to the original implementation of grabbing the service state and then using puppet resource service to apply it again.

configs/components/puppet.rb
```diff
@@ -58,15 +58,15 @@
       [<<-HERE.undent
         mkdir -p  #{rpm_statedir} && chown root #{rpm_statedir} && chmod 0700 #{rpm_statedir} || :
         if [ -x #{puppet_bin} ] ; then
-          #{puppet_bin} resource service puppet > #{service_statefile} || :
+          #{puppet_bin} resource service puppet | grep ensure | sed 's/ensure => //' | sed s'/.$//' | sed 's/^..//' | sed "s/'//g" > #{service_statefile} || :
```
abottchen: This can be tightened up to just:

`#{puppet_bin} resource service puppet | grep -Pzo "(?<=ensure => ').*(?=')"`

branan: We use this same pre-inst step for AIX. Do all those grep flags work correctly there?

My inclination would be a `grep ensure | awk <some expr>`, to ensure compatibility with weirder platforms that use RPM

abottchen:
Yea, AIX borks up grep...  NVM

npwalker:
okay @branan we have a single `awk` now.  Which I guess is better than 1 `grep` and 4 `sed`s

npwalker:
jenkins please test this,puppet

branan:
Jenkins please test this

branan:
Trying to figure out what's up with those puppet failures. I otherwise believe this is good to go

branan:
jenkins please test this,puppet

branan:
I still don't know what would have triggered those failures. I'm gonna YOLO this and hope it doesn't hit us in Jenkins.

PR #1284, opened 2017-12-15
(PA-673) PA-ENVIRONMENT should default to blank

jcoconnor:
This was addressed originallin in PA-286 (in Puppet for the Win), but
this particular change didn't make it into the reworked Vanagon code.
So - ensure that if no ENVIRONMENT is set for the install then it should
not be set in puppet.conf, i.e. desired behaviour is:
- On a fresh install with PUPPET_AGENT_ENVIRONMENT not specified,
  puppet.conf does not have an environment key.
- On a fresh install with PUPPET_AGENT_ENVIRONMENT=foo specified,
  puppet.conf does have a key of environment=foo.
- On an upgrade where puppet.conf has an environment key, the value
  stays the same.
- On an upgrade where puppet.conf does not have an environment key,
  the key remains out of puppet.conf.

### DIFF
--- a/resources/windows/wix/customactions.wxs.erb
+++ b/resources/windows/wix/customactions.wxs.erb
@@ -125,11 +125,6 @@ End If
       Property="PUPPET_AGENT_ENVIRONMENT"
       Value="[CMDLINE_PUPPET_AGENT_ENVIRONMENT]"
       Execute="firstSequence" />
-    <CustomAction
-      Id="SetDefaultPuppetAgentEnvironment"
-      Property="PUPPET_AGENT_ENVIRONMENT"
-      Value="production"
-      Execute="firstSequence" />
 
     <!-- PUPPET_AGENT_CERTNAME -->
     <CustomAction
--- a/resources/windows/wix/sequences.wxs.erb
+++ b/resources/windows/wix/sequences.wxs.erb
@@ -33,9 +33,6 @@
       <Custom Action='SetFromCmdLinePuppetAgentEnvironment' After='FileCost'>
         CMDLINE_PUPPET_AGENT_ENVIRONMENT
       </Custom>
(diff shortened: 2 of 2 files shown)


jcoconnor:
@natemccurdy /cc - you may want to check this too

McdonaldSeanp:
👍 to the code itself.

McdonaldSeanp:
@jcoconnor should this fix go in all the branches? (i.e. 1.10.x and 5.3.x too?)

If so we need to retarget this at 1.10.x

McdonaldSeanp:
I've added question and don't merge until we decide where this needs to land.

natemccurdy:
@jcoconnor Seems like it'll do the right thing, but I'm not in any position to approve this as I can't test it. Thanks for putting the PR up though 😄

resources/windows/wix/customactions.wxs.erb
```diff
@@ -125,11 +125,6 @@ End If
       Property="PUPPET_AGENT_ENVIRONMENT"
       Value="[CMDLINE_PUPPET_AGENT_ENVIRONMENT]"
       Execute="firstSequence" />
-    <CustomAction
-      Id="SetDefaultPuppetAgentEnvironment"
```
Iristyle: Shouldn't we also delete the action from the sequence in https://github.com/puppetlabs/puppet-agent/blob/954f06b24f68ffdfa0bdb2d2b9bc73970ecb8a37/resources/windows/wix/sequences.wxs.erb#L36?

jcoconnor: Yip - doing that now - sorry - had submitted PR before checking build had completed.

jcoconnor:
Closing this PR - Tried to rebase it against ```1.10.x``` but hit too many conflicts, so have cherry-picked changes to new branch and created #1286

PR #1286, opened 2017-12-18
(PA-673) PA ENVIRONMENT should default to blank

jcoconnor:
This was addressed originallin in PA-286 (in Puppet for the Win), but
this particular change didn't make it into the reworked Vanagon code.
So - ensure that if no ENVIRONMENT is set for the install then it should
not be set in puppet.conf, i.e. desired behaviour is:
- On a fresh install with PUPPET_AGENT_ENVIRONMENT not specified,
  puppet.conf does not have an environment key.
- On a fresh install with PUPPET_AGENT_ENVIRONMENT=foo specified,
  puppet.conf does have a key of environment=foo.
- On an upgrade where puppet.conf has an environment key, the value
  stays the same.
- On an upgrade where puppet.conf does not have an environment key,
  the key remains out of puppet.conf.

### DIFF
--- a/resources/windows/wix/customactions.wxs.erb
+++ b/resources/windows/wix/customactions.wxs.erb
@@ -102,11 +102,6 @@ End If
       Property="PUPPET_AGENT_ENVIRONMENT"
       Value="[CMDLINE_PUPPET_AGENT_ENVIRONMENT]"
       Execute="firstSequence" />
-    <CustomAction
-      Id="SetDefaultPuppetAgentEnvironment"
-      Property="PUPPET_AGENT_ENVIRONMENT"
-      Value="production"
-      Execute="firstSequence" />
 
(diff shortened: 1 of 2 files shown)


jcoconnor:
This replaces #1284

jcoconnor:
@Iristyle & @McdonaldSeanp I have built and tested the clean installation and this appears to install correctly.

Artefacts are available at: ```http://builds.delivery.puppetlabs.net/puppet-agent/d34bc88548b0bf375b3eb8a88af76016733661a6/artifacts/windows/```

McdonaldSeanp:
👍

Iristyle:
Looks like this matches the original PR merged in https://github.com/puppetlabs/puppet_for_the_win/pull/142/files

PR #1288, opened 2017-12-20
(packaging) Update Gemfile to allow shipping to Artifactory

shrug:


McdonaldSeanp:
jenkins please test all

McdonaldSeanp:
jenkins please test all

Gemfile
```diff
@@ -11,7 +11,8 @@ def vanagon_location_for(place)
 end
 
 gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '~> 0.14.1')
-gem 'packaging', '~> 0.4', :github => 'puppetlabs/packaging'
+gem 'packaging', :github => 'puppetlabs/packaging', :branch => '1.0.x'
+gem 'artifactory'
```
joshcooper: Would it make more sense to add `artifactory` as a dependency of `packaging` (in packaging.gemspec) so that consumers of packaging don't need to know about packaging's dependencies, e.g. https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/artifactory.rb#L23?

shrug: Yes, it would. And it is already in the packaging.gemspec, but we're not installing Packaging as a gem, so the deps don't get resolved. That's coming soon though, hopefully.

PR #1298, opened 2018-01-16
(PA-1731) Update curl to v7.57.0

caseywilliams:
Updates curl from 7.56.1 to 7.57.0.

+ The additional `environment` else case here is backported from 5.3.x and master, which had this path setting in their curl components.


caseywilliams:
jenkins please test this

McdonaldSeanp:
jenkins please test this on aix61-POWER,windows2012r2-64,osx1012-64

PR #1312, opened 2018-01-29
(PA-349) Remove unused windows build files

McdonaldSeanp:
Long ago we build windows differently (and did not use vanagon to build things
for that OS). Now that windows has been building with vanagon this PR removes
  stale windows configs


geoffnichols:
👍

PR #1314, opened 2018-01-31
(maint) Bump to packaging 1.0.x

mwaggett:
This commit changes the branch of packaging to clone from. As RE is working towards fully automating the shipping process, there is certain functionality in packaging#1.0.x that is necessary, but not present in packaging#master.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -1,6 +1,6 @@
 ---
 project: 'puppet-agent'
-packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
+packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=1.0.x'
 packaging_repo: 'packaging'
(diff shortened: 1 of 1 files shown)


branan:
jenkins please test this

branan:
I primarily wanted to see this go through PEZ, which apparently doesn't happen in PR testing, so... ¯\\\_(ツ)_/¯

mwaggett:
@branan since your other branches are on packaging#1.0.x, I can't think of anything that would break.. but I admit I don't know a ton of PEZ stuff.

branan:
@mwaggett I think the only thing I was concerned about were any puppet-5-related path changes?

mwaggett:
@branan We put backwards compatibility in. https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/paths.rb#L81-L88

branan:
Closing in favor of #1315

PR #1315, opened 2018-01-31
(maint) Update to use packaging as a gem

underscorgan:
Rather than cloning packaging to ext/packaging, we should be using
packaging as a gem. This allows us to not directly depend on
artifactory, as that is only a dependency of packaging and also vastly
simplifies logic in the Rakefile. Support for the PACKAGING_VERSION
environment variable has also been added to support runtime switching of
which packaging we're building/shipping/etc with.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-def vanagon_location_for(place)
+def location_for(place)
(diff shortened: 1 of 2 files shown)


branan:
Jenkins please test this

branan:
Jenkins please test this

underscorgan:
pipelines don't use `bundle exec rake whatever` so it's failing to find the packaging gem. :sigh:

underscorgan:
Jenkins please test this

PR #1313, opened 2018-01-31
(PA-1765) Use puppet-runtime to provide several dependencies

caseywilliams:
Uses puppet-runtime builds to provide the following dependencies:
- curl
- libedit
- libxml2
- libxslt
- openssl
- puppet-ca-bundle
- ruby
- all rubygems

This depends on https://github.com/puppetlabs/puppet-runtime/pull/5, which generates runtime tarballs for all branches of puppet-agent. 

This PR includes a new `puppet-runtime.rb` component which pulls in the runtime build and untars it in the root directory; This replaces all the removed components. I believe I've updated the [runtime PR](https://github.com/puppetlabs/puppet-runtime/pull/5) so that it contains all recent additions to the 1.10.x branch for these components, but a second set of eyes would be welcome.

This relies on the puppet-runtime repo having a tag with the version specified in the component (in this case, it's a tag in my puppet-runtime fork - we'll need to wait on the merging and tagging of https://github.com/puppetlabs/puppet-runtime/pull/5 before this PR works for the agent - I've been testing my builds using the [generic vanagon Jenkins job](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/1371/), though).

PR #1316, opened 2018-02-05
(PA-1245) Add defs for cfacter-source-gem project

McdonaldSeanp:
This will be a new vanagon project included with the agent. It makes use of 5
components:

* cpp-hocon-source < just the entire cpp-hocon source copied in to place
* leatherman-source < just the entire leatherman project copied in to place
* facter-source < just the entire facter project copied in to place
* cfacter-source-gem < a component with build rules to construct the source gem
from sources and resource files.
* cfacter-precompiled-gem < a component with build rules to actually precompile a gem and package it up

This commit is entirely additive, and should not effect the agent in any way.

The cfacter source gem is a gem that can be installed on a system with new
enough gcc/boost/cmake to give the user access to facter.

the cfacter precompiled gem should be installable right away on a system matching the arch and OS type

This commit makes use of changes from
https://github.com/puppetlabs/vanagon/pull/543

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -146,3 +146,4 @@ vanagon_project: TRUE
 apt_repo_name: 'PC1'
 yum_repo_name: 'PC1'
 build_tar: FALSE
+build_gem: TRUE
--- a/Gemfile
+++ b/Gemfile
@@ -10,9 +10,10 @@ def vanagon_location_for(place)
   end
 end
 
-gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '~> 0.14.1')
+gem 'vanagon', *vanagon_location_for(ENV['VANAGON_LOCATION'] || '~> 0.15.4')
 gem 'packaging', :github => 'puppetlabs/packaging', :branch => '1.0.x'
 gem 'artifactory'
 gem 'rake'
 gem 'json'
+gem 'octokit'
 gem 'rubocop', "~> 0.34.2"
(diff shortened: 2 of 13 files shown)


McdonaldSeanp:
This PR is actually blocked on https://github.com/puppetlabs/puppet-agent/pull/1313

Once that PR lands the el-7-x86_64_gem platform target can be removed, so don't review that.

McdonaldSeanp:
Everything was merged in vanagon, this should be ready for merge.

McdonaldSeanp:
@ekinanp the source gem is available for folks who might want to attempt to install the gem somewhere our precompiled one's aren't working. It also gives folks a place to edit and rebuild the gem (because the source gem literally comes with the facter/leatherman/cpp-hocon source)

and the |component|-source components don't build anything, they literally just copy the entire source project in to the staging dir where the gem will be built. The actual build of those components happens in cfacter-precompiled-gem.rb

ext/build_defaults.yaml
```diff
@@ -146,3 +146,4 @@ vanagon_project: TRUE
 apt_repo_name: 'PC1'
 yum_repo_name: 'PC1'
 build_tar: FALSE
+build_gem: TRUE
```
McdonaldSeanp: @underscorgan I looked at puppet's build_defaults (and a little in packaging) and this looked like the only thing I needed to enable to get gems to ship. Does that sound right to you?

underscorgan: @McdonaldSeanp that's the only thing I know of!

branan: Missing newline here. I dunno if it'll confuse any real-world YAML parsers, but I think it's technically out of spec :)

McdonaldSeanp: @branan updated

PR #1329, opened 2018-02-16
(PA-1846) Bump gettext-setup gem to 0.30

smcclellan:


McdonaldSeanp:
The failure was from something strange in jenkins, I consider the passing EL OS to be acceptable

PR #1330, opened 2018-02-16
(PA-1130) Add Debian producer trigger puppetserver-restart

ScottGarman:
This Debian package trigger will be produced during package upgrades and
in conjunction with the puppetserver package, can be used to  trigger a
restart of the puppetserver service.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -73,6 +73,10 @@ component "puppet" do |pkg, settings, platform|
       ]
   end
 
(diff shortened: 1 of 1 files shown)


ScottGarman:
I've tested this now with a trivial consumer Debian package that declares "interest" in the puppetserver-restart trigger, and then acts on that trigger in its postinst script. The trigger is successfully activated after I install puppet-agent after I install this "triggertest" pacakge, as well as during agent package upgrade.

PR #1334, opened 2018-02-23
(PA-1850) Copy dependent .dll files to bin directory for pxp-agent

ekinanp:
In PA-1850, it was found that when there is a system version of openssl
installed on Windows, then the pxp-agent will fail to start properly if
the user tries to restart it. Specifically, Puppet shows the service as
being in the "paused" state. The reason this happened is because when Windows
executes the pxp-agent.exe file, it searches for its required .dll files first
in the .exe file's directory, then in the system library directory
(e.g. C:\Windows:\System32), and then the custom PATH last (as a rough approximation,
it actually searches some other locations, but the relative ordering of ".exe directory"
=> "system library directory" => PATH is still there). See https://msdn.microsoft.com/en-us/library/7d83bc18.aspx
for more details. Since we do not have any .dll files in pxp-agent/bin, the system
openssl is detected as pxp-agent's openssl library instead of Puppet's openssl, which
is installed in [INSTALLDIR]\Puppet\puppet\bin. Of course, pxp-agent is not built to work
with the system openssl, so it fails to start properly.

By copying the dependent .dll files in the puppet/bin directory to pxp-agent/bin,
we ensure that Windows uses our in-house .dll files to start up pxp-agent.

See https://tickets.puppetlabs.com/browse/PA-1850 for all the details.


ekinanp:
I tested things out by doing the following:

(1) Installing an old version of the puppet-agent -- I used https://downloads.puppetlabs.com/windows/puppet-agent-1.10.10-x64.msi

(2) "Installing" openssl -- did this by copying a .dll file and naming it as libeay32.dll

(3) Using the puppetlabs-puppet_agent module to upgrade the agent to my dev package (here: http://builds.puppetlabs.lan/puppet-agent/c90ac595a23ce630f90a717cc3c58128598c577d/artifacts/windows/puppet-agent-x64.msi). I declared the class as:
    class {'::puppet_agent':
      package_version => "5.3.3.439.gc90ac59",
      source => 'C:\Users\Administrator\puppet_agent_PA_1850.msi',
    }
on my master VM.

(4) After the upgrade, I did an agent run. The version was OK -- 5.3.6 (b/c in the agent repo, we list the NEXT version to ship now). Note I cherry-picked this commit to the 5.3.x branch of the agent. The agent run went through without any issues. I also looked at services.msc and it was showing that the pxp-agent service was running.

NOTE just to double check, I repeated these steps using a released agent package for (3), and was able to repo. the paused pxp-agent service just like the customer -- the agent run in (4) also failed.

configs/components/pxp-agent.rb
```diff
@@ -75,6 +75,49 @@
...
+    # openssl is detected as pxp-agent's openssl library instead of Puppet's openssl, which
+    # is installed in [INSTALLDIR]\Puppet\puppet\bin. Of course, pxp-agent is not built to work
+    # with the system openssl, so it fails to start properly.
+    #
+    # By copying the dependent .dll files in the puppet/bin directory to pxp-agent/bin,
+    # we ensure that Windows uses our in-house .dll files to start up pxp-agent.
+    #
+    # See https://tickets.puppetlabs.com/browse/PA-1850 for all the details.
+    pkg.install do
+      dependent_dlls = [
```
ekinanp: These dependencies were obtained by using http://www.dependencywalker.com/ on the pxp-agent.exe file per @Iristyle's suggestion.

ekinanp:
Wrong branch, oops.

PR #1335, opened 2018-02-23
(PA-1850) Copy dependent .dll files to bin directory for pxp-agent

ekinanp:
In PA-1850, it was found that when there is a system version of openssl
installed on Windows, then the pxp-agent will fail to start properly if
the user tries to restart it. Specifically, Puppet shows the service as
being in the "paused" state. The reason this happened is because when Windows
executes the pxp-agent.exe file, it searches for its required .dll files first
in the .exe file's directory, then in the system library directory
(e.g. C:\Windows:\System32), and then the custom PATH last (as a rough approximation,
it actually searches some other locations, but the relative ordering of ".exe directory"
=> "system library directory" => PATH is still there). See https://msdn.microsoft.com/en-us/library/7d83bc18.aspx
for more details. Since we do not have any .dll files in pxp-agent/bin, the system
openssl is detected as pxp-agent's openssl library instead of Puppet's openssl, which
is installed in [INSTALLDIR]\Puppet\puppet\bin. Of course, pxp-agent is not built to work
with the system openssl, so it fails to start properly.

By copying the dependent .dll files in the puppet/bin directory to pxp-agent/bin,
we ensure that Windows uses our in-house .dll files to start up pxp-agent.

See https://tickets.puppetlabs.com/browse/PA-1850 for all the details.

### DIFF
--- a/configs/components/pxp-agent.rb
+++ b/configs/components/pxp-agent.rb
@@ -82,6 +82,49 @@ component "pxp-agent" do |pkg, settings, platform|
     ["#{make} -j$(shell expr $(shell #{platform[:num_cores]}) + 1) install"]
   end
 
+  if platform.is_windows?
+    # In PA-1850, it was found that when there is a system version of openssl
+    # installed on Windows, then the pxp-agent will fail to start properly if
+    # the user tries to restart it. Specifically, Puppet shows the service as
+    # being in the "paused" state. The reason this happened is because when Windows
+    # executes the pxp-agent.exe file, it searches for its required .dll files first
+    # in the .exe file's directory, then in the system library directory
+    # (e.g. C:\Windows:\System32), and then the custom PATH last (as a rough approximation,
+    # it actually searches some other locations, but the relative ordering of ".exe directory"
+    # => "system library directory" => PATH is still there). See https://msdn.microsoft.com/en-us/library/7d83bc18.aspx
+    # for more details. Since we do not have any .dll files in pxp-agent/bin, the system
+    # openssl is detected as pxp-agent's openssl library instead of Puppet's openssl, which
+    # is installed in [INSTALLDIR]\Puppet\puppet\bin. Of course, pxp-agent is not built to work
+    # with the system openssl, so it fails to start properly.
+    #
+    # By copying the dependent .dll files in the puppet/bin directory to pxp-agent/bin,
+    # we ensure that Windows uses our in-house .dll files to start up pxp-agent.
+    #
+    # See https://tickets.puppetlabs.com/browse/PA-1850 for all the details.
+    pkg.install do
+      dependent_dlls = [
+        "leatherman_curl.dll",
+        "leatherman_execution.dll",
+        "leatherman_file_util.dll",
+        "leatherman_json_container.dll",
+        "leatherman_locale.dll",
+        "leatherman_logging.dll",
+        "leatherman_nowide.dll",
+        "leatherman_util.dll",
+        "leatherman_windows.dll",
+        "libcpp-pcp-client.dll",
+        "libeay32.dll",
+        "ssleay32.dll",
+        platform.architecture == "x64" ? "libgcc_s_seh-1.dll" : "libgcc_s_sjlj-1.dll",
+        "libstdc++-6.dll"
+      ]
+
+      dependent_dlls.map do |dll|
+        "C:/cygwin64/bin/cp.exe #{settings[:prefix]}/bin/#{dll} #{settings[:pxp_root]}/bin"
+      end
+    end
+  end
+
   pkg.directory File.join(settings[:sysconfdir], 'pxp-agent')
   if platform.is_windows?
     pkg.directory File.join(settings[:sysconfdir], 'pxp-agent', 'etc', 'modules')


ekinanp:
I tested things out by doing the following:

(1) Installing an old version of the puppet-agent -- I used https://downloads.puppetlabs.com/windows/puppet-agent-1.10.10-x64.msi

(2) "Installing" openssl -- did this by copying a .dll file and naming it as libeay32.dll

(3) Using the puppetlabs-puppet_agent module to upgrade the agent to my dev package (here: http://builds.puppetlabs.lan/puppet-agent/c90ac595a23ce630f90a717cc3c58128598c577d/artifacts/windows/puppet-agent-x64.msi). I declared the class as:
class {'::puppet_agent':
package_version => "5.3.3.439.gc90ac59",
source => 'C:\Users\Administrator\puppet_agent_PA_1850.msi',
}
on my master VM.

(4) After the upgrade, I did an agent run. The version was OK -- 5.3.6 (b/c in the agent repo, we list the NEXT version to ship now). Note I cherry-picked this commit to the 5.3.x branch of the agent. The agent run went through without any issues. I also looked at services.msc and it was showing that the pxp-agent service was running.

NOTE just to double check, I repeated these steps using a released agent package for (3), and was able to repo. the paused pxp-agent service just like the customer -- the agent run in (4) also failed.

configs/components/pxp-agent.rb
```diff
@@ -82,6 +82,49 @@
...
+    # openssl is detected as pxp-agent's openssl library instead of Puppet's openssl, which
+    # is installed in [INSTALLDIR]\Puppet\puppet\bin. Of course, pxp-agent is not built to work
+    # with the system openssl, so it fails to start properly.
+    #
+    # By copying the dependent .dll files in the puppet/bin directory to pxp-agent/bin,
+    # we ensure that Windows uses our in-house .dll files to start up pxp-agent.
+    #
+    # See https://tickets.puppetlabs.com/browse/PA-1850 for all the details.
+    pkg.install do
+      dependent_dlls = [
```
ekinanp: These dependencies were obtained by using http://www.dependencywalker.com/ on the pxp-agent.exe file per @Iristyle's suggestion.

ekinanp: What's interesting is that none of the openssl dlls (libeay32.dll and ssleay32.dll) appear in the list of dependencies for the 1.10.x branch.

In the 5.3.x branch, we have these same dependencies plus:
  leatherman_curl.dll
  libstdc++-6.dll
  libeay32.dll

(The other OpenSSL dll, ssleay32.dll, does not appear there).

ekinanp: https://github.com/ekinanp/puppet-agent/blob/PA-1850-5.3.x/configs/components/pxp-agent.rb#L93-L106 shows all the dependencies on the 5.3.x branch -- ssleay32.dll is not a dependency there, but I added it anyways since it's an OpenSSL dll.

Iristyle: Since this list of dependencies could change in the future, is there a more way to specify this list? I would expect these to be outputs of the OpenSSL build process and referenced as such.

@McdonaldSeanp you know if this is possible in Vanagon?

McdonaldSeanp: @Iristyle no there's no way in vanagon to output information in one component and consume that info in another component.

We had to make similar changes to this for ruby and we had to do the same thing (where we just list them statically).

TBH I think this is the number 1 reason to try and get https://tickets.puppetlabs.com/browse/PA-761 working for puppet6, because this entire issue would go away.

ekinanp:
jenkins please test all on windows10ent-64a,windows10ent-32a

ekinanp:
@McdonaldSeanp The acceptance tests failed all-around on pxp-agent, which I don't think is related to anything done in this PR (since it failed on RHEL and Ubuntu). Everything else passed. If you want, I can re-run the PR pipeline again excluding pxp-agent so we can merge this while it's green (or even just running the puppet-agent acceptance tests so things finish faster).

I have also downloaded and installed the built agent package (http://builds.puppetlabs.lan/puppet-agent/f6bcf7a7ca4c6fc70b1f2222c817ae54fa787d5f/artifacts/windows/puppet-agent-1.10.10.68.gf6bcf7a-x64.msi) on a Windows machine and verified that the .dll files were copied over to pxp-agent/bin.

McdonaldSeanp:
No need to re-run it, I count that as a passing run (the pxp-agent failures are due to the issue Scott G. is fixing)

branan:
Everything about this makes me sad, and I wish we could just do PA-761 instead. But for now it looks like it'll do the needful :+1: 

I'd like to wait for our CentOS images to be fixed up so we can see a run of this go through before merging, though.

ekinanp:
Pending a merge on this PR until we figure out our CentOS imaging issues and also get a passing PR testing run.

Iristyle:
@branan FWIW it's not uncommon for Windows apps to duplicate their libraries across multiple directories under "Program Files" to avoid DLL Hell issues - in particular have a look at https://en.wikipedia.org/wiki/DLL_Hell#Running_conflicting_DLLs_simultaneously. So don't feel too bad, it's just a thorny issue at times. Fortunately disk space is cheap now, so we're not burdened by the constraints of 100MB hard disks anymore!

@McdonaldSeanp and I discussed the PA-761 approach in a bit more detail yesterday as a long-term solution to some of our problems... I'm hopeful we can move forward on cleaning things up, but we just have to be careful about the broader impact across our ecosystem, which I've tried to enumerate on the ticket.

geoffnichols:
jenkins please test all on windows10ent-64a,windows10ent-32a

ekinanp:
jenkins please test all on windows10ent-64a,windows10ent-32a

ekinanp:
Nice, PR testing passed. This should be good to go.

PR #1339, opened 2018-03-06
(maint) Augeas URL should be external

stahnma:
Prior to this the augeas url and mirror url were identical, which means
that anybody attempting to build Puppet Agent not on our VPN couldn't
download augeas.


caseywilliams:
Ah, it looks like this could go into 5.3.x instead - it seems we have the same problem there

stahnma:
Ok, I'll close and retarget. (just reworking via the UI was a terrible idea).

PR #1341, opened 2018-03-06
(maint) Augeas URL should be external

stahnma:
Prior to this the augeas url and mirror url were identical, which means
that anybody attempting to build Puppet Agent not on our VPN couldn't
download augeas.


stahnma:
So the checks here that are failing aren't from my commit 🤷‍♂️

PR #1343, opened 2018-03-07
(PA-1896) Enable builds for Debian 9 on armhf

stahnma:
This commit enables builds for Debian 9 armhf, which is the target for
running raspian 9 (raspberry PI Debian). This build also has the
advantage of not requiring anything from pl-build-tools as this compiles
all natively using the xcc toolchain provided by Debian.

Since this uses system yaml-cpp and boost, it does add runtime
requirements on those libraries (as well as libicu).

## Points of note

* New run time dependencies for this agent (system boost, yaml-cpp, icu)
* New leatherman compile flag (USE_ICU)
* <s>This should be pretty easy to flip the rest of Debian/Ubuntu to use this setup. </s>
* Removes all usages of pl-build-tools. Also does not use the "runtime" component in PA. 

## Future Work

*<s> Look at all Deb builds this way.</s>
* Try a build not on the VPN (since all this stuff *should* be available)

Actual build available: 
 http://unsupported.s3.amazonaws.com/puppet-agent_5.4.0.412.g8f7446e2-1stretch_armhf.deb

### DIFF
--- a/configs/components/augeas.rb
+++ b/configs/components/augeas.rb
@@ -6,6 +6,14 @@ component 'augeas' do |pkg, settings, platform|
 
   pkg.replaces 'pe-augeas'
   pkg.build_requires "libxml2"
+
+  if platform.name =~ /debian-9-armhf/
+    pkg.build_requires "libreadline-dev:#{platform.architecture}"
+    pkg.build_requires "pkg-config"
+    pkg.environment "CFLAGS", settings[:cflags]
+    pkg.environment "LDFLAGS", settings[:ldflags]
+  end
+
   if platform.name =~ /^el-(5|6|7)-.*/ || platform.is_fedora?
     # Augeas needs a libselinux pkgconfig file on these platforms
     pkg.build_requires 'ruby-selinux'
--- a/configs/components/cpp-hocon.rb
+++ b/configs/components/cpp-hocon.rb
@@ -12,8 +12,14 @@ component "cpp-hocon" do |pkg, settings, platform|
     cmake = "/usr/local/bin/cmake"
     special_flags = "-DCMAKE_CXX_FLAGS='#{settings[:cflags]}'"
   elsif platform.is_cross_compiled_linux?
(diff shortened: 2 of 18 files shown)


stahnma:
Putting on don't merge label because of a few minor touch ups.

stahnma:
After researching building 32-bit (i386) binaries on x86_64(amd64), basically we can't until Buster is the stable Debian. This is not a gcc problem, but a debian packaging issue with dh_strip with needs i686-linux-gnu-objcopy and that's not available until Buster, because it wasn't common to cross-compile for 32 bit when you could just run it (or use pbuilder). That puts a dead-end on moving all builds to amd64 base, but doesn't mean we still can't drop pl-build-tools.

stahnma:
Ready for review.

stahnma:
This combined with #1341 means you can build armhf off the VPN. 💃 🎉 🎈

ScottGarman:
I'm really excited about the potential to eliminate the pl-build-tools bootstrapping process, as it would dramatically reduce our time and effort to bring up a new agent platform, not to mention be much friendlier to community efforts. So I'm :+1: on this PR. 

I am concerned though about when and how we decide to adopt this as the Way of the Future, because if we don't have a plan for that, we'll be maintaining yet another special case build process within the codebase, which can be cumbersome to maintain. Hence I'd like to get more people to chime in on this, especially @branan.

jhg03a:
This is a major step in helping to manage the Internet of Security Holes devices.

stahnma:
@ScottGarman My goal would be we put this on master and do it as part of the Puppet 6 push. I am nearly 100% we can move all debian-based builds to drop pl-build-tools. From there, I'll start looking into Fedora and EL. 

We won't be able to all of pl-build-tools due to things like AIX, and crazy architectures in some cases, but this should make the happy path (and some other paths) happier.

stahnma:
obsoleted by #2055

PR #1345, opened 2018-03-08
Merge 5.3.x -> 5.4.x

McdonaldSeanp:


configs/components/puppet-runtime.rb
```diff
@@ -0,0 +1,84 @@
+component 'puppet-runtime' do |pkg, settings, platform|
+  # This component is a placeholder for the future migration of several vendor
+  # dependencies into a separate project, puppet-runtime. It is not included in
+  # the agent project yet.
+  runtime_details = JSON.parse(File.read('configs/components/puppet-runtime.json'))
+  runtime_tag = runtime_details['ref'][/refs\/tags\/(.*)/, 1]
+  raise "Unable to determine a tag for puppet-runtime (given #{runtime_details['ref']})" unless runtime_tag
+  pkg.version runtime_tag
+
+  tarball_name = "agent-runtime-1.10.x-#{pkg.get_version}.#{platform.name}.tar.gz"
```
caseywilliams: This project name will need to differ for each branch - seems like we can address it separately, though, since this component isn't used yet

PR #1353, opened 2018-03-12
(PA-1245) Split cfacter source/precompiled in to seperate projects

McdonaldSeanp:
We should be able to independently build the cfacter source and precompiled
so this commit adds the cfacter source project, and then updates the cfacter
project (which builds precompiled gems) to not ship the source gem as
well

### DIFF
--- a/configs/components/cfacter-precompiled-gem.rb
+++ b/configs/components/cfacter-precompiled-gem.rb
@@ -41,6 +41,7 @@ component "cfacter-precompiled-gem" do |pkg, settings, platform|
   pkg.install do
     [
       "pushd #{settings[:gemdir]}",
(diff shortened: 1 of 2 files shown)


McdonaldSeanp:
I looked at that and I dislike that it checks out itself to get the settings (i.e. vanagon has no 'local' option for inheriting settings, so it will literally check out the repo it's already in to grab settings)

ekinanp:
Makes sense. What about an instance_eval? Something like `cfacter-common`, `cfacter-source`, `cfacter`?

McdonaldSeanp:
TBH I would rather put in a feature request to vanagon to add a local option to inherit_settings and use that when we get it.

PR #1355, opened 2018-03-13
(FACT-1547) Build working cfacter gem for windows

McdonaldSeanp:
This commit makes a few necessary changes to facilitate the cfacter gem on
windows:

Paths have been updated and settings added to allow the puppet-runtime to
work.

The glob of files for the file list in the precompiled gem has been moved to
an erb template so the windows gem can include both lib/ and bin/ dirs

the windows gem now copies both the lib and bin dirs (since on windows
facter.rb expects to reside in lib/ and call out to bin/)

### DIFF
--- a/configs/projects/cfacter.rb
+++ b/configs/projects/cfacter.rb
@@ -29,17 +29,22 @@ project "cfacter" do |proj|
   proj.setting(:project_version, gem_version)
   proj.setting(:gemdir, '/var/tmp/facter_gem')
   if platform.is_windows?
-    proj.setting(:ruby_dir, '/cygdrive/c/Program\\ Files/Puppet\\ Labs/Puppet/sys/ruby')
-    proj.setting(:gem_binary, 'cmd /c "C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\gem.bat"')
-    proj.setting(:ruby_binary, 'cmd /c "C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\ruby.exe"')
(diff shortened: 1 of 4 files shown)


configs/projects/cfacter.rb
```diff
@@ -29,17 +29,22 @@
   proj.setting(:project_version, gem_version)
   proj.setting(:gemdir, '/var/tmp/facter_gem')
   if platform.is_windows?
-    proj.setting(:ruby_dir, '/cygdrive/c/Program\\ Files/Puppet\\ Labs/Puppet/sys/ruby')
-    proj.setting(:gem_binary, 'cmd /c "C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\gem.bat"')
-    proj.setting(:ruby_binary, 'cmd /c "C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\ruby.exe"')
+    proj.setting(:artifactory_url, "https://artifactory.delivery.puppetlabs.net/artifactory")
+    proj.setting(:buildsources_url, "#{proj.artifactory_url}/generic/buildsources")
+    proj.setting(:ruby_dir, '/cygdrive/c/ProgramFiles64Folder/PuppetLabs/Puppet/sys/ruby')
```
caseywilliams: Will there be 32-bit builds for these gems? I'm thinking about `ProgramFilesFolder` vs `ProgramFiles64Folder` and whether that could make for any trouble - I don't really understand the windows process well enough to know, though.

McdonaldSeanp: there may be in the future but for now lets stick with the 64 bit build.

PR #1356, opened 2018-03-13
(PA-1908) Remove logrotate from MCollective

Sharpie:
MCollective configures the Ruby Logger to perform rotation in-process. This
behavior is enabled by default if the file_logger is used with 5 archived logs
retained and rotation occurring after 2 MB of logs. This behavior renders the
logrotate configuration for MCollective redundant. Aside from adding
complexity, certain versions of the logrotate configuration executed a full
restart of the daemon which could wake up daemons that had been disabled and
also created thundering herd issues as logrotate tends to execute around the
same time each day.

### DIFF
--- a/configs/components/marionette-collective.rb
+++ b/configs/components/marionette-collective.rb
@@ -25,7 +25,6 @@ component "marionette-collective" do |pkg, settings, platform|
   case platform.servicetype
   when "systemd"
     pkg.install_service "ext/aio/redhat/mcollective.service", "ext/aio/redhat/mcollective.sysconfig", "mcollective"
-    pkg.install_file "ext/aio/redhat/mcollective-systemd.logrotate", "/etc/logrotate.d/mcollective"
   when "sysv"
     if platform.is_deb?
       pkg.install_service "ext/aio/debian/mcollective.init", "ext/aio/debian/mcollective.default", "mcollective"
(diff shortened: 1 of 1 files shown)


Sharpie:
Ref. puppetlabs/marionette-collective#470

Sharpie:
Tested on CentOS 6 and CentOS 7, an upgrade removes the logrotate configuration in both cases.

MikaelSmith:
One note, the logger rotation doesn't seem to compress the files. However with a cap of 10MB, that doesn't seem like it'll cause problems.

geoffnichols:
jenkins please test this

MikaelSmith:
Looks like we could merge it. Anyone want to do that? My understanding is you want 2 approvals.

Sharpie:
I think 10 MB is fine. CentOS defaults to rotating logs weekly, so we are already hitting this cap in active installations and also "loosing" historical data as that weekly logrotate only grabs `mcollective.log` and not the rotated `mcollective.log.{0,1,2,3,4}` files.

The real pain point with MCollective logging has historically been ActiveMQ as it only stored 5 MB worth of logs which would roll very quickly when serving a population of 1000s of nodes.

In any case, the limit on how much logging data is retained can be configured via `server.cfg`.

PR #1364, opened 2018-03-14
(maint) Actually remove nonfinal_link_target from build_defaults

McdonaldSeanp:
Simply setting the string to empty will not work, so actually completely
remove it


ScottGarman:
:+1:

PR #1367, opened 2018-03-15
Merge 1.10 to 5.3

MikaelSmith:


MikaelSmith:
I think this is right but can you take a look @branan? There were some changes besides ones I merged that hadn't been merged up.

PR #1370, opened 2018-03-15
(FACT-1549) Update cfacter gems to be pre-release facter gems

McdonaldSeanp:
After realizing ruby dependency managemant would make the name change to
cfacter a problem, we have decided to use pre-release functionality in
rubygems and bundler and release cfacter as a 'pre-release' facter gem.

For users, this will mean they can use:

gem 'facter', '~> 3.11.0.cfacter'

and opt in to the cfacter gem.

Note that we will still use an extra .rc for pre-release cfacter gems, meaning
for a pre-release cfacter gem the version would look like:

  3.11.0.cfacter.rc.YMD

  And this will still function like a normal pre-release for cfacter (i.e.
  requiring 3.11.0.cfacter will not pull 3.11.0.cfacter.rc gems, but requiring
  3.11.0.cfacter.rc will pull them in)


McdonaldSeanp:
Cherry-pick of https://github.com/puppetlabs/puppet-agent/pull/1369 to mainline branches

PR #1374, opened 2018-03-16
(maint) Do not try to execute mcollective state file

ekinanp:
The mcollective state file is not a scrip that returns data, it's a
file that needs to be dumped. Treat it as such.


ekinanp:
This is just a cherry-pick of https://github.com/puppetlabs/puppet-agent/commit/e8d7e8ed6f033a6bd23272e8201379c8056e5815

McdonaldSeanp:
🙈

PR #1382, opened 2018-03-26
(PDK-526) Add puppet-resource_api gem

DavidS:
```
david@davids:~/git/puppet-agent$ sudo dpkg -i output/deb/stretch/PC1/puppet-agent_5.5.0.89.g0c885104-1stretch_amd64.deb
[sudo] password for david:
dpkg: warning: downgrading puppet-agent from 5.5.0.89.g9adc14d0-1stretch to 5.5.0.89.g0c885104-1stretch
(Reading database ... 295203 files and directories currently installed.)
Preparing to unpack .../puppet-agent_5.5.0.89.g0c885104-1stretch_amd64.deb ...
Unpacking puppet-agent (5.5.0.89.g0c885104-1stretch) over (5.5.0.89.g9adc14d0-1stretch) ...
Setting up puppet-agent (5.5.0.89.g0c885104-1stretch) ...
Processing triggers for libc-bin (2.27-2) ...
david@davids:~/git/puppet-agent$ /opt/puppetlabs/puppet/bin/gem list --local | grep resource
puppet-resource_api (1.0.1)
david@davids:~/git/puppet-agent$ cd ../puppet-resource_api/spec/fixtures/modules
david@davids:~/git/puppet-resource_api/spec/fixtures/modules$ /opt/puppetlabs/bin/puppet resource --modulepath . test_simple_get_filter
test_simple_get_filter { "bar":
  ensure => 'present',
  test_string => 'default',
}
test_simple_get_filter { "foo":
  ensure => 'present',
  test_string => 'default',
}
david@davids:~/git/puppet-resource_api/spec/fixtures/modules$
```

### DIFF
--- a/configs/components/rubygem-puppet-resource_api.rb
+++ b/configs/components/rubygem-puppet-resource_api.rb
@@ -1,0 +1,22 @@
+component "rubygem-puppet-resource_api" do |pkg, settings, platform|
+  pkg.load_from_json("configs/components/rubygem-puppet-resource_api.json")
+
+  pkg.build_requires "puppet-runtime"
+
+  # Install into the directory for gems shared by puppet and puppetserver
+  pkg.environment "GEM_HOME", settings[:puppet_gem_vendor_dir]
+
+  # PA-25 in order to install gems in a cross-compiled environment we need to
+  # set RUBYLIB to include puppet and hiera, so that their gemspecs can resolve
+  # hiera/version and puppet/version requires. Without this the gem install
+  # will fail by blowing out the stack.
+  pkg.environment "RUBYLIB", "#{settings[:ruby_vendordir]}:$(RUBYLIB)"
+
+  pkg.build do
+    ["gem build puppet-resource_api.gemspec"]
+  end
+
+  pkg.install do
+    ["#{settings[:gem_install]} puppet-resource_api-*.gem"]
+  end
+end
--- a/acceptance/tests/validate_vendored_ruby.rb
+++ b/acceptance/tests/validate_vendored_ruby.rb
@@ -146,7 +146,7 @@ test_name 'PA-1319: Validate that the vendored ruby can load gems and is configu
   step "'irb' successfully loads gems" do
     agents.each do |agent|
       irb = irb_command(agent)
-      ['hocon', 'deep_merge'].each do |gem|
+      ['hocon', 'deep_merge', 'puppet/resource_api'].each do |gem|
         stdout = on(agent, "echo \"require '#{gem}'\" | #{irb}").stdout.chomp
         assert_match(/true/, stdout, "'irb' failed to require the #{gem} gem")
       end
@@ -168,4 +168,3 @@ test_name 'PA-1319: Validate that the vendored ruby can load gems and is configu
     end
   end
 end
-
--- a/configs/components/rubygem-puppet-resource_api.json
+++ b/configs/components/rubygem-puppet-resource_api.json
@@ -1,0 +1,4 @@
+{
+  "url": "git://github.com/puppetlabs/puppet-resource_api.git",
+  "ref": "9caab67cc2e347096bae3c93044d7d2da02e81ee"
+}
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -98,6 +98,7 @@ project "puppet-agent" do |proj|
   # Provides augeas, curl, libedit, libxml2, libxslt, openssl, puppet-ca-bundle, ruby and rubygem-*
   proj.component "puppet-runtime"
   proj.component "nssm" if platform.is_windows?
+  proj.component "rubygem-puppet-resource_api"
 
   # These utilites don't really work on unix
   if platform.is_linux?


branan:
@DavidS What's the expected update cadence of this gem in the agent? If it's infrequent, we can save some boilerplate by treating it as if it was a 3rd-party dependency and dropping it into the runtime instead of this project - we have a helper for Gems there. If it will be frequent, I think I'd rather we ensure it can be updated continuously from jenkins/travis/pipelines/whatever - either from every commit to the gem codebase, or at least from every tag.

We could go so far as to treat the gem as a full-fledged component of the agent. Given my understanding is that we eventually want to make Puppet and the Resource API work more closely together, this might be a worthwhile move?

Either way, if this is a gem that's also needed server-side, we have a new gem path that allows the server to load gems provided by the puppet-agent package (to avoid duplicating those gems in both packages). See https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/rubygem-multi_json.rb#L9-L10 for how we set that up.

Lastly: It looks like the only dependency of the resource api gem is hocon, which is already vendored in the agent. Is that correct? If there's anything else, it should be added explicitly so that we have a reproducible dependency tree.

DavidS:
@branan 

* expected update frequency is expected to be high in the next couple of months
* auto-update is separately ticketed in PDK-528, for today I'm fine with adding the update to the gem's release process
* When we get around to higher levels of integration between Core and Resource API, I do expet the gem to go away at one point, because it won't make sense to ship a Core piece in a separate deliverable. (The gem can be used back to puppet 4.7)
* I've changed the GEM_HOME path to the shared location
* I've added hocon as a requirement

In the future the gem might grow a dependency on childprocess and FFI, see PDK-847. But that's a long ways off.

DavidS:
cc/fyi @joshcooper

branan:
The more I think about this, the more I'm inclined to treat this like our other "core" components - that means automatic per-merge updates from day 1, and full integration into our tagging/shipping process. I don't want to lean on the "we're going to merge it into Puppet itself" crutch to avoid this work unless we're committed to a tight timeline for getting that work done. If so, we can discuss the tradeoffs of that timeline in more detail

* Any acceptance tests we add around the simplified API will be much more useful if we're integrating continuously (and I really would like to see at least a basic "it works" test added to this repo)
* If we /do/ uncover any resource_api bugs in tests we add here, a full release cycle of the gem is heavyweight to get a bugfix into the agent
* The agent deliverable should have a single stop-ship line. As this becomes part of core, we're likely to see other teams contributing to it. It's important to know when code needs to land, and having to account for a gem release process (including possibly vacations of those who normally ship it!) for this single component is a hassle
* As soon as we have multiple Y streams of this gem tied to multiple Y streams of the agent, with bugfixes possibly coming in from multiple teams, we'll really want that merge-up automation to be easy. Setting things up now to play nice with our agent tooling will make that trivial in the future

Being integrated with that process doesn't make out-of-band gem releases any more difficult, and it could just automate away regular z releases for the life of the gem (until it's properly merged into puppet).

DavidS:
Thanks to @smcelmurry https://github.com/puppetlabs/ci-job-configs/pull/4226 is a thing now. I'll rework this PR to have the proper JSON-loading component.

@branan what else needs to be put in place to make this happen? 

- [ ] acceptance tests
- [ ] process documentation

DavidS:
This now has the JSON loading necessary for lining up with the jenkins promotion jobs. I expect their first mergely run to fix the ref to something more standard.

The final piece I'm still unclear on, and need input from someone with experience here, is whether and which acceptance tests we need for this integration. As also mentioned in https://github.com/puppetlabs/ci-job-configs/pull/4226#discussion_r177682272 , the gem has a number of "acceptance" tests in https://github.com/puppetlabs/puppet-resource_api/tree/master/spec/acceptance exercising the resource api through calling `puppet resource` and friends. Those tests run on travis against a number of puppet versions (including master) on each PR, and are required to pass before a merge. It seems reasonable to re-run them "here" against the final package.

DavidS:
@branan the second commit has another problem I've run into when trying to build this. I would be grateful for hints how to address this.

branan:
@DavisS all we need for testing in this repo is a simple smoke test - lay down a minimal dummy module and confirm `puppet resource foo` works, or something in that direction. Just enough to verify that the gem is being packaged correctly and can be loaded by Puppet. I'll look in to what's failing to build here, and follow up before my day is done. Hopefully that'll give you everything you need for your AM tomorrow.

branan:
jenkins please test this

branan:
@DavidS The build issue is due to https://github.com/puppetlabs/puppet-resource_api/blob/master/puppet-resource_api.gemspec#L14-L16. Vanagon does not copy the .git directory to the builders, so git operations do not work. You might be able to just use Ruby's Dir.glob there?

DavidS:
jenkins please test this

DavidS:
@branan I fixed the gem building in https://github.com/puppetlabs/puppet-resource_api/pull/63 and https://github.com/puppetlabs/puppet-resource_api/pull/61 , and I've added the most basic test to the local acceptance suite that the gem loads.

for more in-depth checking, I'd need to write, build, publish, and install a module using the Resource API.

DavidS:
@branan fix for that is pending travis (thanks, jruby) at https://github.com/puppetlabs/puppet-resource_api/pull/65

DavidS:
jenkins please test this

DavidS:
https://github.com/puppetlabs/puppet-resource_api/pull/65 has been merged and addresses the previous build failure.

branan:
jenkins please test this

branan:
@DavidS if you can get the tests green during your workday, I think we can merge this first thing in the AM portland time.

DavidS:
jenkins please test this

DavidS:
:tada: No idea what did not align yesterday night. This morning I forced it to use the current master's SHA and it ran through :man_shrugging: 

> After 5pm, we write the bugs of tomorrow.

configs/components/rubygem-puppet-resource_api.rb
```diff
@@ -0,0 +1,22 @@
...
+  pkg.environment "GEM_HOME", settings[:puppet_gem_vendor_dir]
+
+  # PA-25 in order to install gems in a cross-compiled environment we need to
+  # set RUBYLIB to include puppet and hiera, so that their gemspecs can resolve
+  # hiera/version and puppet/version requires. Without this the gem install
+  # will fail by blowing out the stack.
+  pkg.environment "RUBYLIB", "#{settings[:ruby_vendordir]}:$(RUBYLIB)"
+
+  pkg.build do
+    ["gem build puppet-resource_api.gemspec"]
```
MikaelSmith: I think this needs to use a full path. `gem` doesn't appear to be available on Solaris.

MikaelSmith: Something like `["#{settings[:host_gem]} build puppet-resource_api.gemspec"]`

MikaelSmith: Fixed by https://github.com/puppetlabs/puppet-agent/pull/1404.

DavidS: Thanks!

PR #1386, opened 2018-03-28
Merge up 5.5.x to master

caseywilliams:
This mergeup applies the use of puppet-runtime to the master branch - the equivalent project files in puppet-runtime that replace the removed files are [here](https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/agent-runtime-master.rb) and [here](https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/base-agent-runtime.rb).

### DIFF
--- a/configs/components/puppet.json
+++ b/configs/components/puppet.json
@@ -1,1 +1,1 @@
-{"url":"git://github.com/puppetlabs/puppet.git","ref":"5846904f9f5c9ea65c87d51d5e0dd6d087b48bff"}
+{"url":"git://github.com/puppetlabs/puppet.git","ref":"7435bf2a6504c01d0c9cf57c83bb62bf6f80b842"}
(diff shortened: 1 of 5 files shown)


configs/components/puppet.json
```diff
@@ -1 +1 @@
-{"url":"git://github.com/puppetlabs/puppet.git","ref":"5846904f9f5c9ea65c87d51d5e0dd6d087b48bff"}
+{"url":"git://github.com/puppetlabs/puppet.git","ref":"7435bf2a6504c01d0c9cf57c83bb62bf6f80b842"}
```
melissa: don't we want to avoid merging these up?

caseywilliams: Yes, but 7435bf2a is the current ref in master: https://github.com/puppetlabs/puppet-agent/commit/83c757c4006e7f501da0dd61a61c6faf0dbe88c4 - I'm not exactly sure why this is showing up as a diff...

melissa: Maybe it's that you did this merge up before `7435bf2a6504c01d0c9cf57c83bb62bf6f80b842` was promoted? that is strange...

but yes, `7435bf2a6504c01d0c9cf57c83bb62bf6f80b842` is the current ref in master and `5846904f9f5c9ea65c87d51d5e0dd6d087b48bff` was the previous ref. huh. Well, I'm fine with merging this as is

PR #1389, opened 2018-04-03
(maint) Update leatherman/facter rbconfig paths

caseywilliams:
puppet-runtime now allows shipping multiple rubies at once - the original
rbconfig in `datadir` for cross-compiled platforms and sparc now has a
suffix applied to designate the ruby version. This updates the rbconfig
path accordingly for facter and leatherman.


ScottGarman:
:+1:

PR #1394, opened 2018-04-09
(PA-2028) Update to support running debian under systemd or sysv

underscorgan:
Needs https://github.com/puppetlabs/vanagon/pull/684 and vanagon release
Systemd check was updated to the one used in systemd service provider.

### DIFF
--- a/configs/components/pxp-agent.rb
+++ b/configs/components/pxp-agent.rb
@@ -112,35 +112,37 @@ component "pxp-agent" do |pkg, settings, platform|
     pkg.directory File.join(settings[:logdir], 'pxp-agent'), mode: "0750"
   end
 
-  case platform.servicetype
-  when "systemd"
-    pkg.install_service "ext/systemd/pxp-agent.service", "ext/redhat/pxp-agent.sysconfig"
-    pkg.install_configfile "ext/systemd/pxp-agent.logrotate", "/etc/logrotate.d/pxp-agent"
-    if platform.is_deb?
-      pkg.add_postinstall_action ["install"], ["systemctl disable pxp-agent.service >/dev/null || :"]
+  platform.get_service_types.each do |servicetype|
+    case servicetype
+    when "systemd"
+      pkg.install_service "ext/systemd/pxp-agent.service", "ext/redhat/pxp-agent.sysconfig", init_system: servicetype
+      pkg.install_configfile "ext/systemd/pxp-agent.logrotate", "/etc/logrotate.d/pxp-agent"
+      if platform.is_deb?
+        pkg.add_postinstall_action ["install"], ["if [ -d '/run/systemd/system' ] ; then systemctl disable pxp-agent.service >/dev/null || :; fi"]
+      end
+    when "sysv"
+      if platform.is_deb?
+        pkg.install_service "ext/debian/pxp-agent.init", "ext/debian/pxp-agent.default", init_system: servicetype
+        pkg.add_postinstall_action ["install"], ["update-rc.d pxp-agent disable > /dev/null || :"]
+      elsif platform.is_sles?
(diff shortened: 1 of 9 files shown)


underscorgan:
This is now supported in vanagon.

branan:
Please move this to a PA ticket, rather than a VANAGON ticket, so we can track it in our releases

branan:
I'm also still pretty disinclined to merge this until we figure out what our testing/supportability story is, FWIW

McdonaldSeanp:
I agree with @branan, we need to determine what expectations are around our support in this situation from both us and customers.

We need to be able to identify if this scenario requires it's own row in test matrices and what that would look like before I'm comfortable merging

configs/components/pxp-agent.rb
```diff
@@ -142,34 +142,36 @@
...
-      pkg.install_service "ext/suse/pxp-agent.init", "ext/redhat/pxp-agent.sysconfig"
-    elsif platform.is_rpm?
-      pkg.install_service "ext/redhat/pxp-agent.init", "ext/redhat/pxp-agent.sysconfig"
+  platform.get_service_types.each do |servicetype|
+    case servicetype
+    when "systemd"
+      pkg.install_service "ext/systemd/pxp-agent.service", "ext/redhat/pxp-agent.sysconfig", init_system: servicetype
+      pkg.install_configfile "ext/systemd/pxp-agent.logrotate", "/etc/logrotate.d/pxp-agent"
+      if platform.is_deb?
+        pkg.add_postinstall_action ["install"], ["if [ -d '/run/systemd/system' ] ; then systemctl disable pxp-agent.service >/dev/null || :; fi"]
```
mwaggett: Why do we need this check if we're already in the "systemd" case?

mwaggett:
Seems fine to me from a code perspective, but I will defer to @branan and @McdonaldSeanp for whether adding this is a good idea rn.

configs/components/pxp-agent.rb
underscorgan: For debian platforms with both sysv and systemd you need to add all of the scripts, and then run the scripts based on what's currently on the system. You only know you're in the systemd section at build time, but this can change at runtime so it needs to be in the scripts.

mwaggett: oookay! thanks!

olifre: This broke automated deployment for us with Ubuntu, since `/run/systemd/system` does not exist in the installer environment, so services will not be activated anymore at all since this change.

underscorgan: @olifre Is there a way in your environment to tell which init system it's running? We want to be able to support deb-ish systems running either sysv/upstart or systemd

olifre: @underscorgan As outlined in the long-standing https://tickets.puppetlabs.com/browse/PA-136 , which is about the very same issue concerning runtime behaviour of the agent for the same very same failing test against the runtime directory, the correct / portable check appears to be to test whether `/sbin/init` links to `/lib/systemd/systemd`. This works in Ubuntu and Debian installer chroots, and I still have to see a systemd system where it fails ;-).

underscorgan: Honestly, I don't remember why this was needed here. If this work is going to move forward I can investigate further, but I was working on this too long ago to remember edge cases I ran into while testing.

ciprianbadescu:
passing ad-hoc: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/926/

gimmyxd (approved):

GabrielNagy (approved):
good to merge for me, after pxp-agent logrotate for launchd is readded

PR #1396, opened 2018-04-10
(PA-1847) Add puppet docs to MANPATH

smcclellan:
When the user installs puppet-agent, by default, the man pages are not
included in the manpath. This updates the path through the
puppet-agent.sh script in profile.d.


smcclellan:
@caseywilliams I've updated the branch.

PR #1397, opened 2018-04-10
(maint) Switch to acceptance setup from beaker-puppet

melissa:
This commit removes the acceptance setup steps defined in this repo in
favor of those defined in beaker-puppet.


melissa:
@caseywilliams I was testing this with https://github.com/puppetlabs/beaker-puppet/pull/36

I did remove quite a bit, but I'm wondering what functionality we absolutely have to preserve. We don't use `TESTING_RELEASED_PACKAGES` or `AGENT_DOWNLOAD_URL` anywhere else in testing components. You also have different rake task names that have the same functionality. Would you be willing to remove those old tests in favor of the ones that are used by puppet/facter/hiera/pxp-agent?

branan:
Standardizing is good. We'll need to update ci-job-configs to remove a special-case for the agent CI task, but otherwise I don't think anything critical relies on this project having different names from everything else.

caseywilliams:
jenkins please test this

acceptance/tests/validate_vendored_ruby.rb
```diff
@@ -144,7 +145,7 @@ def install_dependencies(agent)
     agents.each do |agent|
       irb = irb_command(agent)
       ['hocon', 'deep_merge'].each do |gem|
-        stdout = on(agent, "echo \"require '#{gem}'\" | #{irb}").stdout.chomp
+        stdout = on(agent, "echo \"require '#{gem}'\" | #{irb}").stdout.chomp)
```
caseywilliams: I think this trailing paren is the source of the syntax error in the PR tests -- @melissa would you mind making an update? Other than that I'm excited to merge this

melissa: I think there was a different reason we were dragging our feet on this, I can't quite remember it since it's been a while....

caseywilliams: Unless there's something I don't know about, I believe [this](https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/vanagon-integration.sh#L78) is the only place we'll need to update CI configuration to make this work -- @branan do you remember anything else?

McdonaldSeanp: Yeah, the CI configuration is just that. And we really need to standardize this

melissa: Oh right! Yeah, the conversation about needing to standardize the rake tasks. I updated this line, so if that's the only thing blocking this.... then we can standardize rake tasks at a later date

melissa:
jenkins please test this

PR #1402, opened 2018-04-13
(maint) Avoid unnecessary duplication

MikaelSmith:
It's highly unlikely that Leatherman or cpp-pcp-client DLLs will be
installed to a Windows system directory. Also, work is underway to move
to statically linking cpp-pcp-client, and this makes doing so
transparently difficult. Only copy common DLLs from OpenSSL and MinGW.


McdonaldSeanp:
👍

MikaelSmith:
It's only a problem for me on master, and might be considered a breaking change.

PR #1404, opened 2018-04-13
(maint) Use built ruby for gem build

MikaelSmith:
Previously `gem build` relied on the system Ruby, which is not present
on all platforms - particularly Solaris and Windows. Use puppet-runtime
ruby instead.


MikaelSmith:
jenkins please test this

PR #1413, opened 2018-04-25
(PA-1989) Fix `RUBYLIB` in vendored ruby acceptance on Windows

caseywilliams:
Attempting to require the puppet resource_api in the vendored ruby
acceptance tests failed on Windows cygwin hosts because Beaker was
unable to set RUBYLIB correctly in that context. This is a workaround;
BKR-1445 and BKR-986 exist to address this problem.

### DIFF
--- a/acceptance/tests/validate_vendored_ruby.rb
+++ b/acceptance/tests/validate_vendored_ruby.rb
@@ -146,12 +146,26 @@ test_name 'PA-1319: Validate that the vendored ruby can load gems and is configu
     end
   end
 
-  step "'irb' successfully loads gems" do
+  step "vendored ruby successfully loads gems" do
     agents.each do |agent|
       irb = irb_command(agent)
+
       ['hocon', 'deep_merge', 'puppet/resource_api'].each do |gem|
-        stdout = on(agent, "echo \"require '#{gem}'\" | #{irb}").stdout.chomp
-        assert_match(/true/, stdout, "'irb' failed to require the #{gem} gem")
+        if agent['platform'] =~ /windows*/
+          # Although beaker can set a PATH for puppet-agent on Windows, it isn't
+          # able to set up the environment for the purpose of loading gems this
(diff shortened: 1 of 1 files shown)


caseywilliams:
jenkins please test this on windows2012r2-64a,redhat7-64a

acceptance/tests/validate_vendored_ruby.rb
```diff
@@ -146,12 +146,26 @@ def install_dependencies(agent)
...
-        assert_match(/true/, stdout, "'irb' failed to require the #{gem} gem")
+        if agent['platform'] =~ /windows*/
+          # Although beaker can set a PATH for puppet-agent on Windows, it isn't
+          # able to set up the environment for the purpose of loading gems this
+          # way; See BKR-1445 and BKR-986. As a workaround, run the environment
+          # script before requiring the gem. We're using `ruby -r` instead of
+          # irb here because `cmd /c` will only run one argument command before
+          # returning to cygwin (additional arguments will be treated as bash)
+          # and using irb introduces too many quotes.
+          cmd = "env PATH=\"#{agent['privatebindir']}:${PATH}\" cmd /c \"environment.bat && ruby -r#{gem} -e 'puts true'\""
```
joshcooper: I would expect the `environment.bat` modifications to the LOAD_PATH/RUBYLIB to only affect the `cmd.exe` process/interpreter, but not the `ruby` process that cygwin invokes? If it doesn't work, you may need to create a wrapper batch script that sources `environment.bat` and then invokes ruby:

```bat
call environment.bat
ruby -r #{gem} -e 'puts true'
```
And then invoke wrapper script using:

```bash
cmd = "env PATH=\"#{agent['privatebindir']}:${PATH}\" cmd /c wrapper.bat"
```

joshcooper: oh, I see that calling `environment.bat` and `ruby` are within the double quoted string passed to `cmd`, so that makes sense. 👍

ScottGarman:
:+1:

PR #1422, opened 2018-05-14
(PA-2031) Update puppet-agent to use packaging as a gem

sarameisburger:
Needs testing...~~I also need to make a puppet-agent specific ticket for this, when I do I will update the PR~~

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-def vanagon_location_for(place)
+def location_for(place)
   if place =~ /^(git[:@][^#]*)#(.*)/
     [{ :git => $1, :branch => $2, :require => false }]
   elsif place =~ /^file:\/\/(.*)/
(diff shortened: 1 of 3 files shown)


sarameisburger:
jenkins please test this

caseywilliams:
This looks good to me, but ooc, what's the reasoning behind targeting it at 5.3.x rather than 1.10.x or 5.5.x?

sarameisburger:
I thought 1.10.x used packaging branch master which I just checked and realized is not true. I can retarget!

sarameisburger:
Ah ok I remember now. The gemfile says 1.10.x is on packaging 1.0.x (https://github.com/puppetlabs/puppet-agent/blob/1.10.x/Gemfile#L14) but ext/build_defaults says packaging branch master (https://github.com/puppetlabs/puppet-agent/blob/1.10.x/ext/build_defaults.yaml#L3) so we'll probably want to test out pointing build_defaults at 1.0.x before we implement packaging as a gem there @caseywilliams

sarameisburger:
opened PR for that here https://github.com/puppetlabs/puppet-agent/pull/1460

sarameisburger:
jenkins please test this

PR #1424, opened 2018-05-15
(maint) Modules should be installed from gh repo

melissa:
This commit alters how we're handling modules during the install
process. Prior, we had been installing them from the forge. This commit
changes the install so that we're installing from a github repo. This
allows us to include development versions of modules in the puppet-agent
package, as we do for all other component projects included here.

This commit also modifies mailalias and maillist to read data from
corresponding json files.

### DIFF
--- a/configs/components/_base-module.rb
+++ b/configs/components/_base-module.rb
@@ -26,10 +26,6 @@ unless match_data
 end
 module_author, module_name = match_data.captures
 
-unless pkg.get_version
-  raise "You must set a version for the #{module_author}-#{module_name} module."
-end
-
 # Modules must have puppet
 pkg.build_requires 'puppet'
 
@@ -43,14 +39,8 @@ else
   pkg.directory(settings[:module_vendordir], mode: '0755', owner: 'root', group: 'root')
 end
 
-# Determine the download URL
-module_slug = "#{module_author}-#{module_name}-#{pkg.get_version}"
-tarball_name = "#{module_slug}.tar.gz"
-pkg.url("#{settings[:forge_download_baseurl]}/#{tarball_name}")
-
 pkg.install do
   [
-    "gunzip -c #{tarball_name} | #{platform.tar} -C #{settings[:module_vendordir]} -xf -",
-    "mv #{File.join(settings[:module_vendordir], module_slug)} #{File.join(settings[:module_vendordir], module_name)}"
+    "mv #{module_author}-#{module_name} #{File.join(settings[:module_vendordir], module_name)}"
   ]
 end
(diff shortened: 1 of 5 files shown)


melissa:
@caseywilliams @McdonaldSeanp @joshcooper I'm not sure if this is the pivot we want, but this is what all other components that auto-update do (read from a gh repo). I have a feeling we can install from the forge still, but I wanted to check in with y'all about this change first to gather opinions.

caseywilliams:
jenkins please test this

melissa:
@caseywilliams I'm not 100% sure this is necessary, but it looks like this is how everything else handles auto promotes (i.e., from github). We could tie the auto promote job to the forge release jobs, in which case we'd guarantee that the artifact is available on the forge. I'm not sure what would be best for us here...

caseywilliams:
@melissa this looks good to me -- I think I'll kick off a test on Solaris, though, since that's where we've had permissions trouble with the existing tarball approach. I think triggering promotions from forge release jobs is a good plan, too.

caseywilliams:
jenkins please test this on solaris10-64,solaris10-SPARC,solaris11-64a,solaris11-SPARC

caseywilliams:
jenkins please test this on solaris10-64,solaris10-SPARC,solaris11-64a,solaris11-SPARC

caseywilliams:
I'm not sure why the install setup step of the PR tests is failing - I don't see any changes in this PR that seem like they could have caused it. We'll look it, but fwiw I installed the finished packages on a few hosts:

Solaris 10 x86 (and sparc):
```
-bash-3.2# ls -lah /opt/puppetlabs/puppet/vendor_modules/    
total 4
drwxr-xr-x   4 root     root           4 May 15 16:24 .
drwxr-xr-x  10 root     sys           11 May 15 16:24 ..
drwxr-xr-x   5 root     sys           22 May 15 16:24 mailalias_core
drwxr-xr-x   5 root     sys           22 May 15 16:24 maillist_core
```

Solaris 11 sparc:
```
root@sol11-5:~# ls -lah /opt/puppetlabs/puppet/vendor_modules/
total 12
drwxr-xr-x   4 root     root           4 May 15 16:32 .
drwxr-xr-x  10 root     bin           11 May 15 16:32 ..
drwxr-xr-x   5 root     bin           22 May 15 16:32 mailalias_core
drwxr-xr-x   5 root     bin           22 May 15 16:32 maillist_core
```

Do those permissions make sense? They seem consistent with other subdirectories. I'm inclined to just merge this unless anyone else has objections.

joshcooper:
I've run into this also recently, but the beaker-puppet will only install on nodes that have the agent role, and since "a" is missing from most of the platforms in "solaris10-64,solaris10-SPARC,solaris11-64a,solaris11-SPARC", the agent is never installed. I'm not sure if the fix is to use "a" in the string when specified or if the automation should always add one?

speedofdark:
jenkins please test this on solaris10-64a,solaris10-SPARCa,solaris11-64a,solaris11-SPARCa

PR #1427, opened 2018-05-21
(maint) use GATech alternate mirrors for Solaris10

speedofdark:


speedofdark:
Needs to be retargeted to 1.10.x and auto-retargeting is too complex.

PR #1432, opened 2018-05-22
(CPR-391) Fedora packages have odd distro tag

mwaggett:
This commit updates the fedora platform configs to not use the 'f' prefix in
versions and, instead, use 'fc<version>' as the dist tag, which is standard for
fedora packages.


mwaggett:
should go in with https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/232

mwaggett:
@branan @McdonaldSeanp https://github.com/puppetlabs/ci-job-configs/pull/4519

kurtwall:
Wow. How did I get notified about this?

PR #1433, opened 2018-05-23
Mergeup 1.10.x to 5.3.x

ScottGarman:
Includes the OpenCSW artifactory revert commit.


speedofdark:
:+1:

configs/components/puppet-runtime.json
```diff
@@ -1 +1 @@
-{"url":"git://github.com/puppetlabs/puppet-runtime.git","ref":"refs/tags/201805210"}
+{"url":"git://github.com/puppetlabs/puppet-runtime.git","ref":"refs/tags/201805220"}
```
caseywilliams: Tag update seems suspect, but considering that 201805220 is newer, we probably wanted this anyway?

caseywilliams: Ah, looks like it's already set to that value in 5.3.x -- not sure why github is highlighting this

PR #1435, opened 2018-05-24
(PA-2061) Add nagios_core component

jhelwig:


jhelwig:
Not re-packaging it into puppet-agent certainly makes things easier.

PR #1436, opened 2018-05-24
(maint) Add zfs_core module to PA

melissa:


joshcooper:
jenkins please test this with agent master

branan:
Please use a PA ticket. Our release tooling gets sad when it can't associate tickets with the right fixversions in Jira

melissa:
I've removed the reference to a PUP ticket in the title

melissa:
@branan since this PR references a sub-ticket, I can't move that ticket to the PA project. Is this a sufficient commit title to get this merged in?

branan:
Yeah, good enough for the ticket management tools to not be sad. Ideally we'd have a ticket in the git history to track the work on this end, but it's not the end of the world. This is one of the many reasons I don't like Jira subtasks

melissa:
I agree, but subtasks seemed like the best tool at the time :( I apparently hadn't learned my lesson from the previous times I've been bitten by subtasks

PR #1437, opened 2018-05-29
(PA-2062) Remove hidden files from vendored modules

caseywilliams:
Removes any hidden files from directories in the `module_vendordir`, including .git/, travis config, etc.

I compared the results of removing hidden files with the the contents of module directories produced by PMT installs, and they seem to be equivalent (from an admittedly small sample). That said, some modules do include nonessential non-hidden things like an `appveyor.yml` even in the PMT-installed versions, so I left those alone for now.

### DIFF
--- a/configs/components/_base-module.rb
+++ b/configs/components/_base-module.rb
@@ -39,8 +39,13 @@ else
   pkg.directory(settings[:module_vendordir], mode: '0755', owner: 'root', group: 'root')
 end
 
+target_directory = File.join(settings[:module_vendordir], module_name)
+
 pkg.install do
   [
-    "mv #{module_author}-#{module_name} #{File.join(settings[:module_vendordir], module_name)}"
+    # Rename appropriately for use with puppet
+    "mv #{module_author}-#{module_name} #{target_directory}",
+    # Remove git and CI-related files
+    "rm -r #{File.join(target_directory, '.[!.]*')} 2>/dev/null",
   ]
 end


caseywilliams:
Jenkins please test this

configs/components/_base-module.rb
```diff
@@ -39,8 +39,13 @@
...
 
+target_directory = File.join(settings[:module_vendordir], module_name)
+
 pkg.install do
   [
-    "mv #{module_author}-#{module_name} #{File.join(settings[:module_vendordir], module_name)}"
+    # Rename appropriately for use with puppet
+    "mv #{module_author}-#{module_name} #{target_directory}",
+    # Remove git and CI-related files
+    "rm -r #{File.join(target_directory, '.*')} 2>/dev/null || true",
```
speedofdark: Is this executed in a shell context such that the "." is a literal?

caseywilliams: It doesn't remove non-hidden files, so it appears so -- I'll post some directory listings once the jenkins tests finish

speedofdark:
Excellent. Thank you.

configs/components/_base-module.rb
speedofdark: Ya know, upon a moment's reflection, thats a silly question. If there was *anything* left in the directory, then  it was operating with shell globbing.

caseywilliams:
The jenkins failure is due to job configuration rather than changes in this commit, which I'll work on figuring out, but here's the `vendor_modules/mailalias_core` directory on a machine where I installed the build:

```
root@ozx64g3g3dx0lag:~# ls -lah /opt/puppetlabs/puppet/vendor_modules/mailalias_core/
total 40K
drwxr-xr-x 4 root root 4.0K May 29 13:14 .
drwxr-xr-x 4 root root 4.0K May 29 13:14 ..
-rw-r--r-- 1 root root 1.3K May 29 12:37 appveyor.yml
-rw-r--r-- 1 root root 1.3K May 29 12:37 CHANGELOG.md
-rw-r--r-- 1 root root 3.8K May 29 12:37 Gemfile
drwxr-xr-x 3 root root 4.0K May 29 13:14 lib
-rw-r--r-- 1 root root 1.8K May 29 12:37 metadata.json
-rw-r--r-- 1 root root  275 May 29 12:37 Rakefile
-rw-r--r-- 1 root root 3.2K May 29 12:37 README.md
drwxr-xr-x 8 root root 4.0K May 29 13:14 spec
```

For comparison, here's the same module installed via PMT:
```
root@ozx64g3g3dx0lag:~# ls -lah /etc/puppetlabs/code/environments/production/modules/mailalias_core/
total 44K
drwxr-xr-x 4 root root 4.0K May 21 12:57 .
drwxr-xr-x 3 root root 4.0K May 29 13:16 ..
-rw-r--r-- 1 root root 1.3K May 21 12:56 appveyor.yml
-rw-r--r-- 1 root root 1.3K May 21 12:56 CHANGELOG.md
-rw-r--r-- 1 root root 1.6K May 21 12:57 checksums.json
-rw-r--r-- 1 root root 3.8K May 21 12:56 Gemfile
drwxr-xr-x 3 root root 4.0K May 21 12:57 lib
-rw-r--r-- 1 root root 1.8K May 21 12:57 metadata.json
-rw-r--r-- 1 root root  275 May 21 12:56 Rakefile
-rw-r--r-- 1 root root 3.2K Apr 30 12:43 README.md
drwxr-xr-x 8 root root 4.0K May 21 12:57 spec
```

caseywilliams:
jenkins please test puppet-agent

configs/components/_base-module.rb
joshcooper: Why `|| true`? If a file in `.git` is read-only with the `rm` fail?

caseywilliams: This will fail the vanagon build when rm is unable to remove `.` or `..` unless it returns successfully. Although, come to think of it, it would be better to avoid this workaround completely -- I'll see what else will work on all our platforms.

branan: How about `find . -mindepth 1 -maxdepth 1 -name '.*' | xargs rm -f`

caseywilliams: hmm looks like on solaris at least `mindepth` and `maxdepth` are not a thing -- `rm -r .[!.]*` seems like it may work though

caseywilliams:
jenkins please test puppet-agent

PR #1438, opened 2018-05-29
(PA-2188) migrate docker build scripts for puppet-agent

underscorgan:
These were migrated from https://github.com/puppetlabs/puppet-in-docker

### DIFF
--- a/docker/puppet-agent-alpine/Dockerfile
+++ b/docker/puppet-agent-alpine/Dockerfile
@@ -1,0 +1,35 @@
+FROM alpine:3.4
+
+ENV PUPPET_VERSION="5.5.1" FACTER_VERSION="2.5.1"
+
+LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.name="Puppet Agent (Alpine)" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=$PUPPET_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      org.label-schema.build-date="2018-05-09T20:10:01Z" \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN apk add --no-cache \
+      ca-certificates \
+      pciutils \
+      ruby \
+      ruby-irb \
+      ruby-rdoc \
+      && \
+    echo http://dl-4.alpinelinux.org/alpine/edge/community/ >> /etc/apk/repositories && \
+    apk add --no-cache shadow && \
+    gem install puppet:"$PUPPET_VERSION" facter:"$FACTER_VERSION" && \
+    /usr/bin/puppet module install puppetlabs-apk
+
+# Workaround for https://tickets.puppetlabs.com/browse/FACT-1351
+RUN rm /usr/lib/ruby/gems/2.3.0/gems/facter-"$FACTER_VERSION"/lib/facter/blockdevices.rb
+
+ENTRYPOINT ["/usr/bin/puppet"]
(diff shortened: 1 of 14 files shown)


docker/puppet-agent/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.name="Puppet Agent" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=$PUPPET_AGENT_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppet-in-docker" \
+      org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      org.label-schema.build-date="2018-05-09T20:10:27Z" \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
```
mwaggett: Is there a reason that all these labels (and perhaps some additional Dockerfile info) can't/shouldn't be shared?

branan:
This should be a PA ticket, please, or our release tooling will complain that it cannot map the ticket to a fixversion.

Also: The linked ticket is very broad, it is part of an epic for *server* work, and all of the commits in this PR are single-sentence descriptions with no additional information or justification. Why is this code moving into this repo? Who is going to be expected to maintain it in the future?

docker/puppet-agent/Dockerfile
underscorgan: Docker doesn't allow that. 

This image is the puppet-agent-ubuntu image with different labels (you can tell the base image from the FROM statement)

mwaggett: yeah, I was wondering if certain things could be DRY'd up, but sounds like that's not really possible and potentially not ideal anyway.

underscorgan:
@branan I'll open a separate ticket for the agent-related work and update my commit messages.

The epic is relating to phase one of platform containerization. Most of the work is around enabling the server stack to run in docker, but it does involve some agent parts.

Code is being moved from puppet-in-docker so we can more easily tie in container builds to the main project's development and release cycles. There is more work to do there, but these are the initial pieces.

Maintenance is still an open question. There will be more discussions of that in the near term (Rob says there's something supposed to happen next week).

branan:
I'd like us to have a better idea of ownership and the future of containerized agents before we land this here - it may turn out that it doesn't make sense to belong in this repo (or even to exist at all). I'm going to flag this and #1439 as blocked pending our discussions next week.

underscorgan:
@branan if I update commits to reference PA-2188 can this go in or do you still need more information?  

Containers and container automation are currently maintained by the folks working on pupperware, but having the dockerfiles/etc for building each container living in the project repos makes it easier for us to automate things and maintains consistency with how we're doing this for all the other projects

branan:
Yeah, I think that's sufficient.

underscorgan:
This is ready for review now, though there's a failure in check commits (due to how long it's been since 5.5.x has been tagged)

docker/puppet-agent-alpine/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN apk add --no-cache \
+      ca-certificates \
+      pciutils \
+      ruby \
+      ruby-irb \
+      ruby-rdoc \
+      && \
+    echo http://dl-4.alpinelinux.org/alpine/edge/community/ >> /etc/apk/repositories && \
+    apk add --no-cache shadow && \
+    gem install puppet:"$PUPPET_VERSION" facter:"$FACTER_VERSION" && \
```
stahnma: With Puppet6, we will we need any other T&P/Modules now that they are separate? 

Also, does Puppet 6 work with Facter 2.5?

Also, why is this not Puppet 6?

stahnma: And *then* I saw the branch.

Should this whole thing just move to the 6 branch?

underscorgan: Are we going to need any 5.x containers? Since 6 is out now, I could move this, mostly was targeting 5.x since we have containers for the other projects in that stream

underscorgan:
Closing this in favor of the 6.0.x PR

PR #1439, opened 2018-05-29
(CPR-557) Add scripts to build puppet-agent docker images

underscorgan:


branan:
All of my comments in #1438 also apply here

underscorgan:
Going to fold this in to #1438

PR #1441, opened 2018-06-08
(PA-2019) Change CustomAction directory references

Iristyle:
- 7cdb9e8b57b93d44b60a4e29f96e40ffcf449d6f introduced a change to
   reference directories in CustomAction tables by their component
   name, rather than by their directory name.

   However, it seems that this is, in some cases, resulting in empty
   strings, which causes takeown / icacls to run against \* instead
   of the desired fully qualified directory. This can damage file
   permissions on a Windows machine in a very destructive manner.

   MSDN documentation https://msdn.microsoft.com/library/aa368609.aspx
   suggests that this is because:

   "
   Note that if a component is already installed, and is not
   reinstalled, removed, or moved during the current installation, the
   action state of the component is null and the string
   [$componentkey] evaluates to Null.
   "

### DIFF
--- a/resources/windows/wix/customactions.wxs.erb
+++ b/resources/windows/wix/customactions.wxs.erb
@@ -198,7 +198,7 @@ End If
     <!-- Due to PUP-6729, system may not have permission to modify DACL, so first take ownership -->
     <SetProperty
         Id="TakeownPuppetPermissions"
-        Value="&quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[$PuppetAppDir]\*&quot; /R /A /D N"
+        Value="&quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[%PROGRAMDATA]\PuppetLabs\puppet\*&quot; /R /A /D N"
         Before="TakeownPuppetPermissions"
         Sequence="execute">
         NOT (REMOVE~="ALL")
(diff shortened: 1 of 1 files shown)


Iristyle:
jenkins please test this on windows2012r2-64a,windows10ent-32a,windows2012r2_ja-64a%7Blocale=ja%7D,windows2016-64a,windows10ent-64a

resources/windows/wix/customactions.wxs.erb
```diff
@@ -198,7 +198,7 @@ End If
     <!-- Due to PUP-6729, system may not have permission to modify DACL, so first take ownership -->
     <SetProperty
         Id="TakeownPuppetPermissions"
-        Value="&quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[$PuppetAppDir]\*&quot; /R /A /D N"
+        Value="&quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[PuppetAppData]\*&quot; /R /A /D N"
```
Iristyle: This still does not work properly, under specific circumstances, due to as of yet unknown circumstances.

It will however work with:

```
Value="&quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[%PROGRAMDATA]\PuppetLabs\puppet\*&quot; /R /A /D N"
```

joshcooper: During a repair, I see the following, which I assume is what they mean about "the action state of the component is null":

```
MSI (s) (04:98) [03:24:27:695]: Component: PuppetAppDir; Installed: Local;   Request: Null;   Action: Null
```

Iristyle:
@joshcooper I think the commit message on 29b7161 is kind of weak if you have a better explanation there.

joshcooper:
@Iristyle I don't think the change from `NOT(REMOVE~="ALL")` to `NOT REMOVE` makes any semantic difference, because our MSI only has one feature. So either we're removing "ALL" (during the uninstall phase of an upgrade or a complete uninstall) or we're not removing anything (during an install, repair, install phase of an upgrade). But I think `NOT REMOVE` is easier to reason about.

Iristyle:
Alternate branch / PR is up at https://github.com/puppetlabs/puppet-agent/compare/1.10.x...joshcooper:dont-use-components?expand=1

Iristyle:
This was superseded by #1442

PR #1442, opened 2018-06-10
(PA-2075) Use ProgramData to resolve puppet's data directory

joshcooper:
7cdb9e8 introduced a change to reference directories in CustomAction
tables by their component name, rather than by their directory name.

However, during a repair `[$componentkey]` resolves to an empty string,
resulting in takeown & icacls running from the root drive, damaging
file permissions system-wide.

MSDN documentation https://msdn.microsoft.com/library/aa368609.aspx
suggests that this is because:

    Note that if a component is already installed, and is not
    reinstalled, removed, or moved during the current installation, the
    action state of the component is null and the string
    [$componentkey] evaluates to Null.

This commit modifies the custom action logic so that it:

 * Uses `[%ProgramData]` which resolves to the value of the ProgramData
   environment variable, and should always be available during all MSI
   stages, e.g. install, repair, upgrade. Note environment variables are
   case insensitive.
 * Includes `PuppetLabs\{code,puppet,facter,pxp-agent,mcollective}` as
   part of the path, so even if ProgramData resolves to empty string,
   the path will be puppet specific.
 * Checks if the directory exists before trying to modify permissions.
 * Changes the custom action condition to `NOT REMOVE` instead of
   `NOT(REMOVE~="ALL")`. Since puppet-agent MSIs only have a single
   feature, the REMOVE property is either absent (during an initial
   install, repair, or the install phase of an upgrade), or it is
   present and has the value "ALL" (during the uninstall phase of a
   major upgrade, or a complete uninstall). So the two forms are
   equivalent, but the `NOT REMOVE` form is easier to reason about.
 * Converts the `SetProperty` WIX elements to the more generic
   `CustomAction` with a `Property` attribute, as that matches other
   properties in our WIX file and moves all of the conditional logic
   and sequencing into one file. The resulting property-based custom
   actions remain as type 51.

Co-authored-by: Ethan J. Brown <ethan@puppet.com>

### DIFF
--- a/resources/windows/wix/customactions.wxs.erb
+++ b/resources/windows/wix/customactions.wxs.erb
@@ -196,13 +196,11 @@ End If
     <%-end-%>
 
     <!-- Due to PUP-6729, system may not have permission to modify DACL, so first take ownership -->
-    <SetProperty
-        Id="TakeownPuppetPermissions"
-        Value="&quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[$PuppetAppDir]\*&quot; /R /A /D N"
-        Before="TakeownPuppetPermissions"
-        Sequence="execute">
-        NOT (REMOVE~="ALL")
-    </SetProperty>
+    <CustomAction
+        Id="SetTakeownPuppetPermissions"
+        Property="TakeownPuppetPermissions"
+        Value="&quot;[%WINDIR]\System32\cmd.exe&quot; /c IF EXIST &quot;[%ProgramData]\PuppetLabs\puppet&quot; &quot;[%WINDIR]\System32\takeown.exe&quot; /F &quot;[%ProgramData]\PuppetLabs\puppet\*&quot; /R /A /D N"
+        Execute="immediate"/>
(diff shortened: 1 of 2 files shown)


joshcooper:
jenkins please test this on windows2012r2-64a,windows10ent-32a,windows2012r2_ja-64a%7Blocale=ja%7D,windows2016-64a,windows10ent-64a

joshcooper:
Using SuperOrca, here's a diff between 1.10.13 and a version I built from f7bf1adc. The version goes "backwards" because I don't have the 1.10.13 tag locally.

The custom actions have the same type (51) and the target includes `%ProgramData` and the `IF EXIST` check:

CustomAction

The custom actions have the same sequence numbers as before and the conditions are `NOT REMOVE`:

InstallExecuteSequence

The `Property` table changes are expected due to version strings.

Property

The `File` table has different versions and a different libfacter.so:

File

The `Component` and `MsiFileHash` tables are different because during a major upgrade we assign new component Ids.

Iristyle:
Supersedes #1441

PR #1449, opened 2018-06-13
(PA-1995) Use location & version in puppet-runtime.json

caseywilliams:
Replaces the git ref and url values in puppet-runtime.json with version
and location values instead. This obviates the need to push, tag, and
ship runtime revisions before using them to build an agent; The README
has been updated with instructions on using a local puppet-runtime
output directory as a tarball source instead of Puppet's internal build
sources.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ def vanagon_location_for(place)
   end
 end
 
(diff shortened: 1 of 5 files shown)


caseywilliams:
jenkins please test this

PR #1459, opened 2018-06-21
(PA-2055) Update puppet-agent to dynamically link boost/yaml-cpp

McdonaldSeanp:
Now that we are using boost and yaml-cpp components in puppet-runtime (and not
using pl-boost or pl-yaml-cpp) We need to update the agent to dynamically link
against everything. For some reason builds fail on leatherman when the new
boost (boost 1.67.0) is statically linked.


branan:
@MikaelSmith https://github.com/puppetlabs/puppet-runtime/pull/64

McdonaldSeanp:
jenkins please test this on aix61-POWERa,sles12-64a,solaris11-SPARC,windows2012r2-64a

McdonaldSeanp:
The windows build doesn't understand how to find boost. I'm looking in to it.

PR #1461, opened 2018-06-26
Revert "(PA-2070) Match RPM build paths to shipped repo paths"

caseywilliams:
This reverts commit 6d815014e701a5b30c78427e715b7207e35cdf8e. Other tooling isn't ready for these changes yet.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -87,9 +87,9 @@ platform_repos:
   - name: sles-12-ppc64le
     repo_location: repos/sles/12/**/ppc64le
   - name: fedora-26-x86_64
-    repo_location: repos/fedora/26/**/x86_64
+    repo_location: repos/fedora/f26/**/x86_64
   - name: fedora-27-x86_64
-    repo_location: repos/fedora/27/**/x86_64
+    repo_location: repos/fedora/f27/**/x86_64
   - name: debian-8-i386
     repo_location: repos/apt/jessie
   - name: debian-8-amd64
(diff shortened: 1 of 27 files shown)


caseywilliams:
jenkins please test this

ext/build_defaults.yaml
```diff
@@ -87,9 +87,9 @@ platform_repos:
   - name: sles-12-ppc64le
     repo_location: repos/sles/12/**/ppc64le
   - name: fedora-26-x86_64
-    repo_location: repos/fedora/26/**/x86_64
+    repo_location: repos/fedora/f26/**/x86_64
   - name: fedora-27-x86_64
-    repo_location: repos/fedora/27/**/x86_64
+    repo_location: repos/fedora/f27/**/x86_64
```
mwaggett: I'm not really sure if you need these or not, since idk what you actually use these values for, but I don't *think* this affects the promote/compose issues at all.

caseywilliams: Hmm, it would be good to keep the non-f versions -- I'll update these and re-test with fedora

mwaggett:
This seems fine to me. 

Only thing I want us to be aware of is that as we make these path changes, we often use the repo name (e.g. puppet5 vs PC1) to determine whether to use old paths or new paths. Right now, I don't think that should matter (or ideally in the long long run), but the short-term fix for this that Britt was working on for this DOES make that assumption, so if that's what we decide is the best way to resolve this for now, we'll need to coordinate changes since you're keeping puppet5 as the repo name (which I think is fine).

caseywilliams:
jenkins please test this on fedora26-64a,fedora27-64a

PR #1475, opened 2018-07-05
(maint) Remove maillist and mailalias

joshcooper:


caseywilliams:
[Tests will need to be updated too, I think](https://github.com/puppetlabs/puppet-agent/blob/master/acceptance/tests/validate_vendored_modules.rb#L25)

joshcooper:
Ah thanks @caseywilliams, forgot about that. I'll update it.

joshcooper:
jenkins please test this on windows2012r2-64a

PR #1490, opened 2018-07-19
Add Release Lead helpers

smcclellan:
This adds improvements to automate engineering release lead duties:

- A rake task for finding platform additions and removals.
- A rake task for checking component versions included at a SHA for puppet-agent.
- A script for smoke testing puppet-agent from pre-release packages.
- A script for smoke testing puppet-agent from released repos.

### DIFF
--- a/ext/release-lead.rake
+++ b/ext/release-lead.rake
@@ -1,0 +1,118 @@
+require 'json'
+require 'optparse'
+
+namespace :release_lead do
+  desc "Find platforms added and removed between releases"
+  task :platform_diff, [:from, :to] do |t, args|
+    abort ('`from` argument is required') unless args.from
+    abort ('`to` argument is required') unless args.to
+    puts `git diff --summary #{args.from}..#{args.to} configs/platforms`.
+        gsub('configs/platforms/', '').gsub(/mode \d+ /, '').
+        gsub('delete', 'Retired').gsub('create', 'Added').
+        gsub('rename', 'Renamed')
+  end
+
+  # Calculate name of repository from the .git URL.
+  def url_to_component_name(url)
+    url.split('/').last.split('.').first
+  end
+
+  # Calculate name of repository from the name of the component's .json file.
+  def json_name_to_component_name(json_filename)
+    json_filename.split('/').last.split('.').first
+  end
+
+  # Input a git repository, a SHA or tag to check out, and a place where it can be cloned.
+  # Output the resulting `git describe` for that reference.
+  def git_describe_repo(name, url, sha_or_tag, where_to_clone, show_extra_commits)
+    Dir.chdir(where_to_clone) do
+      puts "Cloning #{name}..."
+      `git clone #{url} 2> /dev/null`
+      # `git clone --local '../../#{name}' 2> /dev/null`
+      # 0 for successful clone, 128 for already exists.
+      unless [0, 128].include? $?.exitstatus
+        next 'Could not find remote repository'
+      end
+      Dir.chdir(name) do
+        `git fetch 2> /dev/null`
+        `git checkout #{sha_or_tag} 2> /dev/null`
+        unless $?.exitstatus == 0
+          puts "Could not find reference #{sha_or_tag}"
+          exit
+        end
+        `git describe --tags`.strip +
+            if show_extra_commits
+              lines = `git log \`git describe --tags --abbrev=0\`..HEAD --oneline`.split("\n")
+              lines.map do |value|
+                if value.include?('[no-promote]')
+                  # Output these in red.
+                  "\e[31m#{value}\e[0m"
+                else
+                  value
+                end
+              end.join("\n  ")
+            else
+              ''
+            end
+      end
+    end
+  end
+
+  def check_components(args, show_extra_commits = false)
+    result = {}
+    abort('Error: puppet_agent_branch argument is required') unless args.puppet_agent_branch
+
+    where_to_clone = File.join(File.dirname(__FILE__), 'pkg')
+    Dir.mkdir(where_to_clone) unless Dir.exists?(where_to_clone)
+
+    # Let's ensure puppet-agent is on the right branch.
+    `git fetch`
+    begin
+      `git checkout #{args.puppet_agent_branch} 2> /dev/null`
+
+      Dir.glob("./configs/components/*.json").each do |component|
+        json = JSON.parse(File.read(component))
+        url = json['url']
+        if url.nil?
+          name = json_name_to_component_name(component)
+          result[name] = 'No URL provided'
+          next
+        end
+        name = url_to_component_name(url)
+        ref = json['ref']
+        sha_or_tag = ref =~ /^refs\/tags/ ? ref.gsub('refs/tags/', '') : ref
+        result[name] = git_describe_repo(name, url, sha_or_tag, where_to_clone, show_extra_commits)
+      end
+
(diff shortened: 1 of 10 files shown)


ext/smoke/repos/run-smoke-test.sh
```diff
@@ -0,0 +1,46 @@
...
+agent_vm2="$4$domain"
+
+echo "##### Setting up masters..."
+./steps/setup-masters.sh ${master_vm1} ${master_vm2} ${agent_version} ${server_version} ${puppetdb_version}
+
+# One agent starts with master 1, one agent starts with master 2
+echo "##### Master 1 (PuppetDB Module) + Agent 1"
+./steps/setup-agent.sh          ${master_vm1} ${agent_vm1} ${agent_version}
+./steps/run-validation-tests.sh ${master_vm1} ${agent_vm1}
+echo "##### Master 2 (PuppetDB Package)+ Agent 1"
```
ekinanp: If you insist on testing the combinations, you should bring it up in the release retro. Until everybody agrees this is the way we're smoke testing things, let's stick to Master 1 - Agent 1, Master 2 - Agent 2. (The agents are separate so that each master would have their own agent, start from a clean slate.). If we stick to that, the labels should just be changed to "Master (PuppetDB Module) + Agent".

ext/smoke/packages/steps/setup-master.sh
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
+
```
ekinanp: This should probably take the collection as a param. Johnson's going to be using puppet6.

ext/smoke/repos/steps/setup-masters.sh
```diff
@@ -0,0 +1,287 @@
...
+  echo ""
+  echo ""
+
+  start_puppetdb ${master_vm}
+  echo "Finished installing PuppetDB via. the module on ${which_master}!"
+  echo ""
+  echo ""
+}
+
+function install_puppetdb_from_package() {
```
ekinanp: This code could probably be shared with the repos test b/c the set-up from the package is the same (the only part that is different is the installation of the puppetdb packages, which I imagine we could probably pass in as a string).. Dunno if that should be done in this PR but I imagine you can source from another common bash file. I don't care if you do it, just a thought.

ext/smoke/packages/steps/setup-agent.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+set -e
```
ekinanp: This could also be re-used, aside from the installation of the agent package step (repo vs. raw RPM). Maybe an extra arg. could be added to differentiate if we're installing the raw package or from the repos and case it on that? Should be OK since we don't have any optional arguments here.

ext/smoke/repos/steps/run-validation-tests.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+
+set -e
+
+function on_host() {
```
ekinanp: These validation tests could be re-used too. They're pretty much the same. I think I just got lazy and wanted to copy-paste to put things in the same directory. Sorry :)

ext/release-lead.rake
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  # This task performs the following for each repository:
+  # - Looks up each component
+  # - Clones the component into the 'pkg' directory, if not done already
+  # - Checks out the git reference that's in the component's .json.
+  # - Outputs the `git describe` for that reference.
+  # This is useful for determining tag versions for an upcoming release.
+  desc "Output `git describe` for each component's git branch"
+  task :check_components_diff, [:puppet_agent_branch] do |t, args|
```
ekinanp: Why not have an option to show the commits? Seems weird having two separate tasks. Also, I wonder if it'd be worthwhile to:
```
- Have an option to show only one component
- Have an option to show the summary of each commit (hence why the previous option might be useful)
```

If people really want to see the full diff, they can use the git UI.

Don't have strong feelings, just a thought.

smcclellan: The shortcoming of adding an option is in how arguments are passed to rake... in this code, `:puppet_agent_branch` is the only argument passed in `[]`; using an option would mean invocation via `[5.5.x, <...?>]`. If arguments were named, I'd be fine with that, but it's not simple right now, so I split it into a separate task.

What do you mean by showing the summary of each commit, like the git commit message summary?

ekinanp: Ah I see. So rake doesn't let you do option flags then? Something like `rake task --opt1=val1 ...`?

Yup, that's what I mean by summary. So something like:
```
<commit_title>:
<commit_summary>
<commit_title>:
<commit_summary>
```

would be useful.

smcclellan: I think you have to do one or the other? I don't want to type out `puppet_agent_branch=...` for invocations if I can help it.

I see how commit title and description would be helpful, making it easier to hone in on what's changed in a certain component. I'll leave that as a possible improvement for the next time release lead duties come up.

ekinanp: Cool, sounds good.

ext/smoke/packages/steps/setup-agent.sh
ekinanp: Do we still need this? I see you have it at the top level.

smcclellan:
Removed the unused file, good eye.

geoffnichols:
Leaving a test comment here

ext/smoke/packages/steps/setup-master.sh
```diff
@@ -0,0 +1,102 @@
...
+set -e
+set -x
```
speedofdark: Should this be turned off before we promote?

ext/smoke/steps/run-validation-tests.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+
+set -e
+set -x
```
speedofdark: Should this be turned off before we promote?

ext/smoke/packages/steps/setup-master.sh
smcclellan: I think I prefer leaving it in; if it's gone, and e.g. an rpm command fails, it'd be helpful to know the exact command being executed to dive in, saving the need to repeat with the option enabled. A small amount of verbosity seems worth it, for now.

ext/smoke/steps/run-validation-tests.sh
smcclellan: Addressed above.

smcclellan:
Added the Don't Merge label until this release is complete.

smcclellan:
Some of the curl commands that download rpm files save an html 404 text file rather than throwing an error; not sure the best way to catch this, but it currently fails silently when it tries to install the rpm.

branan:
@smcclellan `curl -f` will fail for codes >= 400

smcclellan:
Updated to use `curl -f`.

ext/smoke/packages/steps/setup-master.sh
smcclellan: After a bit more consideration, this should indeed be removed. My concern was silent failures, but I think that will be solved by the `curl -f` fix suggested by Branan.

smcclellan:
Follow-up here: https://github.com/puppetlabs/puppet-agent/pull/1509

PR #1492, opened 2018-07-20
(maint) Add scheduled_task module to PA

melissa:
See PUP-8576


melissa:
jenkins please test this on windows2012r2-64

melissa:
screen shot 2018-07-20 at 2 31 07 pm

joshcooper:
One question is that the scheduled task module defaults to the `v2` provider. Will existing manifests continue to work? Or does the `v2` provider require param/properties to be specified?

PR #1495, opened 2018-07-25
(maint) Reduce choco log verbosity, replace deprecated update option

caseywilliams:
These changes match the ones applied in https://github.com/puppetlabs/puppet-runtime/pull/79:

- Removes download progress logging when calling `choco install`
- Uses `choco upgrade` instead of the deprecated `choco update`


caseywilliams:
Ah, I see this has already been taken care of! Closing this PR.

PR #1498, opened 2018-07-30
(PA-2055) Don't remove OpenSSL headers after building

caseywilliams:
Most headers are removed after building the last C++ component,
pxp-agent, to reduce the size of the final puppet-agent package. Some
donwstream vanagon projects that consume puppet-agent depend on OpenSSL
headers, so keep those.

(The eventual goal here is to produce separate puppet-agent (without headers) and puppet-agent-dev (with headers) packages, but until that time, we've removed as many headers as possible)

### DIFF
--- a/configs/components/cleanup.rb
+++ b/configs/components/cleanup.rb
@@ -1,0 +1,38 @@
+# This component exists only to remove unnecessary files that bloat the final puppet-agent package.
+component "cleanup" do |pkg, settings, platform|
+  # This component must depend on all other C++ components in order to be
+  # executed last, after they all finish building.
+  pkg.build_requires "cpp-pcp-client"
+  pkg.build_requires "cpp-hocon"
+  pkg.build_requires "facter"
+  pkg.build_requires "leatherman"
+  pkg.build_requires "libwhereami"
+  pkg.build_requires "pxp-agent"
+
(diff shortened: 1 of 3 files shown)


caseywilliams:
jenkins please test this on arista4-32a

caseywilliams:
jenkins please test this on arista4-32a

caseywilliams:
I've updated this to perform the header removals in their own dedicated component, which requires all the C++ projects, instead of cluttering the pxp-agent file.

configs/components/cleanup.rb
```diff
@@ -0,0 +1,38 @@
...
+    # the package to an unacceptable degree).
+    #
+    # Note that:
+    # - Ruby is not in this list because its headers are required to build native
+    #   extensions for gems.
+    # - OpenSSL is not in this list because its headers are required to build
+    #   other libraries (e.g. libssh2) in other projects that rely on
+    #   puppet-agent (e.g. pe-r10k-vanagon).
+    unwanted_headers = ["augeas.h", "boost", "cpp-pcp-client", "curl", "fa.h",
+                        "facter", "hocon", "leatherman", "libexslt", "libxml2",
```
branan: `libexslt`? Is that right?

PR #1500, opened 2018-08-01
(maint) Don't remove curl headers after build

caseywilliams:
pe-r10k-vanagon depends on puppet-agent and needs its curl headers for
libgit2.


caseywilliams:
jenkins please test this on arista4-32

PR #1503, opened 2018-08-03
(PA-1941) Fix libnames for new OpenSSL dlls

caseywilliams:
We need to copy openssl dlls around on windows. This commit updates the
names of those dlls since they changed in OpenSSL 1.1

Should be merged at the same time as https://github.com/puppetlabs/puppet-runtime/pull/82

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -211,10 +211,17 @@ component "facter" do |pkg, settings, platform|
     # Copy these into facter's bindir, they've already been copied into ruby's
     # in the runtime component
     pkg.install_file "#{settings[:tools_root]}/bin/zlib1.dll", "#{settings[:facter_root]}/bin/zlib1.dll"
+    if platform.architecture == "x64"
+      gcc_postfix = 'seh'
+      ssl_postfix = '-x64'
+    else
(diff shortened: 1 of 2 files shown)


configs/components/facter.rb
```diff
@@ -211,10 +211,17 @@
...
+      ssl_postfix = '-x64'
+    else
+      gcc_postfix = 'sjlj'
+      ssl_postfix = ''
+    end
     [
-      "libeay32.dll",
-      platform.architecture == "x64" ? "libgcc_s_seh-1.dll" : "libgcc_s_sjlj-1.dll",
-      "ssleay32.dll"
+      "libssl-1_1#{ssl_postfix}.dll",
```
ekinanp: Would be cool if this info. was loaded as part of the yaml output for the settings. Maybe something like `settings[:imported_libraries][:openssl]`.

ekinanp: Guess it won't matter once the PATH'ing stuff is fixed.

McdonaldSeanp: Yeah, at this point the plan is to get pathing fixed soon anyway, so the refactor is unnecessary

PR #1504, opened 2018-08-17
(maint) Update puppet to 60de165b86c5b00d409f590d52836f6be3ed7a0f

Magisus:
This version of puppet contains deprecations for webrick master related
settings, as well as rake file updates needed when running with JRuby
1.7.


smcclellan:
This may be done via pipelines instead.

PR #1509, opened 2018-08-22
(maint) Improve output of smoke test scripts

smcclellan:


smcclellan:
Added a commit that will copy the smoke test scripts' output to a predefined filename for ease of uploading to the ticket.

PR #1514, opened 2018-08-30
(PA-2157) Add support for sles-15-x86_64

ScottGarman:
This platform is being built with the default OS toolchain, and doesn't
rely on pl-build-tools packages.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -20,7 +20,7 @@ component "facter" do |pkg, settings, platform|
 
   pkg.build_requires 'puppet-runtime' # Provides openssl, ruby, augeas, curl
   pkg.build_requires 'leatherman'
-  pkg.build_requires 'runtime'
+  pkg.build_requires 'runtime' unless platform.name =~ /sles-15/
   pkg.build_requires 'cpp-hocon'
   pkg.build_requires 'libwhereami'
 
@@ -47,6 +47,9 @@ component "facter" do |pkg, settings, platform|
     pkg.build_requires "http://pl-build-tools.delivery.puppetlabs.net/aix/#{platform.os_version}/ppc/pl-yaml-cpp-0.5.1-1.aix#{platform.os_version}.ppc.rpm"
   elsif platform.is_windows?
     pkg.build_requires "pl-yaml-cpp-#{platform.architecture}"
+  elsif platform.name =~ /sles-15/
+    # These platforms use their default OS toolchain and have package
+    # dependencies configured in the platform provisioning step.
   else
     pkg.build_requires "pl-yaml-cpp"
   end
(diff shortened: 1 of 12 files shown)


configs/components/facter.rb
```diff
@@ -20,7 +20,7 @@
 
   pkg.build_requires 'puppet-runtime' # Provides openssl, ruby, augeas, curl
   pkg.build_requires 'leatherman'
-  pkg.build_requires 'runtime'
+  pkg.build_requires 'runtime' unless platform.name =~ /sles-15/
```
ekinanp: Why the special casing here?

configs/platforms/sles-15-x86_64.rb
```diff
@@ -0,0 +1,9 @@
+platform "sles-15-x86_64" do |plat|
+  plat.servicedir "/usr/lib/systemd/system"
+  plat.defaultdir "/etc/sysconfig"
+  plat.servicetype "systemd"
+
+  plat.provision_with "zypper -n --no-gpg-checks install -y aaa_base autoconf automake rsync gcc make rpm-build libboost_atomic1_66_0-devel libboost_chrono1_66_0-devel libboost_container1_66_0-devel libboost_date_time1_66_0-devel libboost_filesystem1_66_0-devel libboost_graph1_66_0-devel libboost_iostreams1_66_0-devel libboost_locale1_66_0-devel libboost_log1_66_0-devel libboost_math1_66_0-devel libboost_program_options1_66_0-devel libboost_random1_66_0-devel libboost_regex1_66_0-devel libboost_serialization1_66_0-devel libboost_signals1_66_0-devel libboost_system1_66_0-devel libboost_test1_66_0-devel libboost_thread1_66_0-devel libboost_timer1_66_0-devel libboost_wave1_66_0-devel gettext-tools cmake yaml-cpp-devel"
```
ekinanp: Nitpick, but could these packages be formatted like https://github.com/puppetlabs/puppet-runtime/blob/master/configs/platforms/sles-12-x86_64.rb#L7 ?

configs/components/facter.rb
ekinanp: Nevermind, I thought `runtime` == `puppet-runtime`.

configs/platforms/sles-15-x86_64.rb
ScottGarman: The way it's formatted currently is consistent with the other platform configs in this repo, so it might be better to change them all at once in a separate PR. I will say though that it's been handy to have the package list all on one line as I've needed to copy/paste it into a terminal frequently during development. I suppose that won't matter though once the platform is finalized.

ekinanp: Ok that makes sense. At least everything's in one place.

PR #1515, opened 2018-09-04
(PA-2157) Build dependency fix for sles-15-x86_64

ScottGarman:
Was missing the gcc-c++ package in the provisioning pkg list.


speedofdark:
:+1:

PR #1516, opened 2018-09-04
(PA-2000) Add sles-15-x86_64 to foss_platforms in build defaults

ScottGarman:


ext/build_defaults.yaml
```diff
@@ -85,6 +86,8 @@ platform_repos:
     repo_location: repos/sles/12/**/x86_64
   - name: sles-12-ppc64le
     repo_location: repos/sles/12/**/ppc64le
+  - name: sles-15-x86_64
+    repo_location: repos/sles/15/**/x86_64
```
caseywilliams: Does `rpm_targets` need to be updated below, too?

ScottGarman: I checked with RelEng, and it sounds like rpm_targets and deb_targets only list PE master platforms, and these vars are used in compose and promotion. So I wouldn't want to add sles-15 to that.

PR #1519, opened 2018-09-06
(MODULES-7753) Update scheduled_task ref for 1.0.0

michaeltlombardi:
This commit updates the reference to the `scheduled_task` module to 1.0.0 in
anticipation of the first supported release of that module.

### DIFF
--- a/acceptance/tests/validate_vendored_modules.rb
+++ b/acceptance/tests/validate_vendored_modules.rb
@@ -25,6 +25,7 @@ test_name "PA-1998: Validate that vendored modules are installed" do
       cron
       host
       mount
+      scheduled_task
       selboolean
       selmodule
       ssh_authorized_key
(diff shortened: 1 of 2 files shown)


caseywilliams:
This looks right -- You may want to update the acceptance test [here](https://github.com/puppetlabs/puppet-agent/blob/master/acceptance/tests/validate_vendored_modules.rb#L23) with the types from this module so we can check that they're actually there in integration, too

michaeltlombardi:
That's interesting, I'd expect them to be there already since the reference is already there. Will peek and update. :)

joshcooper:
I think we (coremunity) forgot to add the scheduled_task type to the acceptance test. Also we'll need to wait to merge this until the 1.0.0 version is tagged and released.

joshcooper:
Looks good, just added the blocked label until the scheduled_task module is tagged as 1.0.0

michaeltlombardi:
It should go live monday or tuesday/ :)

michaeltlombardi:
`scheduled_task` is released and tagged.

https://github.com/puppetlabs/puppetlabs-scheduled_task/tree/1.0.0

joshcooper:
I think we do want to target master (as is this PR) instead of 6.0.x

PR #1524, opened 2018-09-12
(PA-2158) Add el-7-ppc64 build support for 6.x

pdemonaco:
Hello,

This patch adds support for RHEL 7 on big-endian powerpc systems. In particular,
this has been built for a POWER8 environment running RHEL 7.5. Changes are
focused on adding the ppc64 architecture to various locations in the config
files which explicitly check for ppc64le instead.

Note that I do not run any instances of the 6.x series agent in my environment
as the series has yet to be released officially, however, it does build successfully.

I also have a set of similar changes available for the 5.5.x branch - should I create a separate pull request for those as well?

This request is related to both of the following requests:
* [puppet-runtime](https://github.com/puppetlabs/puppet-runtime/pull/93)
* [pl-build-tools-vanagon](https://github.com/puppetlabs/pl-build-tools-vanagon/pull/56)

Thanks!
Phil

### DIFF
--- a/configs/components/runtime.rb
+++ b/configs/components/runtime.rb
@@ -30,7 +30,7 @@ component "runtime" do |pkg, settings, platform|
 
   if platform.is_cross_compiled?
     libdir = File.join("/opt/pl-build-tools", settings[:platform_triple], "lib")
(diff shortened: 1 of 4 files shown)


pdemonaco:
Oh, that would be ideal. I actually made these changes to the 5.5.x branch originally but I wasn't sure how to correctly propagate them up to the master branch from there.

I'll make a clean version of my 5.5.x changes and resubmit.

PR #1525, opened 2018-09-12
(PA-619) open puppet command prompt in puppet appdata dir

McdonaldSeanp:
This commit updates the script used for "Start Command Prompt with Puppet" to
start in the appdata directory rather than C:\Windows\system32 (which is
where the script would start by default with no changes)

### DIFF
--- a/resources/windows/wix/shortcuts.wxs.erb
+++ b/resources/windows/wix/shortcuts.wxs.erb
@@ -57,7 +57,7 @@
           Target="[%ComSpec]"
           Arguments='/E:ON /K "[INSTALLDIR]bin\puppet_shell.bat"'
           Icon="PuppetShellShortcutIcon"
-          WorkingDirectory="bin">
+          WorkingDirectory="APPDATADIR">
           <Icon
             Id="PuppetShellShortcutIcon"
             SourceFile="wix\icon\puppet.ico"/>


McdonaldSeanp:
We would like to look around and see why things are not starting where they are supposed to from the WiX side

McdonaldSeanp:
jenkins please test this on windows2012r2-64a

resources/windows/wix/shortcuts.wxs.erb
```diff
@@ -57,7 +57,7 @@
           Target="[%ComSpec]"
           Arguments='/E:ON /K "[INSTALLDIR]bin\puppet_shell.bat"'
           Icon="PuppetShellShortcutIcon"
-          WorkingDirectory="bin">
+          WorkingDirectory="APPDATADIR">
```
joshcooper: I'm 👍 though I wanted to make sure y'all are aware that it's important that untrusted users not be able to write files into the current working directory, as it can lead to dll planting attacks. In this case, puppet's appdata directory is locked down, so it's ok:

```
C:\ProgramData\PuppetLabs>icacls .
. Everyone:(OI)(CI)(RX)
  NT AUTHORITY\SYSTEM:(OI)(CI)(F)
  BUILTIN\Administrators:(OI)(CI)(F)
```

PR #1530, opened 2018-09-18
(maint) Fix hardcoded smoke test references to puppet5 collection

ScottGarman:
This has been tested to work correctly with both puppet5 and puppet6
collections. It also includes a change in the command line to sign
all certs that was made in puppet6.


branan:
Since we'll (hopefully) never have to use this in a Puppet 4 release again, maybe we should derive collection automagically from the first digit of the agent package? But that's a separate PR

PR #1532, opened 2018-09-19
(PA-2158) Add el-7-ppc64 build support for 5.5.x

pdemonaco:
This patch adds support for RHEL 7 on big-endian powerpc systems. In particular, it has been built for a POWER8 environment running RHEL 7.5. Changes are focused on including ppc64 in all areas of the config which currently target ppc64le.

This is an alternative to #1524. @ScottGarman I reworked my original 5.5.x changes into a clean commit instead of rebasing my other branch that targeted 6.x.

Thanks,
Phil

### DIFF
--- a/resources/files/ruby_243/rbconfig/rbconfig-ppc64-redhat-linux.rb
+++ b/resources/files/ruby_243/rbconfig/rbconfig-ppc64-redhat-linux.rb
@@ -1,0 +1,313 @@
+# encoding: ascii-8bit
+# frozen-string-literal: false
+#
+# The module storing Ruby interpreter configurations on building.
+#
+# This file was created from a native ruby build on a RHEL 7.3 ppc64 system.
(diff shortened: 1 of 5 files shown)


resources/files/ruby_243/rbconfig/rbconfig-ppc64-redhat-linux.rb
```diff
@@ -0,0 +1,313 @@
+# encoding: ascii-8bit
```
branan: This file shouldn't be needed at all - rbconfigs are handled in the puppet-runtime project now, and should be automatically generated.

pdemonaco: True, however, they appear to still be referenced in the puppet-agent project. I included it for completeness.

PR #1533, opened 2018-09-24
(maint) Mark platforms which are community-supported

ScottGarman:
Add a note to the platform config for community-supported OS platforms.

And yes, the "s390" platform was never officially supported, just the "s390x" ones.


ScottGarman:
I assume it's safe for me to have targeted this at 1.10.x to make the merge-up process simpler?

PR #1537, opened 2018-09-25
(maint) cherry-pick release_lead tasks, add supported_platforms to it

McdonaldSeanp:
This PR does two things:

1. cherry-picks 67c52cba8884ea644cd879c652fdb529ec6b9ce3 to the 5.5.x branch

2. adds a supported_platforms task to list all platforms we have a platform file for and which ones are PE only

### DIFF
--- a/ext/release-lead.rake
+++ b/ext/release-lead.rake
@@ -1,15 +1,39 @@
 require 'json'
 require 'optparse'
+require 'yaml'
 
 namespace :release_lead do
   desc "Find platforms added and removed between releases"
   task :platform_diff, [:from, :to] do |t, args|
     abort ('`from` argument is required') unless args.from
     abort ('`to` argument is required') unless args.to
+    puts "Platform Changes:"
+    puts
     puts `git diff --summary #{args.from}..#{args.to} configs/platforms`.
         gsub('configs/platforms/', '').gsub(/mode \d+ /, '').
         gsub('delete', 'Retired').gsub('create', 'Added').
         gsub('rename', 'Renamed')
+  end
+
+  desc "List currently supported platforms on the current branch"
+  task :supported_platforms do |t, args|
+    puts "Currently Supported Platforms:"
+    puts
+    platform_names = Dir.glob("./configs/platforms/*").map do |platform_file|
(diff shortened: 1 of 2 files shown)


ext/release-lead.rake
```diff
@@ -0,0 +1,139 @@
...
+    puts "Platform Changes:"
+    puts
+    puts `git diff --summary #{args.from}..#{args.to} configs/platforms`.
+        gsub('configs/platforms/', '').gsub(/mode \d+ /, '').
+        gsub('delete', 'Retired').gsub('create', 'Added').
+        gsub('rename', 'Renamed')
+  end
+
+  desc "List currently supported platforms on the current branch"
+  task :supported_platforms do |t, args|
```
smcclellan: I like this addition; does it make sense to reference this on the [Confluence page](https://confluence.puppetlabs.com/display/systemsEngineering/Puppet+Agent+Release+Lead+Guide)?

ext/release-lead.rake
```diff
@@ -1,17 +1,38 @@
...
         gsub('configs/platforms/', '').gsub(/mode \d+ /, '').
         gsub('delete', 'Retired').gsub('create', 'Added').
         gsub('rename', 'Renamed')
   end
 
+  desc "List currently supported platforms on the current branch"
+  task :supported_platforms do |t, args|
+    puts "Currently Supported Platforms:"
+    puts
+    Dir.glob("./configs/platforms/*").map { |platform_file| File.basename(platform_file).gsub(".rb",'') }.each do |_platform|
```
smcclellan: I'd add a few line-breaks to split this logic up. Namely using `do` rather than `{}` in the `map` block.

ext/release-lead.rake
McdonaldSeanp: Yeah, it likely does. We need to mention on the confluence page that these tasks are not exact: sometimes names can look similar enough git thinks it's a rename when it's both an addition + removal.

Additionally supported_platforms won't list the platforms where we build one package for multiple supported platforms (AIX, Windows, el 7 (we use the 7 package for AWS) etc.)

gimmyxd:
Gonna close this as the rake tasks are on 6.x and main.

PR #1543, opened 2018-09-28
(PA-2188) Agent docker containers for 6.0.x

underscorgan:


docker/puppet-agent-alpine/Dockerfile
```diff
@@ -0,0 +1,130 @@
+FROM alpine:3.8 as build
+
+RUN apk add --no-cache cmake boost-dev make curl git curl-dev ruby ruby-dev gcc g++ yaml-cpp-dev
+
+RUN mkdir /workspace 
+WORKDIR /workspace
+RUN git clone -b 1.5.x https://github.com/puppetlabs/leatherman && \
```
stahnma: Why do this in one layer since it's a build image? You're throwing away each layer if any component changes.

docker/puppet-agent-centos/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+      org.label-schema.version="$PUPPET_AGENT_VERSION" \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppet-agent" \
+      org.label-schema.vcs-ref="$vcs_ref" \
+      org.label-schema.build-date="$build_date" \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN rpm -Uvh https://yum.puppetlabs.com/puppet6/puppet6-release-el-7.noarch.rpm && \
+    yum upgrade -y && \
+    yum update -y && \
```
stahnma: upgrade and update are the same thing in yum

docker/puppet-agent-debian/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+      org.label-schema.version=$PUPPET_AGENT_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppet-agent" \
+      org.label-schema.vcs-ref="$vcs_ref" \
+      org.label-schema.build-date="$build_date" \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends wget ca-certificates && \
+    wget https://apt.puppetlabs.com/puppet6-release-"$DEBIAN_CODENAME".deb && \
```
stahnma: Curl might be in base rather than having to install wget, but I'm not certain on that.

docker/puppet-agent-alpine/Dockerfile
underscorgan: Good point, will rework

docker/puppet-agent-debian/Dockerfile
underscorgan: Will check

underscorgan: nope, no curl

underscorgan:
depends on https://github.com/puppetlabs/puppet_docker_tools/pull/30

underscorgan:
That's been merged and a new gem has been released, so this is ready to go!

underscorgan:
@ScottGarman sure thing, will get some squashing done

underscorgan:
@ScottGarman updated!

underscorgan:
@stahnma I'm not sure, that's mostly what already existed. But maybe an alpine and an ubuntu would be plenty (so we have one with actual p-a packages, and the alpine one seems like it's useful in the built-from-source / smaller image ways). If that sounds reasonable I'll get rid of the others

stahnma:
I'd be +1 on that. Let's commit to maintain less, particularly if we think it's sufficient.

docker/puppet-agent-alpine/Dockerfile
```diff
@@ -0,0 +1,137 @@
...
+
+RUN apk add --no-cache cmake boost-dev make curl git curl-dev ruby ruby-dev gcc g++ yaml-cpp-dev
+
+RUN mkdir /workspace 
+WORKDIR /workspace
+RUN sed -i -e 's/sys\/poll/poll/' /usr/include/boost/asio/detail/socket_types.hpp
+
+RUN git clone -b 1.5.x https://github.com/puppetlabs/leatherman && mkdir -p /workspace/leatherman/build
+WORKDIR /workspace/leatherman/build
+RUN cmake  -DBOOST_STATIC=OFF -DCMAKE_VERBOSE_MAKEFILE=ON ..; make ; make install
```
Iristyle: We can defer to a subsequent PR, but shouldn't we be running `; make clean` and doing a few other operations to tidy up the disk contents to remove source / compile time tooling / intermediate output?

PR #1546, opened 2018-10-02
(maint) Remove mco reference

joshcooper:
The test was migrated from puppet to puppet-agent in 5.5.x, but we
removed mco in 6, so remove the expectation for those binaries.


joshcooper:
jenkins please test this on windows2012r2-64a with agent 6.0.x

PR #1547, opened 2018-10-09
(PUP-8351) Update bash wrappers to keep LD_LIBRARY_PATH with RHSCL

McdonaldSeanp:
When users want to make use of python (and other tools) from the redhat
software collections, those packages _do_ require that the LD_LIBRARY_PATH be
set to include directories containing those packages.

Since we unset LD_LIBRARY_PATH we have broken any customer using RHSCL
packages. This commit updates the wrapper script for puppet and friends to
search through each existing path in LD_LIBRARY_PATH for any path that starts
with /opt/rh/ (the canonical place for these packages to be installed) and
keep those paths in LD_LIBRARY_PATH as puppet and friends' run

### DIFF
--- a/resources/files/sysv-wrapper.sh
+++ b/resources/files/sysv-wrapper.sh
@@ -1,6 +1,13 @@
 #!/bin/sh
 
-unset LD_LIBRARY_PATH
+# We keep around any paths in LD_LIBRARY_PATH that start with /opt/rh/. Those paths are
+# from redhat software collections packages and are required for things like SCL python
+# to work. See PUP-8351.
+STRIP_LDLYP_COMMAND=" \
+if ENV['LD_LIBRARY_PATH']; \
+  print ENV['LD_LIBRARY_PATH'].split(':', -1).keep_if { |path| path.start_with?('/opt/rh/') }.join(':') \
+end"
+export LD_LIBRARY_PATH=$(/opt/puppetlabs/puppet/bin/ruby -e "$STRIP_LDLYP_COMMAND")
 unset LD_PRELOAD
 
 COMMAND=`basename "${0}"`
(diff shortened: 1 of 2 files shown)


McdonaldSeanp:
jenkins please test this on centos7-64a,centos6-64a

resources/files/sysv-wrapper.sh
```diff
@@ -1,6 +1,14 @@
 #!/bin/sh
 
+OLD_LD_LIB=${LD_LIBRARY_PATH//:/ }
```
branan: If any of the elements of `LD_LIBRARY_PATH` contain spaces, I think this is gonna be not-great. Maybe awk can help us here?

Supposedly this will also work, but I'm unsure if it's a bash-ism:

```
IFS=':'
OLD_LD_LIB="$LD_LIBRARY_PATH"
for path in $OLD_LD_LIB; do
 ...
done
unset IFS
```

branan: "IFS" is "internal field separator"

speedofdark: I'm reasonably sure it's a bash-ism.

McdonaldSeanp:
jenkins please test this on centos7-64a,centos6-64a,solaris11-64a,debian8-32a

resources/files/sysv-wrapper.sh
McdonaldSeanp: I realized that we have a ruby installed with our package, and ruby sucks way less then bash and CLI.

resources/files/sysv-wrapper.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
-
-unset LD_LIBRARY_PATH
+STRIP_LDLYP_COMMAND=" \
```
speedofdark: How did you come up with that variable name? :)

McdonaldSeanp:
Passed acceptance at https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_pr/746/

PR #1551, opened 2018-10-16
(maint) Fix sysv-wrapper.sh to work with Solaris 10

ekinanp:
On Solaris 10, /bin/sh uses the legacy Bourne shell syntax. This means
that some of the more advanced syntax like '$()' or 'export VAR=VALUE'
do not work.


ekinanp:
jenkins please test this on centos7-64a,centos6-64a,solaris10-64a,solaris11-64a,debian8-32a

PR #1576, opened 2018-11-14
Build committed versions of components and add jruby support

pcarlisle:


pcarlisle:
Waiting on #1575 first to change the docker context so it has access to the full repo

PR #1579, opened 2018-11-15
(maint) Build alpine with repo root as context

pcarlisle:


docker/Makefile
```diff
@@ -4,6 +4,8 @@ build_date := $(shell date -u +%FT%T)
 hadolint_available := $(shell hadolint --help > /dev/null 2>&1; echo $$?)
 hadolint_command := hadolint --ignore DL3008 --ignore DL3018 --ignore DL4000 --ignore DL4001
 hadolint_container := hadolint/hadolint:latest
+makefile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+makefile_dir := $(dir $(mkfile_path))
```
underscorgan:
```suggestion
makefile_dir := $(dir $(makefile_path))
```

docker/Makefile
```diff
@@ -38,7 +40,7 @@ build: prep
 	@docker build --pull --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppet-agent-ubuntu/$(dockerfile) --tag puppet/puppet-agent-ubuntu:$(version) puppet-agent-ubuntu
 	@docker tag puppet/puppet-agent-ubuntu:$(version) puppet/puppet-agent:$(version)
 ifneq ($(IS_NIGHTLY),true)
-	@docker build --pull --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppet-agent-alpine/$(dockerfile) --tag puppet/puppet-agent-alpine:$(version) puppet-agent-alpine
+	@docker build --pull --build-arg vcs_ref=$(vcs_ref) --build-arg build_date=$(build_date) --build-arg version=$(version) --file puppet-agent-alpine/$(dockerfile) --tag puppet/puppet-agent-alpine:$(version) $(makefile_dir)/..
```
underscorgan: I don't understand why you are changing this, can you give me more context?

docker/Makefile
pcarlisle: ah right that's why it's messing up

docker/Makefile
pcarlisle: The alpine Dockerfile tries to copy files from the puppet-agent repository. Docker only has access to files in its context, which is that directory passed, so it needs to be passed the whole repository.

pcarlisle:
I'm still working on running this locally so I'm not 100% sure it works

pcarlisle:
ok this works for me now

docker/puppet-agent-alpine/Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN apk add --no-cache cmake boost-dev make curl git curl-dev ruby ruby-dev gcc
 ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
 ENV PATH $PATH:$JAVA_HOME/jre/bin:$JAVA_HOME/bin
 
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
```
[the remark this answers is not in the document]
underscorgan: alpine doesn't have /bin/bash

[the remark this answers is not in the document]
pcarlisle: No, ash is one of the names of the shell on alpine. I use this name because if you leave /bin/sh here hadolint can't tell that you have a shell that supports pipefail.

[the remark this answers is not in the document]
tvpartytonight: 👍

PR #1582, opened 2018-11-17
Set up CI with Azure Pipelines

Iristyle:


Iristyle:
Should be able to close this once my pipeline config PR goes up...

PR #1583, opened 2018-11-17
(maint) Enable azure pipelines

Iristyle:
Follow-on PR will do the heavy lifting of building / testing

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+docker/puppet-agent-alpine/*               text  eol=lf
+docker/puppet-agent-ubuntu/*    text  eol=lf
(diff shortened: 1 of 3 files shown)


docker/ci/build.ps1
```diff
@@ -0,0 +1,38 @@
...
+  $thisName = $MyInvocation.MyCommand.Name
+  [IO.Path]::GetDirectoryName((Get-Content function:$thisName).File)
+}
+
+# installs gems for build and test and grabs base images
+function Invoke-ContainerBuildSetup
+{
+  Push-Location (Get-CurrentDirectory)
+  bundle install --path '.bundle/gems'
+  bundle exec puppet-docker update-base-images ubuntu:16.04
```
underscorgan: Do we want to move this to running `docker` directly since that's what we're doing for other builds/tests/etc?

PR #1585, opened 2018-11-20
(maint) Use 6.0.x-specific puppet-runtime for agent 6.0.x

caseywilliams:
There's no difference at this point between the 6.0.x and master runtimes, but we should use the matching version.


caseywilliams:
jenkins please test this

PR #1587, opened 2018-11-21
Add Windows Agent Container

RandomNoun7:
This change adds a build stage.

The current entry point is intended only to keep the container alive
while any manifests in the manifests folder are edited and then
executed while the entry point script spins in the background.

To build the container:
`cd .\docker\puppet-agent-windows`
`docker build --rm . -t puppet_agent:latest`

To execute the container and let it spin waiting for commands:
`docker run -d -v ${PWD}:C:\manifests -w c:\manifests puppet_agent`

To then execute a manifest and get the output:
`docker exec <container_name> powershell.exe puppet apply .\\manifests\\bill.pp`

At least those were reasonably accurate the last time I tested this. I
wouldn't be surprised if those commands needed to be jiggered slightly
to work properly.

### DIFF
--- a/docker/puppet-agent-windows/Dockerfile
+++ b/docker/puppet-agent-windows/Dockerfile
@@ -1,0 +1,57 @@
+FROM microsoft/windowsservercore:1803 as builder
+
+LABEL org.label-schema.maintainer="Puppet Windows Team <team-windows@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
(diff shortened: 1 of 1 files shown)


Iristyle:
We moved to WSL2 from LCOW on the Windows builders.. which means we lost the ability to build Windows containers.

Going to close this for now - but if we decide to revisit, we'd probably need to use GH actions to build / test the PRs.

PR #1590, opened 2018-11-28
Build containers in azure

Iristyle:


Iristyle:
So it looks like we've run head first into https://github.com/docker/for-win/issues/1357 which should be ultimately resolved in https://github.com/moby/moby/pull/37296.

For now, I'm going to leave code in for alpine builds, but disable in the Azure YAML. We can revisit once moby / LCOW has better memory configurability.

docker/ci/build.ps1
```diff
@@ -0,0 +1,106 @@
...
+  # tags required for versioning
+  git fetch origin 'refs/tags/*:refs/tags/*'
+  (git describe) -replace '-.*', ''
+}
+
+# installs gems for build and test and grabs base images
+function Invoke-ContainerBuildSetup
+{
+  Push-Location (Get-CurrentDirectory)
+  docker pull ubuntu:16.04
```
nwolfe: @underscorgan is this the same thing we figured out we could get rid of by using the `--pull` flag on our docker invocations?

Iristyle: Ah yeah, good point. I basically ported over the `Makefile`, but might have been working off an older version. If we can remove it, that'd be great.

underscorgan: yeah, we can `docker build --pull` for the same functionality

docker/ci/build.ps1
```diff
@@ -0,0 +1,106 @@
...
+  $subdir = if ($Base -eq 'alpine') { 'windows-build' } else { '' }
+  $docker_args = @(
+    '--pull',
+    '--build-arg', "vcs_ref=$Vcs_ref",
+    '--build-arg', "build_date=$build_date",
+    '--build-arg', "version=$Version",
+    '--build-arg', "namespace=$Namespace",
+    '--file', "puppet-agent-$Base/$subdir/Dockerfile",
+    '--tag', "$Namespace/puppet-agent-${Base}:$Version",
+    '--tag', "$Namespace/puppet-agent-${Base}:latest"
```
underscorgan: If we're just using these for local testing we probably don't need the latest tag

Iristyle:
Updated with a few new changes:

* Removed the explicit `docker pull` which had the nice side-effect of removing an Azure stage
* Removed passing a `--build-arg` of `namespace` since that doesn't exist
* Remove the old Ruby script for building the fake multi-stage builds


I think we'll be able to turn multi-stage on for Windows in the near future, but we need to resolve the memory situation first.

PR #1591, opened 2018-11-28
(maint) Update pathing for alpine docker image to be more consistent with agent pathing

underscorgan:


underscorgan:
@nwolfe nope, puppet-agent-ubuntu uses the puppet-agent packages so it will already have the normal paths. Probably need to do the windows alpine builds also

underscorgan:
I know you were running into some memory issues building on windows @Iristyle but here are some updates for the pathing in the windows build steps.

Iristyle:
I think we need to rebase this now because `build_alpine` has been removed.

docker/puppet-agent-alpine/Dockerfile
```diff
@@ -8,32 +8,32 @@ RUN sed -i -e 's/sys\/poll/poll/' /usr/include/boost/asio/detail/socket_types.hp
 
 RUN git clone -b 1.5.x https://github.com/puppetlabs/leatherman && mkdir -p /workspace/leatherman/build
 WORKDIR /workspace/leatherman/build
-RUN cmake  -DBOOST_STATIC=OFF -DCMAKE_VERBOSE_MAKEFILE=ON ..; make ; make install
+RUN cmake -DCMAKE_PREFIX_PATH=/opt/puppetlabs/puppet -DCMAKE_INSTALL_PREFIX=/opt/puppetlabs/puppet -DCMAKE_INSTALL_RPATH=/opt/puppetlabs/puppet/lib -DBOOST_STATIC=OFF -DCMAKE_VERBOSE_MAKEFILE=ON ..; make ; make install
```
Iristyle: Can we setup global defaults for these... or better to be explicit on each call?

underscorgan: I think setting up defaults would be reasonable, will take a look today.

underscorgan:
@Iristyle updated!

underscorgan:
@Iristyle updated!

Iristyle:
Looking good - thanks!

PR #1634, opened 2019-01-17
(PA-2362) Update fedora 29 to use GCC 8

denis-susan:
This change cleans up the build on fedora 29 to use gcc 8

### DIFF
--- a/configs/platforms/fedora-29-x86_64.rb
+++ b/configs/platforms/fedora-29-x86_64.rb
@@ -4,25 +4,10 @@ platform "fedora-29-x86_64" do |plat|
   plat.servicetype "systemd"
   plat.dist "fc29"
 
-  # Becasue GCC 8.2 outputs far more warnings than previous GCC versions,
-  # we need to use an older version to be able to build leatherman.
-  # To use GCC 7 we need to fetch it from Fedora 27 packages.
-  # leatherman ticket that adresses the issue: https://tickets.puppetlabs.com/browse/LTH-161
-  fedora_27_packages = [
-    "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_fedora/releases/27/Everything/x86_64/os/Packages/c/cpp-7.2.1-2.fc27.x86_64.rpm",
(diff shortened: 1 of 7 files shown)


configs/platforms/fedora-29-x86_64.rb
```diff
@@ -4,25 +4,11 @@
...
-    "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_fedora/releases/27/Everything/x86_64/os/Packages/l/libquadmath-7.2.1-2.fc27.x86_64.rpm",
-    "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_fedora/releases/27/Everything/x86_64/os/Packages/l/libquadmath-devel-7.2.1-2.fc27.x86_64.rpm",
-    "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_fedora/releases/27/Everything/x86_64/os/Packages/g/gcc-c++-7.2.1-2.fc27.x86_64.rpm",
-    "https://artifactory.delivery.puppetlabs.net/artifactory/rpm__remote_fedora-cache/updates/27/x86_64/Packages/r/redhat-rpm-config-77-1.fc27.noarch.rpm"
-  ]
 
-  fedora_27_packages.each { |package| plat.provision_with "dnf install -y #{package}" }
+  plat.provision_with "/usr/bin/dnf install -y --best --allowerasing autoconf automake rsync gcc gcc-c++ make rpmdevtools rpm-libs cmake"
+
+  plat.add_build_repository "http://pl-build-tools.delivery.puppetlabs.net/yum/pl-build-tools-release-#{plat.get_os_name}-29.noarch.rpm"
```
gimmyxd: I think this is not needed.

denis-susan: > I think this is not needed.

is it needed or not? can anyone else weigh in? Or can you argument why it's not needed because if i follow the link for pl-build-tools there are a lot of packages there.

denis-susan:
Passing ad-hoc: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/492

configs/platforms/fedora-29-x86_64.rb
gimmyxd: We are not using pl-build-tools anymore when building for new platforms. You can remove it

PR #1640, opened 2019-02-06
(maint) Allow use of FQDNs in smoke test scripts

caseywilliams:
The smoke test scripts originally required a hostname only, and
incorrectly appended a domain when given an FQDN. This adjusts them to
only append a domain if needed.

### DIFF
--- a/ext/smoke/helpers.sh
+++ b/ext/smoke/helpers.sh
@@ -201,3 +201,14 @@ function install_puppetdb_from_package() {
   echo ""
   echo ""
 }
+
+# Given a hostname or FQDN ($1), append a domain ($2, defaults to
+# '.delivery.puppetlabs.net') to form an FQDN if necessary
+function hostname_with_domain() {
+  host=$1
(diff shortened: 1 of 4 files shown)


ext/smoke/helpers.sh
```diff
@@ -201,3 +201,14 @@ function install_puppetdb_from_package() {
   echo ""
   echo ""
 }
+
+# Given a hostname or FQDN ($1), append a domain ($2, defaults to
+# '.delivery.puppetlabs.net') to form an FQDN if necessary
+function hostname_with_domain() {
+  host=$1
+  domain=${2:-.delivery.puppetlabs.net}
+  if [[ ! $host == *$domain ]]; then
```
branan: Instead of checking for the default domain, should we instead just check for a `.`? That way if somebody runs this against a platform9 VM or something it won't append `delivery.puppetlabs.net`

caseywilliams: Yeah, totally -- updated to just check for a dot somewhere

PR #1645, opened 2019-02-09
6.0.x merge -> 6.1.x

Iristyle:
Thanks to @mwaggett I believe this is no longer necessary and we want to merge #1646 instead. Please close if that's the case. Thanks!


Iristyle:
6.1.x branch no longer receives changes from 6.0.x, so closing this...

PR #1646, opened 2019-02-09
6.0.x mergeup to master

Iristyle:
Not sure if this is handled by automation, but put up to pick up some of the Docker changes we wanted to make.

Thanks!

I believe #1645 is not relevant and can be closed?


Iristyle:
I believe #1645 is not relevant and can be closed?

PR #1651, opened 2019-02-19
(maint) add sourcing the helpers script to the repos run-smoke-test.sh

speedofdark:
The script was updated to use helpers, but the sourcing of those helpers was missed.


caseywilliams:
Or maybe 6.0.x -- it looks like 5.5.x has this line, and maybe mergeups missed it going into 6.0.x

PR #1660, opened 2019-02-26
(PA-2499) Fix ssl linking issues for cmake based component builds

ScottGarman:
Our Solaris Sparc builds are no longer finding our vendored openssl
libraries. This adds the explicit paths to our libssl.so and
libcrypto.so to cmake's linker flags to workaround this problem.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -135,6 +135,7 @@ component "facter" do |pkg, settings, platform|
   elsif platform.is_solaris?
     if platform.architecture == 'sparc'
       ruby = "#{settings[:host_ruby]} -r#{settings[:datadir]}/doc/rbconfig-#{settings[:ruby_version]}-orig.rb"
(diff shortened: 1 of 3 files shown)


ScottGarman:
Here's an ad-hoc run I started against this PR: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/546/

There were some artifactory issues being fixed this afternoon which account for the Ubuntu failure. The Solaris Sparc error was due the inability of the system to verify the artifactory SSL cert, which I believe may be a known issue. 

The Solaris 11 x86 failure seems odd and may have something to do with errors running zfs on the host.

Some tests are stilling running, and I'm going to log off for the night. However if someone wants to matrix reload the ad-hoc pipeline and see if some of these errors are transients, feel free to do so. Thus far I haven't seen evidence that this PR is the cause of any problems.

PR #1664, opened 2019-03-07
(maint) Output modules changes in release script

tavilarion:
Output changes on core modules since the last release in check_components_diff script.

This is needed in order for the script to output if there were any changes on the modules since the last release, as commits to core modules are not automatically promoted.

### DIFF
--- a/ext/release-lead.rake
+++ b/ext/release-lead.rake
@@ -12,6 +12,10 @@ namespace :release_lead do
         gsub('rename', 'Renamed')
   end
 
+  def is_module?(component_file)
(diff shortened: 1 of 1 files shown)


caseywilliams:
I just changed the base branch to 5.5.x -- looks like it needs a rebase

caseywilliams:
(I went ahead and switched it back, since the diff got complicated and hard to review, but it seems like we should target this at the earliest branch possible among 5.5.x, 6.0.x, and master)

tavilarion:
Closing this since it was implemented in the new release project: 
https://github.com/puppetlabs/puppet-release_manager

PR #1667, opened 2019-03-12
(maint) Enable facter jruby support on RHEL 8

denis-susan:


Magisus:
Confirmed that this fixes the issue, with this puppetserver can start successfully.

Magisus:
We need this to go in before EOD today to make stop-ship.

caseywilliams:
This won't break anything, and since there is some urgency on the server end of things, I'm going to go ahead and merge it so we can get CI kicked off early. Thanks for the quick fix!

PR #1673, opened 2019-03-19
(maint) Merge 5.5.x -> 6.0.x 

loopinu:


caseywilliams:
I'm still not sure why the automatic [merge drivers](https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/vanagon-helpers.sh#L178) failed to ignore this newline difference, since they are supposed to automatically accept "ours" for [any json file](https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/vanagon-helpers.sh#L225). That said, this makes sense otherwise and I can merge once travis is green.

PR #1674, opened 2019-03-21
(MAINT) Manually update VERSION to 6.4.1

sebastian-miclea:
Y versions do not get updated in the Release Pipeline


caseywilliams:
Since this update is going to the release branch, it should have `6.4.0` as the version (that's the one we want to tag at the end of the release process).

It will also be necessary to make a similar update in the master branch, to version `6.4.1` -- that's the next possible Z version that we could release there after we finish 6.4.0 (It occurs to me now that this 6.4.1 update to master could just be made after the release is finished, too -- not as urgent).

PR #1683, opened 2019-03-26
(Maint) Merge 5.5.x -> 6.0.x

denis-susan:


caseywilliams:
This looks correct to me -- it seems that the container builds on azure may be based on the most recent tag, which has not yet shipped publicly -- these will keep failing until the 6.0.7 agent release finishes shipping. I'm ok with merging this once Travis turns green.

PR #1687, opened 2019-03-29
(PA-2561) Remove Cumulus Platform from build_defaults.yaml only

kris-bosland:
This change removes the Cumulus platform from ext/build_defaults.yaml.
Cumulus is not removed from acceptance/tests/validate_vendored_ruby.rb or
configs/platforms/cumulus-2.2-amd64.rb.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -2,7 +2,6 @@
 project: 'puppet-agent'
 # foss_platforms will be shipped to the nightly repos
 foss_platforms:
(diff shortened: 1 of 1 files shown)


speedofdark:
Please explain why defaults only? Thanks.

kris-bosland:
Only because it said so in the ticket.  Should I remove from those other locations too?

speedofdark:
If that's what the ticket said, its probably for a reason. Was just curious.

PR #1691, opened 2019-04-01
(maint) Adds .idea to gitignore.

tavilarion:


melissa:
@tavilarion would you be able to add a little more detail to the commit message? What creates `.idea`? It would be nice to have more context around why we want to merge this in

tavilarion:
Closing this, will create a new merge request with more details about why this is required.

PR #1692, opened 2019-04-01
(PA-2567) Remove debian-7 x86_64/i386 platforms

gimmyxd:
Remove debian-7 x86_64/i386 platform definition from puppet-agent repository


mihaibuzgau:
👍

PR #1704, opened 2019-04-08
Adds .idea to gitignore.

tavilarion:
The .idea folder is created when using an IDE (eg. Ruby Mine) for local caching and settings. It needs to be added to .gitignore so that developers using an IDE do not accidentally commit it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ acceptance/log/*
 acceptance/tmp/*
 acceptance/merged_options.rb
 acceptance/.beaker/*
+.idea


melissa:
@tavilarion thanks for updating the commit message!

You can update a commit message without opening a new pull request, but you have to force push to your development branch. Be super careful with force pushing, and always be very specific about the destination of the force push.

I'm also not sure this is a change we want to merge in. This file seems specific to the environment you're developing in. We don't want the `.gitignore` checked into puppet to contain every possible aspect of an environment that users don't want to check in. Rather than adding this here, would a global gitignore specific to your development environment satisfy? You can find more infor on global gitignores here: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

PR #1708, opened 2019-04-25
(PA-2668) Add LD_LIBRARY_PATH to environment cleanup in AIX wrapper

abottchen:
Prior to this commit, we cleared LD_LIBRARY_PATH on Linux based
agents, but not AIX.  This meant that customers with that set
globally on their system were subject to library loading errors.

With this commit, we clear the LD_LIBRARY_PATH just as we do on
Linux agents.

### DIFF
--- a/resources/files/aix-wrapper.sh
+++ b/resources/files/aix-wrapper.sh
@@ -3,6 +3,7 @@
 unset LIBPATH
 unset LDR_PRELOAD
 unset LDR_PRELOAD64
+unset LD_LIBRARY_PATH
 
 COMMAND=`basename "${0}"`
 


abottchen:
Travis seemed to fail claiming I did not title the commit correctly, but I'm fairly sure I did.

caseywilliams:
Ah, sorry @abottchen it looks like this will need a rebase if we want it in the 5.5.x branch. I'll set it back to master for now for easier reviewing.

abottchen:
@caseywilliams Ok, I rebased it to 5.5.x.  Sorry for the confusion.  I wasn't clear on which branch to submit against.

PR #1711, opened 2019-04-30
(maint) do not ship OS X 1.10.x to the nightly repos

gimmyxd:
The 1.10.x release series is on extended support for a single customer.
This customer doesn't use OS X.
There is a PR to remove OS X from the pipeline: https://github.com/puppetlabs/ci-job-configs/pull/5781
This commit removes the OS X from the nightly ship because it will try to ship packages that were not built.


gimmyxd:
We should make sure that this changes won't be merged-up.

PR #1713, opened 2019-05-02
(maint) Upgrade Docker container Alpine base from 3.8 to 3.9

Iristyle:
- The Alpine based agent container was running atop 3.8, but can be
   updated to 3.9.

   Alpine 3.9 release notes:
   alpinelinux.org/posts/Alpine-3.9.0-released.html

### DIFF
--- a/docker/ci/build.ps1
+++ b/docker/ci/build.ps1
@@ -50,7 +50,7 @@ function Build-Container(
   }
   else # alpine
   {
-    $docker_args += '--memory 3g'
+    $docker_args += @('--memory', '3g')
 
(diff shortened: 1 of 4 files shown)


Iristyle:
Looks like this will have to be in a holding pattern for now until we can build everything we care about under GCC 8.3, which is what Alpine 3.9 includes.

At the very least we know RapidJSON needs some fixes - see https://github.com/Tencent/rapidjson/issues/1341 and https://github.com/Tencent/rapidjson/issues/1483

Iristyle:
Going to close for now since it will be a while before we can revisit

PR #1714, opened 2019-05-02
(maint) Add Alpine container bind-tools package

Iristyle:
- This will give access to `dig` for diagnosing networking problems
   and should only impact container size minimally


Iristyle:
Contains same commit for fixing the memory switch (currently not used in CI) as #1713 ... figured we could get it in there given I've noticed the problem. (A future PR will hopefully re-enable Alpine builds in Azure)

PR #1720, opened 2019-05-09
(PA-2670) Update puppet, facter, and hiera to include version in gemspec

underscorgan:
Between ruby 2.1 and 2.4, gem stopped being able to properly process
gemspecs if they did not match the pattern <gem>-<version>.gemspec. This
code updates the installation of the puppet, facter, and hiera gems to
include the version calculated from the git sources.

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -239,7 +239,7 @@ component "facter" do |pkg, settings, platform|
     end
   end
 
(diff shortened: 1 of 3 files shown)


underscorgan:
Do not merge until a release of vanagon happens including https://github.com/puppetlabs/vanagon/pull/604

pillarsdotnet:
@gimmyxd Will this fix also appear in puppet 6.x ?

branan:
@pillarsdotnet Yes, we have automation that automatically brings bugfixes up from older branches to new versions

PR #1727, opened 2019-05-24
(PA-2629) Add dependecy on `find` for puppet-agent

gimmyxd:
Find utility is used by the `module tool`. Since some
Fedora minimal distributions could come without `find`
we add this as an dependecy to the `puppet-runtime`

### DIFF
--- a/configs/components/puppet-runtime.rb
+++ b/configs/components/puppet-runtime.rb
@@ -40,6 +40,8 @@ component 'puppet-runtime' do |pkg, settings, platform|
     pkg.requires 'libyaml-cpp0_6'
   end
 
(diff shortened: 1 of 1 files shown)


configs/components/puppet-runtime.rb
```diff
@@ -40,6 +40,8 @@
     pkg.requires 'libyaml-cpp0_6'
   end
 
+  pkg.requires '/usr/bin/find' if platform =~ /^fedora/
```
bastelfreak: Hi. I don't think that this is Fedora specific. We just noticed it first on Fedora, but the dependency exists on every linux platform?

PR #1728, opened 2019-05-28
(PA-2692) Handle unset vars in /etc/profile.d/puppet.sh

pingram3030:
If 'set -u' is set in shell or bash then unset variables will exit non-zero. For scripts which need to use puppet and have this set we must ensure that puppet.sh can be sourced cleanly.

### DIFF
--- a/resources/files/puppet-agent.sh
+++ b/resources/files/puppet-agent.sh
@@ -1,9 +1,13 @@
 # Add /opt/puppetlabs/bin to the path for sh compatible users
 
-if ! echo $PATH | grep -q /opt/puppetlabs/bin ; then
-  export PATH=$PATH:/opt/puppetlabs/bin
+if [ -z ${PATH-} ] ; then
+  export PATH=/opt/puppetlabs/bin
+elif ! echo ${PATH} | grep -q /opt/puppetlabs/bin ; then
+  export PATH=${PATH}:/opt/puppetlabs/bin
 fi
 
-if ! echo $MANPATH | grep -q /opt/puppetlabs/puppet/share/man ; then
-  export MANPATH=$MANPATH:/opt/puppetlabs/puppet/share/man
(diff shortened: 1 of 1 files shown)


oanatmaria:
Hi Phil,
Thank you for your contribution. I saw that Affects Version/s in this ticket is puppet-agent 5.5.7 and you made your pull request on master, could you rebase your changes to 5.5.x please?

pingram3030:
I reset to `5.5.x`, but because this PR was created as a merge to master I think I need to close it and create one fresh. Should I close this PR and create on anew?

KervyN:
Hey,
can we do a rollback to the old version? Because THIS one overwrites the default manpath in a way, that only the puppet manuals are working.

Tested on centos 7 and ubuntu 18.04

GabrielNagy:
@KervyN, thanks for catching this. We're on it.

nmaludy:
Having the same problem, it's clobbering our default lookups (you can check with `man -w`) now the only path that's being used is the Puppet path.

To fix this i removed the MANPATH lines in `/etc/profiles.d/puppet-agent.sh` and instead added the following line to `/etc/man_db.conf`:
```
MANPATH_MAP	/opt/puppetlabs/bin		/opt/puppetlabs/puppet/share/man
```

This will automatically add to the `MANPATH` if `/opt/puppetlabs/bin` is in `PATH`.

pingram3030:
> Edit:
> FFS?!? Why did you do it the way you did?
> You explained here (https://tickets.puppetlabs.com/browse/PA-2692) that your problem will get solved if you puth a hyphen behind the variable.
> 
> Why don't you just wrote it like this?
@KervyN

I guess because I didn't know and I didn't face the same issue as yourself. Thank you for being so cordial, understanding, and constructive with your replies. Es tut mir leid und Mea culpa.

PR #1736, opened 2019-06-11
(maint) Use agent-runtime-6.4.x instead of master

gimmyxd:
PR that introduce 6.4.x runtime: https://github.com/puppetlabs/ci-job-configs/pull/5693


gimmyxd:
`Downloading file 'agent-runtime-6.4.x-201906061.fedora-30-x86_64.tar.gz.sha1'`

PR #1742, opened 2019-06-18
(PA-2750) Ensure /opt/puppetlabs/bin is on the PATH

Sharpie:
Prior to this commit, Puppet subcommands that delegated to external
scripts or binaries, such as `puppet-query` or `puppet-infra`, would
show up as "not found" if `/opt/pupptlabs/bin` was not on the `PATH`.

This could commonly happen:

  - Immediately after installation of Puppet or PE as the shell needs
    to be re-started or re-freshed to pick up the effect of
    `/etc/profile.d/puppet-agent.sh`

  - When running `sudo puppet` in an environment where `/etc/sudoers`
    has the `env_reset` option enabled (it is enabled by default).

This commit updates the wrapper scripts to ensure `/opt/puppetlabs/bin`
is on the `PATH` in addition to unsetting various environment variables
that affect where the linker looks for shared libraries.

### DIFF
--- a/resources/files/aix-wrapper.sh
+++ b/resources/files/aix-wrapper.sh
@@ -5,6 +5,12 @@ unset LDR_PRELOAD
 unset LDR_PRELOAD64
 unset LD_LIBRARY_PATH
 
+# If $PATH does not match a regex for /opt/puppetlabs/bin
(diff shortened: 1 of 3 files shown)


Sharpie:
This works with Bourne shell on AIX 6.1 and Solaris 10u8 in addition to modern Linux varieties.

PR #1749, opened 2019-06-25
(PA-2701) Fix DHCP restart on puppet-agent upgrade

Dorin-Pleava:
Changed the Windows Registry: Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\EventLog\Application\Puppet\EventMessageFile from [INSTALLDIR]puppet\bin\puppetres.dll to %SystemRoot%\System32\EventCreate.exe

This will create custom events and will not use puppetres.dll, as puppet/lib/puppet/util/windows/eventlog.rb does the same thing regarding events.

Now the EventCreate.exe will get locked and loaded into svchost.exe and will not
interfere with puppet-agent upgrade.

Note this fix will fix future puppet-agent versions, but to install the
fixed version when you have the bug, you have to use a workaround and
install puppet-agent with the msi_move_locked_files flag set to true.

### DIFF
--- a/resources/windows/wix/filter.xslt.erb
+++ b/resources/windows/wix/filter.xslt.erb
@@ -30,14 +30,4 @@
   <!-- Above directly copies Vanagons default filter.xslt.erb as Vanagon does not provide a
        way to perform additional transforms against heat harvested file list -->
 
-  <!-- (PA-768) Modify puppetres.dll component as permanent -->
-  <xsl:template match="wix:Component[wix:File[@Source='$(var.AppSourcePath)\puppet\bin\puppetres.dll']]">
-    <!-- Copy original component with all attributes, marking "Permanent" -->
-    <xsl:copy>
-      <xsl:apply-templates select="@*" />
-      <xsl:attribute name="Permanent">yes</xsl:attribute>
-      <xsl:apply-templates select="node()" />
-    </xsl:copy>
-  </xsl:template>
-
 </xsl:stylesheet>
--- a/resources/windows/wix/registryEntries.wxs.erb
+++ b/resources/windows/wix/registryEntries.wxs.erb
@@ -62,7 +62,7 @@
           <RegistryValue
             Type="string"
             Name="EventMessageFile"
-            Value="[INSTALLDIR]puppet\bin\puppetres.dll"/>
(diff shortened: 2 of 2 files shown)


Dorin-Pleava:
Jenkins please test this on windows2016-64

Iristyle:
Hey @Dorin-Pleava - thanks for the PR!

I've never actually seen `eventcreate` used as the `EventMessageFile` like this. Is there some documentation you might be able to point out? I've only seen `eventcreate.exe` used at the command line to create events - as shown in https://ss64.com/nt/eventcreate.html

It would also be useful to have some screenshots of how events appear now in the Windows event log viewer with this change.

Thanks!

glennsarti:
For ref here is the documentation on those registry keys

https://docs.microsoft.com/en-us/windows/desktop/eventlog/event-sources (See EventMessageFile)

Description of message files

https://docs.microsoft.com/en-us/windows/desktop/eventlog/message-files

Dorin-Pleava:
Thanks @glennsarti for the refs.
@Iristyle When puppetres.dll is no longer loaded as a EventMessageFile, I tried to find the simplest way to display the events in Windows Event Log, as all puppet events were Description for event id xx from source cannot be found. Came across this: 
https://stackoverflow.com/questions/3412463/description-for-event-id-from-source-cannot-be-found/46015457#46015457  and if puppet does not have more that 1000 events ID and the error message is the EventData I think it should work, but I may be missing something.

Dorin-Pleava:
Capture

Dorin-Pleava:
Capture2

jcoconnor:
Removing the Don't Merge as this was discussed at the SOW on Tue 23/July.
/cc: @Iristyle

Dorin-Pleava:
@jcoconnor Thanks John, I created a PR [here](https://github.com/puppetlabs/puppet/pull/7640) with some fixes, and writing down the findings.

Dorin-Pleava:
To be merged together with [PUP-9928](https://github.com/puppetlabs/puppet/pull/7640)

jcoconnor:
Hi @Iristyle Can you give a final check & 👍 tick please so I can hit merge on it.
Thanks

PR #1756, opened 2019-07-08
(PA-2768) Extend manpath instead of overriding it

GabrielNagy:
Whoever implemented `MANPATH` apparently didn't do it like `PATH`.

By default, `MANPATH` is not set and man relies on paths specified in `/etc/man_db.conf`. When setting `MANPATH` to a single path, the default paths are overridden with this variable.

To make it work we need to prepend a colon to the exported variable, so that the search path gets extended instead of being replaced.

This is a regression introduced by https://github.com/puppetlabs/puppet-agent/pull/1728.
Thanks @KervyN for noticing and supplying a working version.

### DIFF
--- a/resources/files/puppet-agent.sh
+++ b/resources/files/puppet-agent.sh
@@ -6,8 +6,6 @@ elif ! echo ${PATH} | grep -q /opt/puppetlabs/bin ; then
   export PATH=${PATH}:/opt/puppetlabs/bin
 fi
 
-if [ -z ${MANPATH-} ] ; then
-  export MANPATH=/opt/puppetlabs/puppet/share/man
-elif ! echo ${MANPATH} | grep -q /opt/puppetlabs/puppet/share/man ; then
-  export MANPATH=${MANPATH}:/opt/puppetlabs/puppet/share/man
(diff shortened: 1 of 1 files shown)


mphilipps:
Hi,
I just stumbled over this bug after finding all man pages missing. So thank you for fixing this.  

However after having a closer look at /opt/puppetlabs/puppet/share/man I have to ask why is puppet even providing all  these manpages? Why do I need another version of the openssl manpage? 
It seems like all the provides man pages are either for outdated libraries or of poor quality. Why would a manpage ever start with "This page is autogenerated; any changes will get overwritten"? How likely is it for a user mistakenly edit a man page?

I feel like all manpages except the ones starting with puppet- in section 8 should be removed, they are more likely to cause harm than help.

GabrielNagy:
Hi @mphilipps!

Looks like man pages are being provided for everything that is bundled with puppet-agent (curl/ruby/etc).

However this shouldn't be the problem with the current `MANPAGE` implementation, as we are appending the `/opt/puppetlabs` path, so it will always be searched at the end.

So if your system ruby is 2.6 and puppet ruby is 2.4, you'd get the ruby 2.6 manpage. If you don't have ruby installed, you would get the manpage for the agent-bundled ruby.

I'd argue this doesn't do any harm.

PR #1775, opened 2019-08-06
(maint) Use shared helpers in Azure

Iristyle:


Iristyle:
We still have to move over to the standard `Build-Container` helper to complete this work, but saving that for a separate PR since it will be a little bit of work.

azure-pipelines.yml
```diff
@@ -8,108 +8,100 @@ pr:
...
-  displayName: Test Puppet Agent Ubuntu
-  name: test_puppet_agent_ubuntu
-- task: PublishTestResults@2
-  displayName: Publish Puppet Agent Ubuntu test results
-  inputs:
-    testResultsFormat: 'JUnit'
-    testResultsFiles: 'docker/TEST-*.xml'
-    #mergeTestResults: false # Optional
-    testRunTitle: Puppet Agent Ubuntu Test Results
+    Build-Container -Namespace $ENV:NAMESPACE -Base 'ubuntu'
```
nwolfe: I don't recognize this `-Base` option ?

nwolfe: Oh it's not coming from the shared one, it's coming from the local one

Iristyle: That's a leftover from the existing `Build-Container` helper that I left alone for now (rather than from the shared one)

PR #1776, opened 2019-08-14
(maint) Clear Azure source directory on builds

Iristyle:
- Occasionally files intended to live in containers get written on
   Windows with the wrong line endings. Linux containers need LF line
   endings for executable scripts and config files (for instance), but
   the default checkout on Windows is CRLF, unless explicitly configured
   to be different.

   Unfortunately a subsequent pull of the code will not reset local
   line endings on Windows, even when .gitattributes is set properly.

   This typically results in RDPing into an Azure DevOps builder and
   manually trashing the source code, so that when the code is next
   checked out, the appropriate LF line endings will be set.

   Instead of doing that, have the `s` (source) subdirectory purged
   on each build.

   https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#job

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,9 @@ variables:
   CONTAINER_NAME2: puppet-agent-alpine
   LINT_IGNORES: DL3008 DL3018 DL4000 DL4001 DL3028
 
+workspace:
+  clean: resources
+
 steps:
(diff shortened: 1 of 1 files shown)


Iristyle:
Azure passed, but the check never came back to GH for some reason... hmmph.

PR #1779, opened 2019-08-20
(maint) Save smoke test params as json

mwaggett:
This commit updates the repo smoke test script to send parameters to Artifactory
as json instead of yaml. This allows us to make use of bash tools, like jq, that
already exist on the Jenkins workers that will consume these params.

### DIFF
--- a/ext/smoke/packages/run-smoke-test.sh
+++ b/ext/smoke/packages/run-smoke-test.sh
@@ -31,9 +31,11 @@ echo "#### Running validation"
 $(dirname $0)/../steps/run-validation-tests.sh ${master_vm} ${agent_vm}
 
 echo "#### Sending parameters to Artifactory for use by repo smoke test"
-versions_file=${agent_version}.yaml
-echo "puppetserver: ${server_version}" > ${versions_file}
-echo "puppetdb: ${puppetdb_version}" >> ${versions_file}
+versions_file=${agent_version}.json
+echo "{" > ${versions_file}
+echo "\"puppetserver\":\"${server_version}\"," >> ${versions_file}
+echo "\"puppetdb\":\"${puppetdb_version}\"" >> ${versions_file}
+echo "}" >> ${versions_file}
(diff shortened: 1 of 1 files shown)


ext/smoke/packages/run-smoke-test.sh
```diff
@@ -31,9 +31,11 @@ echo "#### Running validation"
 $(dirname $0)/../steps/run-validation-tests.sh ${master_vm} ${agent_vm}
 
 echo "#### Sending parameters to Artifactory for use by repo smoke test"
-versions_file=${agent_version}.yaml
-echo "puppetserver: ${server_version}" > ${versions_file}
-echo "puppetdb: ${puppetdb_version}" >> ${versions_file}
+versions_file=${agent_version}.json
+echo "{" > ${versions_file}
```
e-gris: I would suggest this as being easier to read:
```
echo "{
\"puppetserver\":\"$server_version\",
\"puppetdb\":\"$puppetdb_version\"
}" > $versions_file
```

ext/smoke/packages/run-smoke-test.sh
```diff
@@ -31,11 +31,14 @@ echo "#### Running validation"
...
-echo "puppetdb: ${puppetdb_version}" >> ${versions_file}
-curl -T ${versions_file} "https://artifactory.delivery.puppetlabs.net/artifactory/scratchpad__local/puppet-agent-version-compatibility/${versions_file}"
-rm ${versions_file}
+versions_file=${agent_version}.json
+version_data="{
+                \"puppetserver\":\"$server_version\",
+                \"puppetdb\":\"$puppetdb_version\"
+              }"
+echo $version_data > $versions_file
+curl -T $versions_file "https://artifactory.delivery.puppetlabs.net/artifactory/scratchpad__local/puppet-agent-version-compatibility/${versions_file}"
```
shrug: What happens if this fails?

mwaggett: If the curl itself fails, then I guess this script will just fail and the data won't be in Artifactory.
I don't think anything disastrous would happen? We'd just end up needing to re-run the repo smoke test job..

PR #1787, opened 2019-09-06
(PA-2840) Enable winfips-2012r2-x64 under PE

gimmyxd:


ext/build_defaults.yaml
```diff
@@ -35,6 +35,7 @@ foss_platforms:
   - ubuntu-18.04-amd64
   - windows-2012-x86
   - windows-2012-x64
+  - windowsfips-2012-x64
```
gimmyxd: @mwaggett / @underscorgan I'm not sure about the naming conventions in this file and how is used(should it be the same name as in `config/platforms/windows-2012r2fips-x64`?)
Also i think we will not ship windows fips packages under http://nightlies.puppet.com/downloads/ right?

mwaggett: @gimmyxd generally yes, these should use the same name as in the vanagon config and in packaging. But yeah, if this is a PE-only platform, it should *not* get added to this list of foss_platforms.

mwaggett: you'll want to add it here instead: https://github.com/puppetlabs/puppet-agent/blob/master/ext/build_defaults.yaml#L37

PR #1802, opened 2019-10-01
(PA-2840) fix windowsfips build_details definition

gimmyxd:


gimmyxd:
Have the same naming as: https://github.com/gimmyxd/packaging/blob/4d0b51d60c34a6ca04396994d37e97a1e1db06d3/lib/packaging/platforms.rb#L343 and https://github.com/gimmyxd/beaker-hostgenerator/blob/4843c64c18a40ddd12cc4037c6a3e373c35cfa8c/lib/beaker-hostgenerator/data.rb#L1438

PR #1809, opened 2019-10-17
(maint) Remove solarisstudio

mihaibuzgau:
This PR removes solarisstudio publisher from Solaris 11 VMs. This publisher is not used in the agent pipeline


joshcooper:
I think the solarisstudio publisher maybe required in the zone/zfs/zpool beaker tests, which are only tested in branch 5.5.x of the puppet-agent pipeline. IIRC the tests end up installing a "small-server". But that wouldn't be necessary during packaging time. So 👍

PR #1836, opened 2019-12-09
(maint) Improve Dockerfile organization

Iristyle:
On Travis, this goes from 12 and 81 steps to 11 and 44 steps respectively - though there didn't seem to be any appreciable time savings in the builds. I'll let you have a look at e9f9f45 specifically for feedback @underscorgan as I've broken apart the grouping of component operations (i.e. facter, pxp-agent) into operations within workspaces - i.e. clone all source for all components, then build each component independently.

We lose 20+ lines of Dockerfile code in the process, but whether this improves or worsens maintenance is probably debatable.


docker/puppet-agent-alpine/Dockerfile
```diff
@@ -9,97 +9,85 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
...
+    mkdir -p /workspace/cpp-pcp-client/build && \
+    git clone https://github.com/puppetlabs/facter && \
+    mkdir -p /workspace/facter/build && \
+    git clone https://github.com/puppetlabs/pxp-agent && \
+    mkdir -p /workspace/pxp-agent/build && \
+    curl -O -L https://people.redhat.com/~rjones/virt-what/files/virt-what-1.18.tar.gz && \
+    tar zxf virt-what-1.18.tar.gz && \
+    git clone https://github.com/puppetlabs/hiera && \
+    git clone https://github.com/puppetlabs/puppet
+
```
underscorgan: so, these were all split up by project before to be able to make better use of reusing layers so that you don't have to rebuild everything if just, say, puppet changes

Iristyle: If I'm not mistaken, any `RUN` command immediately invalidates the Docker build cache from that point forward... so I don't think Docker is smart enough to understand what you were aiming for.

But I could be mistaken - so let me re-verify my understanding since this is one of the more complex containers given its multi-stage.

Iristyle: Ok, for existing build:

* first stage is able to cache 7/12 steps
* second stage is able to cache 71/81 steps 

For updated:

Iristyle:
superseded by #1898

PR #1889, opened 2020-04-16
(maint) Remove LCOW support / convert to WSL2

Iristyle:
- Remove special Windows build process that was needed for LCOW and use
   the same build files as on Linux
 - Turns on Docker Buildkit to speed things up (this is especially noticeable on Windows where steps are cached, but might require some more Dockerfile reorganization to really take advantage of)


Iristyle:
I'll plan on doing a follow-on PR to restructure the Dockerfiles in an effort to clean them up / speed them up.

PR #1892, opened 2020-04-17
(maint) Enable linux docker buildkit

Iristyle:
Turns on Buildkit to improve Docker build performance

Builds on #1889 

This turns out to be trickier than I originally anticipated, so I'm moving this effort to a separate PR.

From what I can tell, removing `gcr.io` from `daemon.json` and restarting Docker doesn't resolve the problem (as described in https://github.com/moby/buildkit/issues/606#issuecomment-453959632)

I solved this once before in https://github.com/puppetlabs/docker-pe-postgres/pull/18 ... there is a 30s penalty to upgrading Docker.

Another solution in Travis is to use the container that runs buildkit - i.e. https://travis-ci.community/t/docker-builds-are-broken-if-buildkit-is-used-docker-buildkit-1/2994/4

Some more ideas are here -- https://blog.alexellis.io/building-containers-without-docker/

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -8,6 +8,7 @@ hadolint_container := hadolint/hadolint:latest
 export BUNDLE_PATH = $(PWD)/.bundle/gems
 export BUNDLE_BIN = $(PWD)/.bundle/bin
 export GEMFILE = $(PWD)/Gemfile
+export DOCKER_BUILDKIT = 1
 
 version = $(shell echo $(git_describe) | sed 's/-.*//')
 dockerfile := Dockerfile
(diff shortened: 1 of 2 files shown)


docker/Makefile
```diff
@@ -8,6 +8,7 @@ hadolint_container := hadolint/hadolint:latest
 export BUNDLE_PATH = $(PWD)/.bundle/gems
 export BUNDLE_BIN = $(PWD)/.bundle/bin
 export GEMFILE = $(PWD)/Gemfile
+export DOCKER_BUILDKIT = 1
```
nwolfe, line 11: Was this ENV supposed to be referenced by your other changes on lines 34 and 36?

Iristyle: Nope, it's a global setting

PR #1898, opened 2020-04-29
(maint) Overhaul Docker Alpine build process for efficiency and reuse  / upgrade Alpine 3.8 -> 3.9

Iristyle:
- Upgrade docker-compose to 1.25.5
- Inline Hadolint exclusions rather than setting them globally
- Move unnecessary ENV vars to ARGs
- Remove wget dependency from Ubuntu
- Fix Alpine container labeling bug
- Don't build native gems in Alpine container - install upstream apk packages (json, etc)
- Rework how json files are parsed during Alpine build, doing away with having to set SHELL or pipefail
- Install virt-what package in Alpine instead of building it
- Restructure Alpine Dockerfile to take advantage of concurrent build
   stages with buildkit, by building out a dependency hierarchy, such
   that the following stages now exist:

   * build - created with all compile tooling + deps
   * leatherman - all Puppet C++ code depends on leatherman

   3 stages that can proceed concurrently:

   * libwhereami, depends only on leatherman
   * cpp-hocon, depends only on leatherman
   * pxp-agent, depends on leatherman + cpp-hocon, builds cpp-pcp-client

   2 subsequent stages:

   * facter, depends on cpp-hocon and libwhereami - will typically build
             concurrently with pxp-agent since facter builds longest
   * puppet, depends on facter and pxp-agent

   1 final stage:

   * creates the runtime container with minimal packages installed and
     results of output from prior build stages

   As structured, this allows for the maximum amount of concurrent work
   with stages, blocking them only when they need to rely on a prior
   stage as a parent or to copy files.
- Switch to Alpine 3.9 from 3.8
- Simplify how WORKDIR is used in build stages, shortening number of total instructions in the Dockerfile


mihaibuzgau:
It would be great to get this merged.
There's a lot of failures trying to pull alpine:3.8 from a travis image.

docker/puppet-agent-alpine/Dockerfile
```diff
@@ -2,15 +2,18 @@
...
 # note: JAVA_HOME cannot be defined in the same ENV block it's used in
 ENV JAVA_HOME=/usr/lib/jvm/java-1.8-openjdk
 ENV PATH=$PATH:$JAVA_HOME/jre/bin:$JAVA_HOME/bin \
     CMAKE_SHARED_OPTIONS='-DCMAKE_PREFIX_PATH=/opt/puppetlabs/puppet -DCMAKE_INSTALL_PREFIX=/opt/puppetlabs/puppet -DCMAKE_INSTALL_RPATH=/opt/puppetlabs/puppet/lib -DCMAKE_VERBOSE_MAKEFILE=ON'
 
+# pin back to boost, gcc and g++ packages from alpine 3.8
+# 3.9+ packages g++6 and gcc6 are difficult to install, even when using --force-broken-world
 # hadolint ignore=DL3018,DL3028
-RUN apk add --no-cache cmake boost-dev make curl git curl-dev ruby ruby-dev gcc g++ yaml-cpp-dev jq openjdk8 augeas ruby-augeas ruby-etc ruby-json ruby-multi_json libressl-dev virt-what && \
+RUN apk add --no-cache --update-cache --repository http://nl.alpinelinux.org/alpine/v3.8/main cmake boost-dev make gcc=6.4.0-r9 g++=6.4.0-r9  && \
```
Iristyle, line 15: Oops a bad rebase unintentionally picked up `cmake`, `boost-dev` and `make` from 3.8. Will fix those unnecessary package installs momentarily.

PR #1906, opened 2020-05-11
(maint) pull docker images before build

mihaibuzgau:
This PR ensures that the base docker images are available before building the Puppet Agent containers.

On travis `--pull` build argument seems to be causing the following transient error:
```
#3 [internal] load metadata for docker.io/library/alpine:3.9
#3 ERROR: docker.io/library/alpine:3.9 not found
#9 [internal] load build context
#9 transferring context: 10.45kB done
#9 DONE 0.0s
#4 [build 1/3] FROM docker.io/library/alpine:3.9
#4 resolve docker.io/library/alpine:3.9 0.0s done
#4 ERROR: docker.io/library/alpine:3.9 not found
```

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -5,6 +5,8 @@ build_date := $(shell date -u +%FT%T)
 hadolint_available := $(shell hadolint --help > /dev/null 2>&1; echo $$?)
 hadolint_command := hadolint
 hadolint_container := hadolint/hadolint:latest
+alpine_version := 3.9
+ubuntu_version := 18.04
 export BUNDLE_PATH = $(PWD)/.bundle/gems
 export BUNDLE_BIN = $(PWD)/.bundle/bin
 export GEMFILE = $(PWD)/Gemfile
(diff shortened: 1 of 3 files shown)


Iristyle:
Very interesting! I wonder if there's a bug in `--pull` with buildkit, because what you did is what it's supposed to be doing. I only saw Alpine 3.8, Alpine 3.10 and CentOS 7 fail, but I see the fail at https://travis-ci.org/github/puppetlabs/puppet-agent/jobs/685511139#L1622-L1624 that probably made you enact this change, so clearly 3.9 has had problems too.  I thought it had something to do with how Travis caches particular images for docker.io (I assumed they had a proxy / mirror setup b/c I've only ever seen this fail in Travis). 

Do you recall seeing any of these transient before the switch to buildkit? And will you let me know if you still see transients after this merge? I wasn't aware of this problem -- but if this pre-pull fixes it and its causing flakiness in other pipelines, I'll need to put up another 10 or so PRs to various repos.

Thanks!

PR #1912, opened 2020-06-18
(PA-3263) fix nssm lock - 5.5.x

ciprianbadescu:
Before this fix, if the puppet-agent upgrade on Windows was performed
while nssm.exe was loaded by EventLog service, many services, including
critical ones lihe dhcp/dns clients were restarted, leading to unreachable
machines.

With this commit we deliver dedicated nssm executable for pxp-agent
(nssm-pxp-agent.exe) and we remove windows registry references to old
nssm.exe in a WiX custom action.

### DIFF
--- a/configs/components/nssm.rb
+++ b/configs/components/nssm.rb
@@ -11,5 +11,5 @@ component "nssm" do |pkg, settings, platform|
     ]
   end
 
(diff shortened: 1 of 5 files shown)


ciprianbadescu:
Binary packages are available at: http://builds.delivery.puppetlabs.net/puppet-agent/f409fac470b0502e87dd9c8af92ffd1927d43cfd/artifacts/windows/

PR #1913, opened 2020-06-18
(PA-3263) fix nssm lock - master

ciprianbadescu:
Before this fix, if the puppet-agent upgrade on Windows was performed
while nssm.exe was loaded by EventLog service, many services, including
critical ones lihe dhcp/dns clients were restarted, leading to unreachable
machines.

With this commit we deliver dedicated nssm executable for pxp-agent
(nssm-pxp-agent.exe) and we remove windows registry references to old
nssm.exe in a WiX custom action.

### DIFF
--- a/configs/components/nssm.rb
+++ b/configs/components/nssm.rb
@@ -11,5 +11,5 @@ component "nssm" do |pkg, settings, platform|
     ]
   end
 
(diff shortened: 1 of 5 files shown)


ciprianbadescu:
Testing packages available at: http://builds.delivery.puppetlabs.net/puppet-agent/0ddeaa33be0514face5b52b730f5d969a9598928/artifacts/windows/

PR #1916, opened 2020-06-22
(maint) bump puppetlabs-sshkey_core module

ciprianbadescu:


gimmyxd:
@ciprianbadescu @GabrielNagy If 2.x is not backwards compatible win 1.x i don't think we want to do this change in the 6 stream.

ciprianbadescu:
probably that's why bump to 2.0.0 was also skipped

PR #1925, opened 2020-07-13
(PA-3280) update tests

ciprianbadescu:
- use `identify_master` only on repo smoke
 - update reading of `route_file` config setting


ciprianbadescu:
tests will fail until 6.17.0 will be released

PR #1957, opened 2020-09-03
(maint) do not provision boost on osx (5.5.x)

ciprianbadescu:
osx provisioning fails as boost 1.60 was disabled on homebrew (https://github.com/Homebrew/homebrew-core/commit/d18d380bbbf90808088460cd244c828171263196)


ciprianbadescu:
still checking if we need to install homebrew at all

gimmyxd (approved):

PR #1973, opened 2020-09-18
(PA-3404) remove %PUPPET_DIR%\bin from PATH

gimmyxd:
Before this commit, on Windows, puppet bin was
prepended to PATH, this was making puppet commands
to always use the ruby and gem provided in the
puppet-agent AIO.

With this change, puppet bin is appended at the
end of PATH, in the same manner we do for linux.

### DIFF
--- a/resources/files/windows/environment.bat
+++ b/resources/files/windows/environment.bat
@@ -9,8 +9,8 @@ SET PUPPET_DIR=%PL_BASEDIR%\puppet
 REM Set a fact so we can easily source the environment.bat file in the future.
 SET FACTER_env_windows_installdir=%PL_BASEDIR%
 
-REM Add puppet's bindirs to the PATH
-SET PATH=%PUPPET_DIR%\bin;%PL_BASEDIR%\bin;%PATH%
+REM Add puppet's bin to the PATH
+SET PATH=%PATH%;%PL_BASEDIR%\bin
 
 REM Set the RUBY LOAD_PATH using the RUBYLIB environment variable
 SET RUBYLIB=%PUPPET_DIR%\lib;%RUBYLIB%
--- a/resources/files/windows/puppet_shell.bat
+++ b/resources/files/windows/puppet_shell.bat
@@ -2,5 +2,8 @@
 
 call "%~dp0environment.bat" %0 %*
 
+REM Prepend puppet's bindirs to the PATH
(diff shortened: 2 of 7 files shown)


gimmyxd:
ad-hoc build - 867
image

resources/files/windows/environment.bat
```diff
@@ -10,7 +10,7 @@ REM Set a fact so we can easily source the environment.bat file in the future.
 SET FACTER_env_windows_installdir=%PL_BASEDIR%
 
 REM Add puppet's bindirs to the PATH
-SET PATH=%PUPPET_DIR%\bin;%PL_BASEDIR%\bin;%PATH%
+SET PATH=%PATH%;%PL_BASEDIR%\bin;
```
GabrielNagy, line 13: last `;` looks redundant I think

resources/files/windows/run_facter_interactive.bat
```diff
@@ -2,4 +2,4 @@
 @echo off
 SETLOCAL
 call "%~dp0environment.bat" %0 %*
-@elevate.exe "%~dp0facter_interactive.bat"
+"%PUPPET_DIR%\bin\elevate.exe" "%~dp0facter_interactive.bat"
```
GabrielNagy, line 5: We should keep the `@` I think, see: https://stackoverflow.com/a/21077142/6648276

same for `run_puppet_interactive.bat`

resources/files/windows/environment.bat
```diff
@@ -10,7 +10,7 @@ REM Set a fact so we can easily source the environment.bat file in the future.
 SET FACTER_env_windows_installdir=%PL_BASEDIR%
 
 REM Add puppet's bindirs to the PATH
```
GabrielNagy, line 12: nit: `bindirs` has now become `bindir`

resources/files/windows/puppet_shell.bat
```diff
@@ -3,4 +3,4 @@
 call "%~dp0environment.bat" %0 %*
```
gimmyxd, line 4: This is used to start a new CMD with puppet loaded, should we add `%PUPPET_DIR%\bin\` in PATH to be able to use the executables from there? Not really sure how people are using this.

gimmyxd:
passing ad-hoc build:869
image

PR #1977, opened 2020-09-23
(maint) Facter-ng version bump

oanatmaria:

PR #1980, opened 2020-09-29
(maint) Fixed Facter 3 and 4 comparisson tool

sebastian-miclea:
There was a small issue with the comparison class, in some cases it added converted the value to string even if it was not needed. A PR on Facter that fixed the output was merged earlier caused failing builds on agent.

### DIFF
--- a/acceptance/lib/puppet/acceptance/fact_dif.rb
+++ b/acceptance/lib/puppet/acceptance/fact_dif.rb
@@ -28,7 +28,7 @@ class FactDif
         path.pop
       end
     else
-      compare(path, sh.to_s)
+      compare(path, sh)
     end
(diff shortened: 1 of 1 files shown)


sebastian-miclea:
Jenkins please test this on all

joshcooper:
puppet-agent#master pipelines are failing for I think a different issue than this, but for visibility I filed https://tickets.puppetlabs.com/browse/PA-3407

sebastian-miclea:
@joshcooper yes this is not related to the master pipeline. I'll close this because it only happens on Facter 4.0.39 https://github.com/puppetlabs/puppet-agent/pull/1977 and will ask Oana add the change there.

PR #1983, opened 2020-10-01
(maint) Work correctly also if PATH contains spaces

fedepell:
Not maybe a very common case in general, but for example easy to spot when in WSL.

### DIFF
--- a/resources/files/puppet-agent.sh
+++ b/resources/files/puppet-agent.sh
@@ -1,11 +1,11 @@
 # Add /opt/puppetlabs/bin to the path for sh compatible users
 
-if [ -z ${PATH-} ] ; then
+if [ -z "${PATH-}" ] ; then
   export PATH=/opt/puppetlabs/bin
-elif ! echo ${PATH} | grep -q /opt/puppetlabs/bin ; then
-  export PATH=${PATH}:/opt/puppetlabs/bin
+elif ! echo "${PATH}" | grep -q /opt/puppetlabs/bin ; then
+  export PATH="${PATH}:/opt/puppetlabs/bin"
 fi
 
-if ! echo ${MANPATH-} | grep -q /opt/puppetlabs/puppet/share/man ; then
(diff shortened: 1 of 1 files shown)


GabrielNagy (commented):
Hi @fedepell,

Thanks for the contribution 🎉, I had some comments:
1. Travis is failing due to the commit check (you should also prefix your commit title with `(maint)`)
2. Lines 6 and 10 still run `export` with unquoted strings which will also get weird if `PATH` or `MANPATH` contain spaces. We should also wrap them in quotes.

Let me know if you have any questions.

fedepell:
Thanks for the feedback!
1. Ok fixed that! Sorry I'm very new to puppetlabs repositories ;)
2. Ok! I think that should have worked anyway (at least from the tests I did) while the one in _if_ really gives a bash syntax error. But for safety (and coherence in the script) indeed better to make them all quoted!

Hope it's better now! I squashed both so and also changed the title.

resources/files/puppet-agent.sh
```diff
@@ -1,11 +1,11 @@
 # Add /opt/puppetlabs/bin to the path for sh compatible users
 
-if [ -z ${PATH-} ] ; then
+if [ -z "${PATH-}" ] ; then
   export PATH=/opt/puppetlabs/bin
-elif ! echo ${PATH} | grep -q /opt/puppetlabs/bin ; then
-  export PATH=${PATH}:/opt/puppetlabs/bin
+elif ! echo "${PATH}" | grep -q /opt/puppetlabs/bin ; then
+  export PATH="${PATH}":/opt/puppetlabs/bin
```
GabrielNagy, line 6: I'd just enclose the entire value in quotes, like: `PATH="${PATH}:/opt/puppetlabs/bin"`, same for `MANPATH`. Works the same but in my opinion it reads better this way.

fedepell: Indeed looks better as you proposed! Fixed!

GabrielNagy (approved):

PR #1989, opened 2020-10-08
(PA-3401) Provide MSI property to configure serverport setting

ciprianbadescu:
Add `PUPPET_SERVER_PORT` installer property to be used to
configure puppet `serverport` setting


Dorin-Pleava (approved):
Tested on windows 2019 and serverport was set correctly when installing with msiexec

PR #2000, opened 2020-11-02
(maint) ensure facter/facts.d directory is created

mihaibuzgau:
In Agent 6 we create the `install_root/facter/facts.d` directory during installation time.
This PR enables this on Puppet Agent 7

### DIFF
--- a/configs/components/facter.rb
+++ b/configs/components/facter.rb
@@ -23,4 +23,9 @@ component "facter" do |pkg, settings, platform|
 
   pkg.install_file "facter.gemspec", "#{settings[:gem_home]}/specifications/#{pkg.get_name}-#{pkg.get_version_forced}.gemspec" unless platform.is_windows?
 
+  if platform.is_windows?
+    pkg.directory File.join(settings[:sysconfdir], 'facter', 'facts.d')
+  else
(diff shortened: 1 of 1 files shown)


GabrielNagy (approved):

GabrielNagy:
Seems that we're testing if `facts.d` under `cache` exists, but not under `/etc`. Maybe we can update the test in a different PR (so we don't block this).

https://github.com/puppetlabs/puppet-agent/blob/main/acceptance/tests/ensure_puppet-agent_paths.rb#L71

mihaibuzgau:
> Seems that we're testing if `facts.d` under `cache` exists, but not under `/etc`. Maybe we can update the test in a different PR (so we don't block this).
> 
> https://github.com/puppetlabs/puppet-agent/blob/main/acceptance/tests/ensure_puppet-agent_paths.rb#L71

agreed

mihaibuzgau:
With this fix:

mihaibuzgau:
With this PR:

joshcooper:
Might be good to review https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppet-agent-6-nix and verify that we're creating the same set of paths at installation (those marked with `*`).

PR #2012, opened 2020-11-12
(PA-3489) Add log rotation for pxp-agent on Solaris

Dorin-Pleava:
On Solaris platforms logrotate does not come by default, but logadm
comes.

Command logadm -C30 -z0 -w /var/log/puppetlabs/pxp-agent/pxp-agent.log
means:
-C30 keps 30 copies of logs
-z0 compress all archived logs
-w /var/log/puppetlabs/pxp-agent/pxp-agent.log adds an entry into the
config file (that is, /etc/logadm.conf) that corresponds to
/var/log/puppetlabs/pxp-agent/pxp-agent.log

DRAFT


Dorin-Pleava:
Closing this in favor of https://github.com/puppetlabs/puppet-enterprise-modules/pull/1199, as Solaris 11 does no longer support postinstall and postremove scripts.

PR #2016, opened 2020-11-18
(PA-3452) Update cert sign command for smoke script

GabrielNagy:
Puppet collections 6 and higher should use the `puppetserver` command to sign certs.


GabrielNagy:
CI fails as the 7.0.0 packages aren't yet publicly available.

mihaibuzgau (approved):

PR #2017, opened 2020-11-18
(PA-3491) Install configfile for OS X log rotation

GabrielNagy:
More info: https://github.com/puppetlabs/pxp-agent/pull/824


gimmyxd (approved):
We'll also have to update pxp-agent-vanagon and puppet-agent#main to handle configuration.

From what i remember, becasue pxp-agent is now a tar on main, we need provide the config files, and puppet-agent will put them in the right place.
eg:
 - https://github.com/puppetlabs/pxp-agent-vanagon/blob/main/configs/components/pxp-agent.rb#L104
https://github.com/puppetlabs/puppet-agent/blob/main/configs/components/pxp-agent.rb#L68

Dorin-Pleava (approved):

PR #2019, opened 2020-11-19
(maint) Promote aix 7.1 to pe#main

joshcooper:
We're now building on aix 7.1, so update the path, for example:

    repos/aix/7.1/puppet7/ppc/puppet-agent-7.0.0.51.g68a734e7-1.aix7.1.ppc.rpm

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -25,15 +25,14 @@ foss_platforms:
   - windows-2012-x86
   - windows-2012-x64
 pe_platforms:
-  - aix-6.1-power
   - aix-7.1-power
   - redhatfips-7-x86_64
   - solaris-11-i386
   - solaris-11-sparc
   - windowsfips-2012-x64
 platform_repos:
-  - name: aix-6.1-power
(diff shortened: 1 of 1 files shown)


ext/build_defaults.yaml
```diff
@@ -25,15 +25,14 @@ foss_platforms:
   - windows-2012-x86
   - windows-2012-x64
 pe_platforms:
-  - aix-6.1-power
   - aix-7.1-power
```
joshcooper, line 28: It's not clear to me why 7.2 is missing, but that seems like a different issue?

gimmyxd: Historically, AIX 7.1 and 7.2 were using AIX 6.1 packages. Now with the removal of AIX 6.1 both 7.1 and 7.2 will use 7.1 packages.

gimmyxd (approved):

gimmyxd (approved):

joshcooper:
superseded by #2020 which was merged up to main in #2026

PR #2021, opened 2020-11-20
(maint) Add repo_link_target for puppet7

GabrielNagy:
Now that we're publicly releasing puppet 7 we should probably add this to the main branch, and remove it from the 6.x stream.

This should have probably been done prior to shipping 7.0.0, but I don't fully know the implications of these params.

/cc @puppetlabs/release-engineering


underscorgan (commented):
These are no longer needed, link targets are set by a stand-alone task not each ship job

GabrielNagy:
Ah cool, thanks for the clarification!

PR #2022, opened 2020-11-20
(maint) Remove repo_link_target for puppet6

GabrielNagy:
See https://github.com/puppetlabs/puppet-agent/pull/2021.


underscorgan (approved):
These are no longer used by individual projects so this is good to remove

PR #2033, opened 2020-12-03
(maint) build el-6 on redhat instead of centos

mihaibuzgau:
Because CentOS 6 went end of life[1], the upstream repos are no longer available, the PR sets the build VMs to RedHat 6  instead of CentOS 6

1. https://phoenixnap.com/kb/centos-6-eol


ciprianbadescu (approved):

joshcooper:
Is puppet-runtime going to have the same problem: https://github.com/puppetlabs/puppet-runtime/blob/125697846327195e670a26f868fe60637578cc30/configs/platforms/el-6-x86_64.rb#L34?

PR #2034, opened 2020-12-03
(maint) dockerfile mkdir -p

sbeaulie:
without this if the directory already exists, the build will fail


nwolfe:
Similar to comment on r10k.... I don't know the answer but this could have unintended consequences. 

What was the original problem that led you to this solution?

docker/puppet-agent-alpine/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN apk add --no-cache --update-cache --repository http://nl.alpinelinux.org/alp
     apk add --no-cache cmake boost-dev make curl git curl-dev ruby ruby-dev yaml-cpp-dev jq openjdk8 augeas ruby-augeas ruby-etc ruby-json ruby-multi_json libressl-dev virt-what && \
     gem install --no-rdoc --no-ri deep_merge semantic_puppet puppet-resource_api locale httpclient fast_gettext concurrent-ruby thor hocon
 
-RUN mkdir /workspace && \
+RUN mkdir -p /workspace && \
```
Iristyle, line 20: Same thing about a `-p` for one level deep?

sbeaulie: same answer here https://github.com/puppetlabs/r10k/pull/1108#discussion_r545250666

Iristyle:
Closing since I don't believe we ever intend to use Kaniko

PR #2043, opened 2020-12-16
(maint) Exclude some legacy blockdevice facts

GabrielNagy:
Exclude the blockdevice size and vendor facts from the facter 3/4
acceptance tests, until we release a new Facter version containing
https://github.com/puppetlabs/facter/pull/2238.


mihaibuzgau:
This needs to be reverted once Facter 4.0.48 is released

mihaibuzgau (approved):

PR #2048, opened 2021-01-14
(PA-3533) Add support for el-8-ppc64le

GabrielNagy:


gimmyxd:
As 5.5.x is EOL we should add this platform starting with 6.x stream

PR #2064, opened 2021-02-10
(packaging) Bump to version '6.21.1' [no-promote]

ciprianbadescu:


GabrielNagy:
`lib/puppet/version.rb` should also be bumped

ciprianbadescu:
I think that is is puppet only

GabrielNagy:
too many repos

PR #2092, opened 2021-04-19
(maint) fix Dockerfile lint errors

ciprianbadescu:

PR #2094, opened 2021-04-22
(maint) Use Docker 19.03 in Travis again

Iristyle:
- Instead of incurring the upgrade cost to Travis 20.10, use the
   existing 19.03 version.

   In practice, upgrading Docker in Travis turned out to be unreliable.

   To enable buildx requires installing the plugin from the github
   releases page. Using the experimental cli flag was supposed to enable
   the plugin for 19.03, but didn't seem to work in the Travis env.

 - Also adds missing generation of named buildx builder to 6.x branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,20 +28,23 @@ matrix:
       rvm: 2.6.6
       env:
         - DOCKER_COMPOSE_VERSION=1.28.6
+        - DOCKER_BUILDX_VERSION=0.5.1
         # necessary to prevent overwhelming TravisCI build output limits
         - DOCKER_BUILD_FLAGS="--progress plain"
       before_install:
-        - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-        - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
-        - sudo apt-get update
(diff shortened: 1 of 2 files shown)


Iristyle:
Lint errors are addressed in https://github.com/puppetlabs/puppet-agent/pull/2092, which should be merged first

Iristyle:
Pulling in the @ciprianbadescu commit from #2092 as building containers requires the change in this PR... the Travis 20.10 upgrade is broken, but we also can't build without the lint fixes :)

So these commits are codependent / need to be in a single PR, but should probably be kept as separate commits.

Iristyle:
Hmmm... failing repeatedly with:

> E: Version '6.22.0-1bionic' for 'puppet-agent' was not found

GabrielNagy:
@Iristyle I think this happens because new releases were tagged but are not yet publicly shipped, and the Makefile runs `git describe` to determine the version. Also, we burned 6.22.0 and 7.6.0, and are doing 6.22.1 and 7.6.1.

However... running `git describe` on 6.x after pulling all tags still gives 6.22.0 instead of 6.22.1, so something's amiss 🧐.

@ciprianbadescu we might need to do something like this again https://github.com/puppetlabs/puppet-agent/pull/2063 ?

GabrielNagy (approved):

ciprianbadescu:
I rebased this to get new tag for puppet-agent in, not sure if commits list is correct

PR #2097, opened 2021-04-27
(PA-3701) Update acceptance tests to ensure matching between facter versions 

BobosilaVictor:
Reworked acceptance test to use puppet facts diff instead of its own implementation


BobosilaVictor:
jenkins please test this on all

BobosilaVictor:
jenkins please test this on all

BobosilaVictor:
jenkins please test this on all

acceptance/tests/ensure_facter_3_and_facter_4_output_matches.rb
```diff
@@ -1,57 +1,52 @@
...
+      on(agent, puppet("facts diff --exclude '#{ignored_facts}'")) do 
+        diff = JSON.parse(stdout)
+        unless diff.size.zero?
+          fail_test("Facter 3 and Facter 4 outputs have the following differences:  #{stdout}")
+        end
       end
     end
   end
-end
+end
```
gimmyxd, line 52: Rather then executing `puppet facts diff` and building a new exclude list, then execute `puppet facts diff` again,  I suggest to run `puppet facts diff` once, with a known exclude list and parse the output.
exclude_list - contains all the known differences, and should not make this test fail.

Something like:
```ruby
on agent, puppet("facts diff '#{exclude_list}'") do |output|
  diff = JSON.parse(output

  # rep_diff is the diff that we are interested in, and should make this test to fail
  # this ignores new facts added only in Facter 4
  rep_diff = diff.delete_if {|k, v| v['old_value'] == 'null' }
  fail_test if rep_diff.keys.size.positive?
end

File: acceptance/tests/ensure_facter_3_and_facter_4_output_matches.rb [hunk 2]
```diff
@@ -1,57 +1,52 @@
 test_name 'Ensure Facter 3 and Facter 4 outputs match' do
   require 'puppet/acceptance/common_utils'
-  require 'puppet/acceptance/fact_dif'
```
gimmyxd, line 3: `puppet/acceptance/fact_dif.rb` can be removed, this was the only place that we used it.

BobosilaVictor:
jenkins please test this on windows2012r2-64a,windows2019-64a,windows2016-64a,windows10ent-32a,windows10ent-64a,windowsfips2012r2-64a

ciprianbadescu:
jenkins please test this on all

BobosilaVictor:
jenkins please test this on all

BobosilaVictor:
jenkins please test this on all

PR #2102, opened 2021-05-14
(maint) update sed pattern for cross compiled

gimmyxd:
We changed the restrictions for Ruby version in
facter-ng.gemspec[1] which requires changes to the
sed pattern used on cross platforms.

Updated the pattern to delete the entire line if it
matches `spec.required_ruby_version`.

[1] puppetlabs/facter@2ba2c20

### DIFF
--- a/configs/components/facter-ng.rb
+++ b/configs/components/facter-ng.rb
@@ -22,7 +22,7 @@ component "facter-ng" do |pkg, settings, platform|
   # Remove required_ruby_version from gemspec, this is needed for cross-compiled
   # platforms as ruby version is lower than 2.3 on these platforms.
   if platform.is_cross_compiled?
-    sed_pattern = %(s/spec.required_ruby_version = '~> 2.3'/ /)
+    sed_pattern = %(/spec.required_ruby_version/d)
     pkg.build do
       [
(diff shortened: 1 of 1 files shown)


ciprianbadescu (approved):

gimmyxd:
Jenkins please test this on redhat7-POWERa

configs/components/facter-ng.rb
```diff
@@ -22,7 +22,7 @@
   # Remove required_ruby_version from gemspec, this is needed for cross-compiled
   # platforms as ruby version is lower than 2.3 on these platforms.
   if platform.is_cross_compiled?
-    sed_pattern = %(s/spec.required_ruby_version = '~> 2.3'/ /)
+    sed_pattern = %(s/spec.required_ruby_version = '>= 2.3', '< 4.0'/ /)
```
GabrielNagy, line 25: Can we just match `spec.required_ruby_version` and remove the entire line without matching the specific version? I'm not sure how to do that off the top of my head with sed but it might be a more future-proof fix.

gimmyxd:
Jenkins please test this on redhat7-POWERa,redhat8-POWERa,redhat8-AARCH64a,sles12-POWERa,solaris10-SPARCa,solaris11-SPARCa

PR #2103, opened 2021-05-18
(maint) Pin cmake on el-8-ppc64le

GabrielNagy:
RedHat recently updated their repos and the cmake version pulled is newer (3.18.2). This uses a libarchive method that is not available on 3.3.2 which is the one that comes with the box.

`cmake: symbol lookup error: cmake: undefined symbol: archive_write_add_filter_zstd`

This can either be fixed by updating the libarchive package to 3.3.3, or by pinning cmake to the previous working version (3.11.4). Pin cmake for now.


GabrielNagy:
jenkins please test this on redhat8-POWER

PR #2111, opened 2021-06-04
Fix name of environments folder

hdeadman:
I believe the folder is supposed to be called /etc/puppetlabs/code/environments/ rather than /etc/puppetlabs/code/environment/ and in the ubuntu image it is /etc/puppetlabs/code/environments/.


joshcooper:
Hi @hdeadman thanks for your contribution! Could you prefix your commit message summary with `(maint)`? This is needed for automation we use to synchronize github commits and JIRA tickets. Thanks!

hdeadman:
OK, I changed commit message and the PR title.

GabrielNagy (approved):

PR #2116, opened 2021-06-09
(PA-3751) Remove Debian 8 platforms and specific cases

Dorin-Pleava:


configs/components/cpp-hocon.rb
```diff
@@ -29,7 +29,7 @@
 
     cmake = "C:/ProgramData/chocolatey/bin/cmake.exe -G \"MinGW Makefiles\""
     toolchain = "-DCMAKE_TOOLCHAIN_FILE=#{settings[:tools_root]}/pl-build-toolchain.cmake"
-  elsif platform.name =~ /debian-[89]|el-[567]|redhatfips-7|sles-(:?11|12)|ubuntu-(:?14.04|16.04|18.04)/ ||
```
Dorin-Pleava, line 32: I didn't remove the debian-8-armel and debian-8-armhf(community-maintained platforms), so not sure if I should change the regex.

GabrielNagy (approved):

configs/components/cpp-hocon.rb
ciprianbadescu: debian-8-arm are not part of puppet-runtime platforms, is it still useful to keep them in puppet-agent?

PR #2117, opened 2021-06-09
(maint) pin resource_api to v1.8.14

mihaibuzgau:


configs/components/puppet-resource_api.json
```diff
@@ -1 +1 @@
-{"url":"git://github.com/puppetlabs/puppet-resource_api.git","ref":"b388f03e12be7ef383108dd6abe60419994611d6"}
+{"url":"git://github.com/puppetlabs/puppet-resource_api.git","ref":"refs/tags/1.8.14"}
```
gimmyxd, line 1: should the tag be `v1.8.14`? as i see it in https://github.com/puppetlabs/puppet-resource_api/tree/v1.8.14

mihaibuzgau: good catch

gimmyxd (approved):

PR #2124, opened 2021-06-15
(PA-3711) Add Debian 11 amd64 platform - main

GabrielNagy:


GabrielNagy:
CI fails because 7.8.0 is not yet available in the repos.

ciprianbadescu (approved):

PR #2134, opened 2021-07-14
(maint) Fix release-lead.rake to clone both facter and facter-ng

BobosilaVictor:


ext/release-lead.rake
```diff
@@ -78,7 +78,7 @@ namespace :release_lead do
           result[name] = 'No URL provided'
           next
         end
-        name = url_to_component_name(url)
+        name = url_to_component_name(url) == 'facter' ? json_name_to_component_name(component) : url_to_component_name(url)
```
gimmyxd, line 81: This is a bit hard to read, as you evaluate `url_to_component_name(url)` and then pass it to the short `if` and assign it back to `name`. I would suggest doing it one step at a time, like:
```ruby
name = url_to_component_name(URL)
name = json_name_to_component_name(component) if name == 'facter'
```

ext/release-lead.rake
```diff
@@ -79,6 +79,7 @@ namespace :release_lead do
           next
         end
         name = url_to_component_name(url)
+	name = json_name_to_component_name(component) if name == 'facter'
```
gimmyxd, line 82: nit: bad indentation :D

gimmyxd (approved):

PR #2143, opened 2021-07-30
(PA-3843) Add option to enable long path names in the Windows Installer

BobosilaVictor:
Add `Enable Long Path Names` checkbox at the ExitDialog window, which adds the LongPathsEnabled registry key, and sets its value to 1

Picture of the new option:
Screenshot 2021-07-30 at 14 18 33

Also added command line option to enable long path names (`ENABLE_LONG_PATHS`), and updated the powershell logic to check if the registry was already set correctly.

### DIFF
--- a/resources/windows/wix/customactions.wxs.erb
+++ b/resources/windows/wix/customactions.wxs.erb
@@ -253,9 +253,21 @@ End If
     <%-if @platform.architecture == "x86"-%>
     <Property Id="WixQuietExecCmdLine" Value="&quot;[%WINDIR]\System32\WindowsPowerShell\v1.0\powershell.exe&quot; -NoLogo -NonInteractive -InputFormat None -NoProfile -ExecutionPolicy Bypass -Command &quot;if (Test-Path -Path 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Components') { Get-ChildItem 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Components' -Recurse | foreach { foreach ($prop in $_.Property) { if($_.GetValue($prop) -like '*service\nssm.exe*') { Set-ItemProperty -Path $_.PSPath -Name $prop -Value '' -ErrorAction Stop } } }; Get-ChildItem 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Components' -Recurse | foreach { foreach ($prop in $_.Property) { if($_.GetValue($prop) -like '*puppet\bin\nssm.exe*') { Set-ItemProperty -Path $_.PSPath -Name $prop -Value '' -ErrorAction Stop } } } }&quot;"/>
     <CustomAction Id="RemoveLegacyNssmRegistryKey" BinaryKey="WixCA" DllEntry="WixQuietExec" Execute="immediate" Return="check"/>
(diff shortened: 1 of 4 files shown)


resources/windows/wix/customactions.wxs.erb
```diff
@@ -258,6 +258,13 @@ End If
     <CustomAction Id="RemoveLegacyNssmRegistryKey" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="immediate" Return="check"/>
     <%-end-%>
 
+    <Property Id="WixSilentExec64CmdLine"  Value="&quot;C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe&quot; New-ItemProperty -Path &quot;HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem&quot;  -Name &quot;LongPathsEnabled&quot; -Value 1 -PropertyType DWORD -Force"/>
```
gimmyxd, line 261: Should `C:Windows` be replaced with `[%WINDIR]` as in the above action?
Also de we need to take into account x86 vs x64? I see we do this for other actions

resources/windows/wix/customactions.wxs.erb
```diff
@@ -253,11 +253,25 @@ End If
...
     <CustomAction Id="RemoveLegacyNssmRegistryKey" BinaryKey="WixCA" DllEntry="WixQuietExec" Execute="immediate" Return="check"/>
+    <Property Id="WixSilentExecCmdLine"  Value="&quot;[%WINDIR]\System32\WindowsPowerShell\v1.0\powershell.exe&quot; New-ItemProperty -Path &quot;HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem&quot;  -Name &quot;LongPathsEnabled&quot; -Value 1 -PropertyType DWORD -Force"/>
```
GabrielNagy, line 256: We should have some logic that checks if the key is already set to 1, so we don't do it if the feature is already enabled.

GabrielNagy (commented):
Checkbox looks good :+1:.

We also need to expose this functionality as an [MSI install option](https://puppet.com/docs/puppet/6/install_agents.html#msi_properties) (ENABLE_LONG_PATHS maybe?) that defaults to false.

resources/windows/wix/sequences.wxs.erb
```diff
@@ -150,7 +150,9 @@
       <Custom Action='SetFromIniPuppetServerPort' Before='FileCost'>
         INI_PUPPET_SERVER_PORT AND NOT PUPPET_SERVER_PORT
       </Custom>
-
+       <Custom Action='EnableLongPathName' Before='InstallFiles'>
+        ENABLE_LONG_PATHS
```
ciprianbadescu, line 154: Could we prepend the property name with PUPPET_?

resources/windows/wix/properties.wxs.erb
```diff
@@ -50,6 +50,10 @@
       Id="WIXUI_EXITDIALOGOPTIONALTEXT"
       Value="Manage your first resources on this node, explore the Puppet community and get support using the shortcuts in the Documentation folder of your Start Menu." />
 
+    <Property
+      Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT"
+      Value="Enable Long Path Names" />
```
GabrielNagy, line 55: Maybe we can make this more detailed: `Disable Windows path length limit (260 character MAX_PATH limitation, requires Windows 10 1607 or later)`

Taken from https://github.com/qbittorrent/qBittorrent/blob/a2ef09466f3beb07c253e6ada2e1aa815d6c358d/dist/windows/installer.nsi#L170

resources/windows/wix/sequences.wxs.erb
BobosilaVictor: I don't think this would be a good idea, since this variable only changes system settings and isn't related to Puppet.

GabrielNagy:
jenkins please test this on windows

GabrielNagy (approved):
Tested the installation via UI (checkbox) and CLI (msiexec parameter) and the key is set correctly 🎉

PR #2185, opened 2022-01-11
(RE-14411) Add redhatfips-8 as an rpm target

e-gris:
redhatfips-8 as an rpm target is needed by puppet-enterprise-vanagon to build PE packages.


GabrielNagy (approved):
I was about to suggest you also do it for 6.x, but it looks like [I did it](https://github.com/puppetlabs/puppet-agent/commit/afc2853de3cd6ed7621f1510469777f9d55d0236) when I added the platform, and for some reason I missed this line on the `main` PR 🙈.

GabrielNagy:
Travis fails because the JIRA ticket is private, so I'm gonna go ahead and merge.

PR #2191, opened 2022-01-13
(maint) Pick up rhel 9 packaging support

e-gris:
Support for el-9 was added to packaging in 0.105.0; update the Gemfile to reflect that

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,8 +10,8 @@ def location_for(place)
   end
 end
 
-gem 'vanagon', *location_for(ENV['VANAGON_LOCATION'] || '~> 0.21')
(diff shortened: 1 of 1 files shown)


joshcooper:
Building puppet-agent on RHEL9 will still fail unless VANAGON_LOCATION is overridden to `https://github.com/puppetlabs/vanagon#main`. Can we tag vanagon at its current `HEAD` as `0.24.0` and then bump the minimum version in this PR? That way any pipeline that doesn't override either `VANAGON_LOCATION` or `PACKAGING_LOCATION` will continue to work with RHEL9.

joshcooper (approved):

PR #2192, opened 2022-01-17
(maint) Fix fact collection script on Windows

GabrielNagy:
Previously we were directly shelling out to cygwin on Windows boxes, causing the facter command to fail with:

    bash: facter: command not found

Fortunately, there's a helper `facter` method that properly namespaces the executable for the target platform:

    cmd.exe /c facter --show-legacy -p -j


BobosilaVictor (approved):
LGTM!

ciprianbadescu (approved):

PR #2197, opened 2022-02-01
(maint) Update osx signing server.

jackie-kinsler:


AriaXLi (commented):
There is another instance of osx-signer.delivery.puppetlabs.net in puppet-agent/ext/build_defaults.yaml. Should this also be changed to prod-2?

jackie-kinsler:
Most likely. Good catch. Updated that too.

PR #2198, opened 2022-02-03
(maint) In el-8-x86 platform switch centos-8 for redhat-8-x86_64

cthorn42:
With recent EOL of centos-8 builds have started to fail using centos-8.
This PR switches centos-8-x86_64 for redhat-8-x86_64.


cthorn42:
Tested this with `dnf install -y --allowerasing make rsync rpm-build` on a almalinux-8-x86_64 VM, and everything was able to install.

PR #2204, opened 2022-02-23
(PA-4185) (PA-4199) Add Ubuntu-20.04 and sles-15 as build targets

isaac-hammes:


cthorn42:
@isaac-hammes shouldn't this be targeted at 6.x and get merged up?

cthorn42:
It depends on what branches of PE this work is supposed to land in. If it is just the main branch, then this PR is fine. If it is intended to also land in the LTS branch, 2019.8.x, then I'd recommend just re-targeting this PR to the 6.x branch.

isaac-hammes:
> It depends on what branches of PE this work is supposed to land in. If it is just the main branch, then this PR is fine. If it is intended to also land in the LTS branch, 2019.8.x, then I'd recommend just re-targeting this PR to the 6.x branch.

Thanks. It is intended to be in both main and LTS, so I will do that.

PR #2209, opened 2022-03-11
(PA-4317) Snyk monitor pushes to 6.x

joshcooper:


joshcooper:
Tested `on pull_request` in https://github.com/puppetlabs/puppet-agent/runs/5517636499

PR #2218, opened 2022-03-18
(maint) Update smoke testing of platform components to postgres 11

cthorn42:
The Postgres 9.6 packages were removed from the repoes we are using.
Since on main, the puppetdb states postgres 11 is a requirement, lets just
bump to that.
This allows the `run-smoke-test.sh` script to pass now.


cthorn42:
Used this PR to validate PA-4302.

cthorn42:
Specifically we were using  https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-42.0-23.noarch.rpm, but that got bumped to 42.0.24, and with that bump postgresql 9.6 was removed.

mhashizume (approved):

PR #2219, opened 2022-03-22
Build Facter for JRuby on sles15 and redhatfips8

Magisus:


Magisus:
These built alright for me.

configs/components/facter.rb
```diff
@@ -52,6 +52,9 @@
   when /debian-10|debian-11|ubuntu-20/
     pkg.build_requires 'openjdk-11-jdk'
     java_home = "/usr/lib/jvm/java-11-openjdk-#{platform.architecture}"
+  when /sles-15/
+    pkg.build_requires 'java-1_8_0-openjdk-devel'
```
justinstoller, line 56: Is there not a Java 11 available? I guess this keeps it relatively similar to the EL rpms, just feels weird to be adding a Java 1.8 dependency nowadays.

Magisus: I can look, I went for 8 because the older versions are using 7, and 11 seemed like a big jump. Also we're still requiring Java 8 for the server packages, rather than 11. If 11 is available, do you want me to look into also making the server packages depend on Java 11?

justinstoller: No    : (   
I don't think we should be adding Java 8 to platforms, but it's probably good to have it be the same as RHEL 8 as they have similar support life times. When we do drop Java 8 we just change all the RPM packages at once.

mwaggett (approved):

configs/components/facter.rb
Magisus: Yeah I think generally we've only done Java 11 for systems where it's hard/impossible to install Java 8. For better or for worse... Also, this only affect 6.x, 7 doesn't need this because it's Ruby Facter. So at least there's that.

cthorn42 (approved):

PR #2228, opened 2022-03-29
(PA-4337) Update libwhereami special_flags for MacOS

cthorn42:
MacOS 12 fails to build libwhereami due to a flag not being added in that
component. We've added the flag to other components so this looks like an oversight.

Also adds back the platform definition for osx-12-x86_64.


cthorn42:
Tested this with a vanagon gem that is pointed at the head of the main branch, was able to run `bundle exec vanagon build puppet-agent osx-12-x86_64 <macos-12-x86_64 VM I checked out`, and was able to build a puppet-agent dmg package.

mhashizume (approved):

PR #2241, opened 2022-04-26
(PA-4239) Add Ubuntu 22.04

yakatz:


cthorn42:
I changed the base from 6.x to main, since 6.x is going EOL pretty soon. We're trying to land any long term changes like this on main.

yakatz:
> On second thought landing this on 6.x is fine.

I was just following the instructions (on Jira)

cthorn42:
Thanks, yup that is my bad. I validated this with the puppet-runtime and pxp-agent changes and was able to build a puppet7 tarball for ubuntu 2204 x86_64. So i think this is good to merge.
One big caviet is we wont be testing the arm64 work right away, it might take us a week or two to get VMs for that setup. We'll go ahead and merge this, but we wont build or test arm64. Once we have the VMs for that we'll add that. 
Thanks again @yakatz for the contribution.

mpdude:
Was the intent of this change to make `arm64` and `amd64` versions of puppet-agent available for Ubuntu 22.04?

If so, should they be available through https://apt.puppet.com?

PR #2242, opened 2022-04-26
(PA-4240) Add Ubuntu 22.04

yakatz:
Does `deb_targets` also need to be changed? I see it was done as separate commits for 20.04, but I don't see a ticket mentioning it in Jira.


joshcooper:
Thanks @yakatz ! Your other PR targeting `6.x` was merged in https://github.com/puppetlabs/puppet-agent/pull/2241 and merged into `main` so I'm going to close this.

PR #2244, opened 2022-05-16
(maint) Updates Rubocop version in Gemfile

mhashizume:
Rubocop 1.29 drops support for Ruby 2.5. The 6.x branch of puppet-
agent use Ruby 2.5.9, so we'll set Rubocop in the Gemfile to
"< 1.29".

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,6 @@ gem 'artifactory'
 gem 'rake'
 gem 'json'
 gem 'octokit'
-gem 'rubocop', "~> 1.22"
+gem 'rubocop', "< 1.29"
 
 eval_gemfile("#{__FILE__}.local") if File.exist?("#{__FILE__}.local")


mhashizume:
@joshcooper I hit this when running a `bundle install` in the repo on my local machine, though now that you point it out, it does say it needed to revert to an older version of Bundler:
```
 bundle install
Bundler 2.3.8 is running, but your lockfile was generated with 2.1.4. Installing Bundler 2.1.4 and restarting using that version.
Fetching gem metadata from https://rubygems.org/.
Fetching bundler 2.1.4
Installing bundler 2.1.4
Fetching gem metadata from https://rubygems.org/......
rubocop-1.29.0 requires ruby version >= 2.6.0, which is incompatible with the current version, ruby 2.5.9p229
```
From what you're saying, it sounds like that's more an issue with my local setup though, so we can close this if it's unnecessary.

PR #2245, opened 2022-05-17
(maint) Merge 6.x to main

joshcooper:
* upstream/6.x:
  (maint) Update puppet to b6f5ca628424af3650bb682a88a440d7078fd952
  (PA-4463) Add macOS 12 Monterey ARM64 (M1) support

Conflicts:
	configs/components/cpp-hocon.rb
	configs/components/cpp-pcp-client.rb
	configs/components/facter.rb
	configs/components/leatherman.rb
	configs/components/libwhereami.rb
	configs/components/puppet.json
	configs/components/pxp-agent.rb

puppet-agent#6.x included changes to cross-compile on macOS 12 ARM, but that's
no longer handled in puppet-agent#main, instead we use pxp-agent-vanagon and
the agent-puppet-main runtime.

### DIFF
--- a/configs/platforms/osx-12-arm64.rb
+++ b/configs/platforms/osx-12-arm64.rb
@@ -1,0 +1,29 @@
+platform 'osx-12-arm64' do |plat|
+    plat.servicetype 'launchd'
+    plat.servicedir '/Library/LaunchDaemons'
+    plat.codename 'monterey'
+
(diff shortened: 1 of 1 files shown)


joshcooper:
The commit message check is failing because there's a commit with `(RE-xxx)` but it's not fixable.

mhashizume (approved):

PR #2246, opened 2022-05-20
(maint) Update Apple macOS signing server

mhashizume:


joshcooper:
This seems good to me, but I'm not sure what other repos need to be updated? And do they all need to be updated at the same time?

This was the only other reference I could find:

https://github.com/puppetlabs/build-data/blob/9a5d91df80bc7e5e727dbae76816029b23e9083c/builder_data.yaml#L124

It doesn't appear packaging, vanagon or ci-job-configs know about it, so that's good.

Is there a way to test this PR in adhoc to verify signing works?

mhashizume:
jenkins please test this on osx12-64

mhashizume:
jenkins please test this on osx12-64

mhashizume:
Successful PR test (with build-data/packaging changes pulled in) here: https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-packaging_pr/BUILD_TARGET=osx-12-x86_64,SLAVE_LABEL=k8s-worker/2682/console

Will need to drop e1d551b3ae9ee52ede51b936decde37eff01aab3 before merging.

AriaXLi (approved):

PR #2251, opened 2022-06-06
(maint) Change ubuntu-22.04 to inherit from defaults

cthorn42:
The default settings (as set in vanagon) for ubuntu-22.04 include 'make'. That
is a hard requirement that we need to make sure is installed. We didn't include that
in our platform definition here, but we do in the defaults. So lets just inherit those.


cthorn42:
I forgot to mention we have a change in vanagon from RE that unblocked us: https://github.com/puppetlabs/vanagon/pull/755.
So we'll need to merge that also, and then we can do a release.

PR #2255, opened 2022-06-09
(maint) Change back the osx-12 x86_64 naming convention

cthorn42:
PE is built to assume the osx intel 64 bit tarball is named without the x86_64
in the file name. Changing this breaks things in PE.


cthorn42:
Closing this in favor of fixing in pe_repo.

joshcooper:
I still don't understand why the `platform_repos` controls the name of the tarball, but I'm guessing that logic lives in the packaging repo?

PR #2269, opened 2022-07-21
(maint) Bump the hiera refrence to point to 3.10.0

cthorn42:


configs/components/hiera.json
```diff
@@ -1 +1 @@
-{"url":"https://github.com/puppetlabs/hiera.git","ref":"84dbdb6f74b6f138347ccb868be033ef37ff0bcf"}
+{"url":"https://github.com/puppetlabs/hiera.git","ref":"68782f0"}
```
joshcooper, line 1: I'm not 100% sure ci-job-configs will handle a short SHA when it goes to auto bump, I'd use the full SHA to be sure.

cthorn42: 👍

joshcooper (approved):

PR #2279, opened 2022-09-08
(maint) Updates sqlite3 install command

mhashizume:
Starting with version 1.5.0 of the sqlite3 gem, it bundles in libsqlite3, which breaks our tests. This commit adds a flag (`--enable-system-libraries`) to continue to use system libsqlite3.


joshcooper (approved):
LGTM I tested this on a few different platforms

PR #2280, opened 2022-09-08
(PA-4560) Switch from TravisCI to GH Actions

AriaXLi:


.github/workflows/checks.yaml
```diff
@@ -0,0 +1,32 @@
+name: Checks
+
+on:
+  push:
+    branches: [main]
```
mhashizume, line 5: If we're merging this to the 6.x branch, we should either change this to include both main and 6.x or just 6.x and we can change this field in main

AriaXLi: Ah shoot, thank you! I changed it so it included both main and 6.x but I'm open to having just 6.x if that's preferred. I used the syntax used [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-filters) but I wasn't 100% sure so I used act to test it again and the job succeeded

mhashizume (approved):

PR #2282, opened 2022-09-12
(maint) Updates Postgres URL

mhashizume:


joshcooper (approved):
LGTM, there's also a latest `https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm` not sure if we should use that?

mhashizume:
Confirmed that `latest` is the same as the versioned URL:
```
[root@fierce-psyllium ~]# yum --releasever=7 install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
. . .
=============================================================================================================================================================================================
 Package                                       Arch                                Version                                Repository                                                    Size
=============================================================================================================================================================================================
Installing:
 pgdg-redhat-repo                              noarch                              42.0-26                                /pgdg-redhat-repo-latest.noarch                               12 k
```

PR #2289, opened 2022-09-21
(PA-4686) Add acceptance test to validate Augeas package and resource type

cthorn42:
We had an escaped bug that broke Augeas for a certian OS type. This test will catch that in the future.

### DIFF
--- a/acceptance/tests/augeas_validation.rb
+++ b/acceptance/tests/augeas_validation.rb
@@ -1,0 +1,38 @@
+skip_test 'non-windows only test' if hosts.any? { |host| host.platform =~ /windows/ }
+tag 'audit:high'
+test_name 'Augeas Validation' do
+  teardown do
+    hosts.each do |agent|
+      file = <<-EOF
+augeas { 'test_ssh':
+  lens    => 'Ssh.lns',
+  incl    => '/etc/ssh/ssh_config',
+  context => '/files/etc/ssh/ssh_config',
+  changes => [
+    'remove Host testing.testville.nil'
+  ]
+}
+EOF
+      on(agent, "puppet apply -e \"#{file}\"")
+    end
+  end
+
+  hosts.each do |agent|
+    step 'validate Augeas binary' do
+      on(agent, '/opt/puppetlabs/puppet/bin/augtool --version')
+    end
+    step 'validate we can apply a resource type augeas' do
+      file = <<-EOF
+augeas { 'test_ssh':
+  lens    => 'Ssh.lns',
+  incl    => '/etc/ssh/ssh_config',
+  context => '/files/etc/ssh/ssh_config',
+  changes => [
+    'set Host testing.testville.nil'
+  ]
+}
+EOF
+      assert_equal(on(agent, "puppet apply -e \"#{file}\"").exit_code, 0, 'Puppet apply of the augeas resource returned a non-zero exit code')
+    end
+  end
+end


cthorn42:
FWIW test takes under 5 seconds:
```
              - Test Case Summary for suite 'tests' -
       Total Suite Time: 4.84 seconds
      Average Test Time: 4.84 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```
The DNM tag is because we need runtime to get promoted and nightlies to run. Also I want to run this against more platforms then just centos-7-x86_64

cthorn42:
This is good to go, I created an adhoc job here testing this against all our supported platforms and it passed: https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1063/
There is an issue with beaker-puppet and the Win-11-ent cell, but the test is skipping on all Windows platform so I'm not worried.

acceptance/tests/augeas_validation.rb
```diff
@@ -0,0 +1,42 @@
...
+  lens    => 'Ssh.lns',
+  incl    => '/etc/ssh/ssh_config',
+  context => '/files/etc/ssh/ssh_config',
+  changes => [
+    'set Host testing.testville.nil'
+  ]
+}
+EOF
+      create_remote_file(agent, '/tmp/augeas_validation.pp', file)
+      on(agent, '/opt/puppetlabs/puppet/bin/puppet apply /tmp/augeas_validation.pp')
```
tvpartytonight, line 39: Should we have an assert here? It feels a little bit unclear that we are just relying on the apply not returning a non-zero exit code as the test, but I could really go either way.

Also, instead creating a remote file, you could try inlining the code so that it reduces some network traffic? I tested this on a pooler instance and it looks to work fine:
```
[4] pry(#<Beaker::TestCase>)> on(agent, "puppet apply -e \"#{file}\"")

maroon-elevator.delivery.puppetlabs.net (maroon-elevator.delivery.puppetlabs.net) 12:11:11$ puppet apply -e "augeas { 'test_ssh':
  lens    => 'Ssh.lns',
  incl    => '/etc/ssh/ssh_config',
  context => '/files/etc/ssh/ssh_config',
  changes => [
    'set Host testing.testville.nil'
  ]
}
"
  Notice: Compiled catalog for maroon-elevator.delivery.puppetlabs.net in environment production in 0.06 seconds
  Notice: /Stage[main]/Main/Augeas[test_ssh]/returns: executed successfully
  Notice: Applied catalog in 0.23 seconds
```

tvpartytonight (approved):

acceptance/tests/augeas_validation.rb
```diff
@@ -0,0 +1,42 @@
...
+  lens    => 'Ssh.lns',
+  incl    => '/etc/ssh/ssh_config',
+  context => '/files/etc/ssh/ssh_config',
+  changes => [
+    'remove Host testing.testville.nil'
+  ]
+}
+EOF
+      create_remote_file(agent, '/tmp/augeas_validation_teardown.pp', file)
+      on(agent, '/opt/puppetlabs/puppet/bin/puppet apply /tmp/augeas_validation_teardown.pp')
```
tvpartytonight, line 17: Oh, and do we normally fully qualify the `puppet` path? Seems fine to just assume it is installed and available?

cthorn42: I think the way we setup our acceptance tests a pre-suite does set the PATH to include this, so yeh it probably isn't needed. 
The habit comes from working on PE acceptance_tests where we didn't muck with the PATH as part of the pre-suite, and I think maybe at one point there was a bug with SPARC or AIX didn't get symlinks created to some of the Puppet binaries.

PR #2305, opened 2022-11-21
(PA-4749) Add solaris-10-sparc

cthorn42:
Also fix the solaris-10-x86_64 output dir, was pointing at puppet6. Added back in some logic to the pl-ruby-patch for the older version of ruby that is on solaris-10.

### DIFF
--- a/configs/components/pl-ruby-patch.rb
+++ b/configs/components/pl-ruby-patch.rb
@@ -26,21 +26,23 @@ component "pl-ruby-patch" do |pkg, settings, platform|
                 end
 
     # weird architecture naming conventions...
-    target_triple = if platform.architecture =~ /ppc64el|ppc64le/
-                      "powerpc64le-linux"
-                    elsif platform.name == 'solaris-11-sparc'
-                      "sparc-solaris-2.11"
-                    elsif platform.is_macos?
-                      "aarch64-darwin"
-                    else
-                      "#{platform.architecture}-linux"
-                    end
+    unless platform.name =~ /solaris-10/
+      target_triple = if platform.architecture =~ /ppc64el|ppc64le/
+                        "powerpc64le-linux"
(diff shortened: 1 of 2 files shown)


configs/components/pl-ruby-patch.rb
```diff
@@ -26,21 +26,23 @@
...
-    target_triple = if platform.architecture =~ /ppc64el|ppc64le/
-                      "powerpc64le-linux"
-                    elsif platform.name == 'solaris-11-sparc'
-                      "sparc-solaris-2.11"
-                    elsif platform.is_macos?
-                      "aarch64-darwin"
-                    else
-                      "#{platform.architecture}-linux"
-                    end
+    unless platform.name =~ /solaris-10/
```
tvpartytonight, line 29: `solaris-10-i386` will now be caught in this `unless`, is that intentional?

cthorn42: Sorry, this was copied over from 6.x and I didn't include the comment above it:
# solaris 10 uses ruby 2.0 which doesn't install native extensions based on architecture
This might also be the cause of the testing issue we're seeing on solaris-10-x86_64 where we attempt to validated the vendored ruby: https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/Acceptance%20Suites/view/main/view/Suite/job/platform_puppet-agent_puppet-agent-integration-suite_daily-main/901/RMM_COMPONENT_TO_TEST_NAME=puppet_agent,SLAVE_LABEL=k8s-beaker,TEST_TARGET=solaris10-64a/testReport/(root)/tests/validate_vendored_ruby_rb/

cthorn42: So I think we need this change to also affect solaris-10-i386|x86_64 for the above issue.

tvpartytonight (approved):

PR #2306, opened 2022-11-28
(PA-4751) Add Solaris 10 (SPARC) to build_defaults for nightlies

AriaXLi:


joshcooper (approved):
:+1: looks like it matches the last passing build: https://builds.delivery.puppetlabs.net/puppet-agent/aad1d59df063ddf85b7a432460ce1652a5ab4347/repos/solaris/10/puppet7/

PR #2307, opened 2022-11-28
Manual 6.x mergeup

mhashizume:
Accepting the minor spelling differences in the Solaris 10 SPARC platform definition so this doesn't come up again.


joshcooper (approved):
Changes look good, but since we recently created the 7.x branch, could you target this for 7.x? Might be good to also do a manual mergeup from 7.x to main until we get the pipelines sorted out?

PR #2313, opened 2022-12-12
(PA-4814) Bump version string and puppet collections/streams for puppet 8

AriaXLi:


AriaXLi:
I ran vanagon-generic-builder for puppet-agent with the plat.output_dir command removed and the directory for each platform used the default (puppet8) defined in ext/build_defaults.yaml. 

See here: http://builds.delivery.puppetlabs.net/puppet-agent/fad95d90535d20db763b27c39464d361cb8604a4/artifacts/?C=M&O=A

joshcooper (approved):

PR #2318, opened 2023-01-12
(maint) Update the puppet-agent project file to use the 7.x branch

cthorn42:
The puppet-agent vanagon build needs to know the name of the branch in order to download the puppet and pxp-agent runtimes.

### DIFF
--- a/configs/projects/puppet-agent.rb
+++ b/configs/projects/puppet-agent.rb
@@ -6,7 +6,7 @@ project "puppet-agent" do |proj|
   # - Settings included in this file should apply only to local components in this repository.
   runtime_details = JSON.parse(File.read('configs/components/puppet-runtime.json'))
   pxp_agent_details = JSON.parse(File.read('configs/components/pxp-agent.json'))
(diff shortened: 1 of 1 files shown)


cthorn42:
Currently 7.x branch is failing CI because it is attempting to download this runtime:
```
https://builds.delivery.puppetlabs.net/puppet-runtime/202301102/artifacts/agent-runtime-main-202301102.aix-7.1-ppc.settings.yaml.sha1
```
note the 'main' in there, that is the problem because we are now creating the 7.x runtime that this branch should be consuming.

joshcooper:
Ah nice catch! Yeah that makes sense. pxp-agent-vanagon does the same (since it inherits settings from the puppet-runtime): 

https://github.com/puppetlabs/pxp-agent-vanagon/blob/1ae651cb50674ca459be4e43523e886c26d021a7/configs/projects/pxp-agent.rb#L18

One thing, it looks like there's an extra merge commit in there, might need to do:

```
git reset --hard 9a69f5809
git push -f <your remote>
```

PR #2320, opened 2023-01-16
(PA-4080) Remove old packaging obsoletes and provides

tvpartytonight:
This change removes old cruft about packages that the puppet-agent would remove on installation of new versions.

### DIFF
--- a/configs/components/puppet.rb
+++ b/configs/components/puppet.rb
@@ -19,17 +19,6 @@ component "puppet" do |pkg, settings, platform|
     pkg.build_requires "gettext"
   end
 
-  pkg.replaces 'puppet', '4.0.0'
-  pkg.provides 'puppet', '4.0.0'
-
-  pkg.replaces 'pe-puppet'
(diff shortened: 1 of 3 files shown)


configs/components/puppet.rb
```diff
@@ -19,17 +19,6 @@
...
-  pkg.provides 'puppet', '4.0.0'
-
-  pkg.replaces 'pe-puppet'
-  pkg.replaces 'pe-ruby-rgen'
-  pkg.replaces 'pe-cloud-provisioner-libs'
-  pkg.replaces 'pe-cloud-provisioner'
-  pkg.replaces 'pe-puppet-enterprise-release', '4.0.0'
-  pkg.replaces 'pe-agent'
-  pkg.replaces 'pe-puppetserver-common', '4.0.0'
-
```
joshcooper, line 32: Should the lines below be deleted too?

joshcooper (approved):
I'll add a follow up PR

PR #2332, opened 2023-02-15
(MODULES-11371) Update core module references

mhashizume:
This commit updates the references for all core module Vanagon components, enabling compatibilty with Puppet 8 and other minor changes.

### DIFF
--- a/configs/components/module-puppetlabs-augeas_core.json
+++ b/configs/components/module-puppetlabs-augeas_core.json
@@ -1,1 +1,1 @@
-{"url":"git@github.com:puppetlabs/puppetlabs-augeas_core.git","ref":"refs/tags/1.1.2"}
+{"url":"git@github.com:puppetlabs/puppetlabs-augeas_core.git","ref":"refs/tags/v1.3.0"}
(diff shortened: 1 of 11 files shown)


mhashizume:
> Wondering if we should also bump `puppetlabs-scheduled_task`? We're currently vendoring `1.0.0` which is from 2018. I think we'll want [v2.3.1](https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/a6b3ee6626c9caee96d0ca9791026fabfd977231/metadata.json#L33) in puppet-agent#7.x or maybe [v3.1.1](https://github.com/puppetlabs/puppetlabs-scheduled_task/blob/v3.1.1/metadata.json). Maybe file a CONT ticket?

Good catch on that, thanks Josh. I've added another commit just for `scheduled_task`.

What did you want to submit a ticket for in CONT? To bump the `metadata.json` to allow for puppet8 and assure compatibility?

joshcooper:
Ah I'm just not sure if there are breaking changes in 3.1.1 that we don't want in puppet-agent#7.x?

joshcooper (approved):

mhashizume:
The CAT team say the only breaking changes they're aware of is end-of-life OSes and Puppet versions: https://perforce.slack.com/archives/C049PL6AUCU/p1676498307080819?thread_ts=1676496650.822039&cid=C049PL6AUCU

Sounds like we're good to go, going to go ahead and merge this.

PR #2337, opened 2023-03-23
(maint) Removes unused platforms

mhashizume:
We do not plan on shipping either Red Hat Enterprise Linux (RHEL) 6 nor macOS 10.15 Catalina in puppet-agent 8. This commit removes these platforms from puppet-agent's Vanagon configuration.


joshcooper (approved):

joshcooper:
I think we can delete https://github.com/puppetlabs/puppet-agent/blob/main/configs/platforms/osx-11-arm64.rb too if you want

PR #2339, opened 2023-03-30
(PUP-11786) Enable Ruby 3.1 compatibility

mhashizume:


mhashizume:
We're taking a different route with this and backporting some changes to Beaker 4 instead of directly switching over to Beaker 5. Closing PR.

PR #2344, opened 2023-04-20
(packaging) PE-16465 Add windows and windowsfips to platform_repos

e-gris:
This adds the windows and windowsfips repos to the
puppet-agent-all.tar.gz tarball.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -94,6 +94,11 @@ platform_repos:
     repo_location: repos/solaris/11/**/*.i386.p5p
   - name: solaris-11-sparc
     repo_location: repos/solaris/11/**/*.sparc.p5p
+  - name: windows
+    repo_location: repos/windows/puppet-agent-[0-9]*.msi
+  - name: windowsfips
+    repo_location: repos/windowsfips/puppet-agent-[0-9]*.msi
+
 deb_targets: 'xenial-amd64 bionic-amd64 focal-amd64 jammy-amd64'
 rpm_targets: 'el-6-x86_64 el-7-x86_64 el-8-x86_64 el-9-x86_64 redhatfips-7-x86_64 redhatfips-8-x86_64 sles-12-x86_64 sles-15-x86_64'
 sign_tar: FALSE


e-gris:
Oops. Let me try that again.

joshcooper:
For posterity, the `platform_repos` hash is [accessed in the `packaging` repo](https://github.com/puppetlabs/packaging/blob/c7fdccb2b5fcfee1392ec5fd32226e9ea1a31b69/lib/packaging/repo.rb#L93)

which is called by the [`pl:jenkins:pack_all_signed_repos_individually` rake task](https://github.com/puppetlabs/packaging/blob/c7fdccb2b5fcfee1392ec5fd32226e9ea1a31b69/tasks/nightly_repos.rake#L65)

And that task is called when performing an [internal promotion](https://github.com/puppetlabs/ci-job-configs/blob/f9f03b25f4fc9385ba0f685ffe5a3ffa7746b900/resources/scripts/puppet-agent-internal-promotion.sh#L177) and [shipping to S3](https://github.com/puppetlabs/ci-job-configs/blob/f9f03b25f4fc9385ba0f685ffe5a3ffa7746b900/resources/scripts/ship-agent-to-s3.sh#L64)

Long term I think it would be preferable for this data to live in CJC, so that the list of platforms we build/test/tar can be versioned together, and it doesn't require a puppet-agent change if we forget to add a platform.

ext/build_defaults.yaml
```diff
@@ -100,6 +100,11 @@ platform_repos:
     repo_location: repos/solaris/11/**/*.i386.p5p
   - name: solaris-11-sparc
     repo_location: repos/solaris/11/**/*.sparc.p5p
+  - name: windows
+    repo_location: repos/windows
+  - name: windowsfips
+    repo_location: repos/windowsfips
```
joshcooper, line 106: Suggest restricting `repo_location: repos/windows/*msi` and same for `windowsfips` so that we don't include the `wixpdb` files, for example, here's the complete directory for 7.24.0: https://builds.delivery.puppetlabs.net/puppet-agent/7.24.0/repos/windows/

I think we also want to exclude the latest symlinks, otherwise we'll end up with two copies of the 32 and 64-bit MSI. I'm not sure how to do that as a file pattern?

joshcooper:
@e-gris could you rebase your PR on upstream/7.x and squash the shrug commit?

joshcooper:
Cool thanks, running this through CI https://jenkins-platform.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/2047/

joshcooper:
@e-gris sorry could you rebase (again)?

PR #2346, opened 2023-05-04
(PA-5398) Add fips shared library

joshcooper:
* Adds fips.so shared library component to puppet-agent on RHEL 7 & 8 FIPS
* Runs `openssl fipsinstall` as a post-install task and enables the fips provider
* Adds beaker tests to verify openssl version and status of the fips provider

Also see the [OpenSSL 3.0.0 FIPS provider](https://github.com/puppetlabs/openssl-fips/blob/main/README.md).

Open Issues:
- [x] Should the post-install task run on upgrades too? Yes
- [x] Rebase on main after https://github.com/puppetlabs/puppet-runtime/pull/663 is merged
- [x] Run adhoc pipeline on all [redhatfips platforms](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1113/)
- [x] Run [main pipeline](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/Acceptance%20Suites/view/main/view/Suite/job/platform_puppet-agent_puppet-agent-integration-suite_daily-main/1026/) with latest [`puppet-runtime#202305040`](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/Acceptance%20Suites/view/main/view/Suite/job/platform_puppet-agent_puppet-agent-integration-suite_daily-main/1026/)
- [x] validate_vendored_openssl passed on [rhelfips7](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1107/RMM_COMPONENT_TO_TEST_NAME=puppet_agent,SLAVE_LABEL=k8s-beaker,TEST_TARGET=redhatfips7-64a/testReport/(root)/tests/validate_vendored_openssl_rb/) and [rhelfips8](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1107/RMM_COMPONENT_TO_TEST_NAME=puppet_agent,SLAVE_LABEL=k8s-beaker,TEST_TARGET=redhatfips8-64a/testReport/(root)/tests/validate_vendored_openssl_rb/)

### DIFF
--- a/configs/components/openssl-fips.rb
+++ b/configs/components/openssl-fips.rb
@@ -1,0 +1,65 @@
+component "openssl-fips" do |pkg, settings, platform|
+
+  pkg.build_requires "puppet-runtime"
+
+  openssl_fips_details = JSON.parse(File.read('configs/components/openssl-fips.json'))
+  openssl_fips_version = openssl_fips_details['version']
+  openssl_fips_location = openssl_fips_details['location']
+
+  tarball_name = "#{pkg.get_name}-#{openssl_fips_version}.#{platform.name}.tar.gz"
+  pkg.url File.join(openssl_fips_location, tarball_name)
+  pkg.sha1sum File.join(openssl_fips_location, "#{tarball_name}.sha1")
+  pkg.version openssl_fips_version
+  pkg.install_only true
+
+  pkg.add_source("file://resources/patches/openssl/openssl-fips.cnf.patch")
+
+  # Overlay openssl-fips shared library onto puppet-runtime in /opt at *build* time
+  #
+  # Don't generate `fipsmodule.cnf` during the build. It must be generated
+  # during postinstall action.
+  #
+  # Don't include FIPS-enabled `openssl.cnf` in the package. It must be moved
+  # into place after `fipsmodule.cnf` is generated. So ship `openssl-fips.cnf`
+  # and rename it to `openssl.cnf` during postinstall action.
+  #
+  # REMIND: update for Windows
+  pkg.install do
+    [
+      "#{platform.tar} --skip-old-files --directory=/ --extract --gunzip --file=#{tarball_name}",
+      "mv /opt/puppetlabs/puppet/ssl/openssl.cnf /opt/puppetlabs/puppet/ssl/openssl-fips.cnf",
+      "#{platform.patch} --strip=1 --fuzz=0 --ignore-whitespace --no-backup-if-mismatch -d /opt/puppetlabs/puppet/ssl/ < openssl-fips.cnf.patch"
+    ]
+  end
+
+  # See "Making all applications use the FIPS module by default" in
+  # https://www.openssl.org/docs/man3.0/man7/fips_module.html
+  #
+  # We have to run `openssl fipsinstall` to generate `fipsmodule.cnf`. This
+  # generates an HMAC for the `fips.so` and stores that value in `fipsmodule.cnf`.
+  # Only later can we enable fips, by overwriting `openssl.cnf` with the fips
+  # enabled version.
+  #
+  # During an upgrade, `fips.so` may be modified, invaliding the checksum in
+  # `fipsmodule.cnf`, and that will prevent `openssl fipsinstall` from working.
+  # So override OPENSSL_CONF to use the pristine `openssl.cnf` without fips
+  # enabled to regenerate `fipsmodule.conf`
+  #
+  # REMIND: update for Windows
+  pkg.add_postinstall_action ["install", "upgrade"],
+    [<<-HERE.undent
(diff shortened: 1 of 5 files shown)


AriaXLi (approved):

configs/components/openssl-fips.rb
```diff
@@ -0,0 +1,56 @@
...
+  # enabled version.
+  #
+  # During an upgrade, `fips.so` may be modified, invaliding the checksum in
+  # `fipsmodule.cnf`, and that will prevent `openssl fipsinstall` from working.
+  # So override OPENSSL_CONF to use the pristine `openssl.cnf` without fips
+  # enabled to regenerate `fipsmodule.conf`
+  #
+  # The `|| :` prevents the install/upgrade from exiting if the command fails.
+  #
+  pkg.add_postinstall_action ["install", "upgrade"],
```
e-gris, line 50: I would probably add preremove_action to clean up the generated files: fipsmodule.cnf && openssl.cnf

During the postinstall, I would also suggest that the same have their ownership and file permissions carefully set to the minimum needed. `openssl fipsinstall` may do this but I would still explicitly do it here as well.

configs/components/openssl-fips.rb
```diff
@@ -0,0 +1,56 @@
...
+  # include it in the package.
+  #
+  # openssl.cnf also needs to be updated at install time, so don't include it in
+  # the package. Instead ship `openssl-fips.cnf` and overwrite `openssl.cnf`
+  # with that version after the `fips.so` library is installed.
+  #
+  pkg.install do
+    [
+      "gunzip -c #{tarball_name} | #{platform.tar} --skip-old-files -C / -xf -",
+      "rm -f /opt/puppetlabs/puppet/ssl/fipsmodule.cnf",
```
e-gris, line 29: I would be tempted to not do this or to remove the `-f`, assuming the preremove_action suggestion, below. If something is not as expected, I rather see an error than it to silently self-correct. 

This has the potential of unwanted annoyance through corner case(s) but I think I'd start with it.

configs/components/openssl-fips.rb
```diff
@@ -0,0 +1,56 @@
...
+  # fipsmodule.cnf needs to be generated at install time on the host, so never
+  # include it in the package.
+  #
+  # openssl.cnf also needs to be updated at install time, so don't include it in
+  # the package. Instead ship `openssl-fips.cnf` and overwrite `openssl.cnf`
+  # with that version after the `fips.so` library is installed.
+  #
+  pkg.install do
+    [
+      "gunzip -c #{tarball_name} | #{platform.tar} --skip-old-files -C / -xf -",
```
e-gris, line 28: Trivial/ignorable: I like using --long-flags as much as portability allows. I also don't _think_ these acrobatics are needed any more since all of the tars we encounter can handle gzip tarballs:

`#{platform.tar} --skip-old-files --directory=/ --extract --file=#{tarball_name}`

e-gris (approved):
This looks right to me. I had a couple of minor comments that can be ignored.

configs/components/openssl-fips.rb
joshcooper: Updated

configs/components/openssl-fips.rb
joshcooper: I added a `preremove_action` and explicitly `chmod` both files to `0644`, which matches what I see on other distributions, e.g. 

```
$ ls -la /etc/ssl/openssl.cnf           
-rw-r--r-- 1 root 11K Apr 20  2020 /etc/ssl/openssl.cnf
```

configs/components/openssl-fips.rb
joshcooper: So this `--force` is because when building the package, I want to ensure the `fipsmodule.cnf` that the `openssl-fips` tarball laid down overlaid does not "leak" into the puppet-agent package. But if in the future `fipsmodule.cnf` is removed from the `openssl-fips` tarball, then it's ok to proceed.

joshcooper: Removed `fipsmodule.cnf` in https://github.com/puppetlabs/openssl-fips/commit/6e94106d4e0796d881e294ca494c2c0a6d678ddf and removed [`rm`](https://github.com/puppetlabs/puppet-agent/compare/63787dfb2beb416732e52ec0903bab077ebf91d4..91df5ad0234e67560c5df10c3ab3045daad03cc9) command

PR #2361, opened 2023-07-12
(PA-4594) added macos13(x86_64) platform

amitkarsale:
added platform support for macos-13(x86_64)


joshcooper (approved):
This will need to be backported to 7.x too (and replace `puppet8` with `puppet7`)

amitkarsale:
@joshcooper , I'll do the needful. Thanks!

PR #2365, opened 2023-07-18
(PA-5661) Ship AIX 7.2 repo tarball and add rpm to all agent tarball

joshcooper:
Previously, we were shipping AIX rpms to the internal build server and agent
downloads, but we weren't creating the AIX repo tarball:

    puppet-agent/<version>/repos/puppet-agent-aix-7.2-power.tar.gz

Nor was the AIX RPM included in the all agents tarball during

    Execute pl:jenkins:pack_all_signed_repos_individually

Adding aix-7.2-power to the `platform_repos` enables both of those things.

Also some discussion in https://perforce.slack.com/archives/G049EJF2Z2A/p1689703744853649 about why we're removing `pe_platforms`

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -25,13 +25,6 @@ foss_platforms:
   - ubuntu-22.04-aarch64
   - windows-2012-x86
   - windows-2012-x64
-pe_platforms:
-  # - aix-7.1-power PA-4719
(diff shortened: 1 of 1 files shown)


cthorn42 (approved):
We got a thumbs up from RE about this in the linked Slack thread, I'm good to go on this also.

PR #2385, opened 2023-08-14
(PA-5665) Install OpenCSW GCC on Solaris

mhashizume:
Compiling native extensions for the sqlite3 gem fails for Solaris currently because it attempts to use a compilation flag not understood by the version of GCC we use on Solaris.

This commit installs the OpenCSW version of GCC on Solaris to enable the compilation of native extensions for the sqlite3 gem.

### DIFF
--- a/acceptance/tests/validate_vendored_ruby.rb
+++ b/acceptance/tests/validate_vendored_ruby.rb
@@ -3,7 +3,6 @@ require 'puppet/acceptance/temp_file_utils'
 extend Puppet::Acceptance::CommandUtils
 
 confine :except, :platform => 'aix-7.2-power' # PA-5654
-confine :except, :platform => 'solaris-11.4-i386' # PA-5665
 
 def package_installer(agent)
   # for some reason, beaker does not have a configured package installer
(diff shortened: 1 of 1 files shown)


mhashizume:
Passing test here: https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/RMM_COMPONENT_TO_TEST_NAME=puppet_agent,SLAVE_LABEL=k8s-beaker,TEST_TARGET=solaris114-64a/1167/

cthorn42 (approved):

acceptance/tests/validate_vendored_ruby.rb
```diff
@@ -31,13 +30,30 @@ def setup_build_environment(agent)
...
+      "rdepend=nocheck\n"\
+      "space=quit\n"\
+      "setuid=nocheck\n"\
+      "conflict=nocheck\n"\
+      "action=nocheck\n"\
+      "basedir=default"
+    vanagon_noask_path = "/var/tmp/vanagon-noask"
+    on(agent, "echo \"#{vanagon_noask_contents}\" > #{vanagon_noask_path}")
+
+    vanagon_pkgutil_contents = "mirror=https://artifactory.delivery.puppetlabs.net/artifactory/generic__remote_opencsw_mirror/testing"
```
cthorn42, line 49: If someone who is running this test is not on our network will using the mirror setting cause a failure on the pkgadd command?

mhashizume: It looks like it will fail, but other parts of this test are dependent on internal domains (AIX packages in `install_dependencies`) and it's an easy enough workaround that I don't mind leaving it in

PR #2399, opened 2023-09-29
(PA-5632) Set SELinux test to high

mhashizume:
Prior to this PR, the SELinux acceptance test was set to low priority, causing the test to be skipped.

However, we've seen multiple regressions on SELinux behavior in puppet-agent, so we should always run this test to catch them early.

This PR updates the SELinux acceptance test to high priority, ensuring that it's always run.


mhashizume:
Setting this PR to draft until other work has been completed.

AriaXLi (approved):

PR #2405, opened 2023-10-10
(PA-4611) Added Debian11 ARM for nightlies ship

amitkarsale:
updated build_defaults to allow Debian11 aarch64 to ship to internal nightlies

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -6,6 +6,7 @@ project: 'puppet-agent'
 foss_platforms:
   - debian-10-amd64
   - debian-11-amd64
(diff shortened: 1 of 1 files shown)


joshcooper:
Just adding a note that this is blocked on adding Debian 11 as a `BUILD_TARGET`. Otherwise, if we merge this first, we'll try to ship non-existent builds. Also the ticket PA-4611 was created in 2022, before our templates were updated, so ignore the ticket description. The confluence page is correct though.

joshcooper:
@amitkarsale this looks to be dup of #2414 so I think we can close this?

PR #2406, opened 2023-10-10
(PA-4611) Added Debian11 ARM for nightlies ship

amitkarsale:
updated build_defaults to allow Debian11 aarch64 to ship to internal nightlies


joshcooper:
@amitkarsale this looks to be a dup of https://github.com/puppetlabs/puppet-agent/pull/2415, can we close this?

PR #2407, opened 2023-10-20
(PA-4611) Update build_defaults.yaml to enable Debian 11 (ARM)

imaqsood:
Update build_defaults.yaml to enable Debian 11 (ARM) for internal nightlies ship


joshcooper (commented):
I think this is a dup of https://github.com/puppetlabs/puppet-agent/pull/2405?

imaqsood:
> I think this is a dup of #2405?

PR #2442, opened 2023-12-08
(PA-5763) add debian 12 platforms for 7.x

h0tw1r3:
Includes amd64 for PA-5553


amitkarsale (approved):
you might need to rebase your code. Rest all looks good.
Also attach vanagon generic pipeline link for reference. Thanks!

PR #2453, opened 2023-12-20
(PA-5555) add debian-12 build_defaults for 7.x internal nightlies

h0tw1r3:
amd64 and aarch64, include PA-5761


amitkarsale (approved):
LGTM

joshcooper (changes requested):
I believe `debian-12-*` have to be added as a BUILD_TARGET in CJC for the `puppet-agent#7.x` pipeline otherwise we'll try to ship something we didn't build.

PR #2456, opened 2023-12-21
(PA-5908) temp PR for updating pxp-agent tag

amitkarsale:


joshcooper:
Rather than doing this (which won't work because only a subset of platforms were built) it's better to run the vanagon generic builder pipeline with `VANAGON_REPO_BRANCH=202312201` and only the missing platforms in `BUILD_TARGETS`. The pipeline will build and promote it to the existing directory https://builds.delivery.puppetlabs.net/pxp-agent/202312201/artifacts/

The only time this doesn't work is if you need changes in the pxp-agent-vanagon that were merged after the `202312201` tag

amitkarsale:
Thanks for the tip @joshcooper

PR #2462, opened 2024-01-16
(maint) Update facter and pxp-agent for already released versions

tvpartytonight:


mhashizume (approved):

configs/components/pxp-agent.json
```diff
@@ -1 +1 @@
-{"location":"https://builds.delivery.puppetlabs.net/pxp-agent/202312203/artifacts/","version":"202312203"}
+{"location":"https://builds.delivery.puppetlabs.net/pxp-agent/202312203/artifacts/","version":"1.15.19"}
```
joshcooper, line 1: I don’t think we want to make this change. 1.15.19 is the tag for the pxp-agent component contained in the pxp-agent-vanagon project. And the latter is a component in this vanagon repo

configs/components/facter.json
```diff
@@ -1 +1 @@
-{"url":"git@github.com:puppetlabs/facter.git","ref":"81a6e127335edd2d045a606a56c6c63d5347179d"}
+{"url":"git@github.com:puppetlabs/facter.git","ref":"4.5.2"}
```
joshcooper, line 1: I don’t think this is needed either as 81a6 is what we tagged at 4.5.2, https://github.com/puppetlabs/facter/releases/tag/4.5.2. My recollection is the pipeline tags the components and later automatically replaces the sha with the corresponding tag

PR #2464, opened 2024-01-17
Revert "(maint) Update facter and pxp-agent for already released vers…

tvpartytonight:
…ions"

This reverts commit 5a052a2369ef40ae490da9d6e2e13c725dcec5ef.


tvpartytonight:
@joshcooper mentioned that the sha is fine to keep for facter, and we actually need the vanagon version already existing, not the tagged version for that particular vanagon project. See https://github.com/puppetlabs/puppet-agent/pull/2462#pullrequestreview-1826365956

joshcooper (approved):

PR #2465, opened 2024-01-18
(PA-5981) Update smoke test to install postgresql-14

tvpartytonight:


tvpartytonight:
I have tested this with the 7.28.0 release and the 8.4.0 release.

donoghuc:
Wouldnt we need this in both 7.x and main? Does the install method from puppetlabs-puppetdb module not run in CI? i'm not sure how that would pass without intervention as the latest puppetdb module as it defaults to trying to install postgres 11 with an out-dated puppetlabs-postgresql module.

tvpartytonight:
@donoghuc we have a backport function with the label `backport 7.x` that we will run after this has been merged.

donoghuc:
OH right, i forgot that branch strategy! LGTM

donoghuc (approved):

mhashizume (approved):

PR #2474, opened 2024-01-23
Upgrade AlmaLinux release package

mhashizume:
We have started to see test failures on AlmaLinux due to outdated keys when validating packages. This commit updates tests to upgrade the AlmaLinux release package to pull in new keys to allow other packages to be installed.


mhashizume:
This will need to be backported to 7.x as well.

joshcooper (approved):

PR #2503, opened 2024-03-06
Update core modules

mhashizume:
This commit updates all core modules. The notable changes from these new modules are mainly:

- Removal of concurrent-ruby as a dependency to avoid issues with Puppet
- Support for Amazon Linux


mhashizume:
This PR should be backported after merge.

tvpartytonight (approved):

PR #2512, opened 2024-04-15
(PA-6381): Added el-9-ppc64le platform for puppet-agent

span786:
Added el-9-ppc64le platform for puppet-agent

### DIFF
--- a/configs/platforms/el-9-ppc64le.rb
+++ b/configs/platforms/el-9-ppc64le.rb
@@ -1,0 +1,8 @@
+platform 'el-9-ppc64le' do |plat|
+  plat.inherit_from_default
+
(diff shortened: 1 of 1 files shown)


span786:
Link to Vanagon generic builder : https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/BUILD_TARGET=el-9-ppc64le,SLAVE_LABEL=k8s-worker/2801/consoleFull

mhashizume:
Can we inherit from the Vanagon defaults for this platform? Do we still need the subscription workaround?

We can't check on the ticket referenced in the comment at the moment because of issues with the archived Jira site, but we should make sure that we're not carrying this workaround forward from RHEL 8 PPC to newer platforms if we don't need it anymore

shubhamshinde360 (approved):

PR #2559, opened 2024-10-04
Update macOS signing server

mhashizume:


mhashizume:
This will need to be backported to 7.x as well.

cthorn42 (approved):

